PR #97, opened 2023-12-08
postfix: enforce TLSv1.2, disallow some insecure TLS ciphers

### LINKED ISSUE #77: Harden TLS settings
According to https://www.hardenize.com/report/c1.testrun.org/1699901096#email_tls, we support several outdated TLS versions (TLS 1.0 and TLS 1.1), as well as some outdated ciphers and too small diffie-hellman parameters.

We might also look at our dovecot configuration, it probably has similar problems.

It's unclear to me if this is a real life problem, or if those outdated TLS stuff might be necesary for compability reasons.

missytake:
https://www.hardenize.com/report/c3.testrun.org/1702069777#email_tls is already mostly happy, but not entirely: 

"Reconfigure server to use forward secrecy and authenticated encryption

Even though this server supports TLS 1.2, the cipher suite configuration is suboptimal. We recommend that you reconfigure the server so that the cipher suites providing forward secrecy (ECDHE or DHE in the name, in this order of preference) and authenticated encryption (GCM or CHACHA20 in the name) are at the top. The server must also be configured to select the best-available suite."

I already tried some of the options mailcow.testrun.org uses, but avoided just copy-pasting everything because I still hope we can keep it minimal, and want to understand what's going on.

fix #77

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/main.cf.j2
+++ b/cmdeploy/src/cmdeploy/postfix/main.cf.j2
@@ -24,7 +24,22 @@ smtp_tls_security_level=may
 smtp_tls_session_cache_database = btree:${data_directory}/smtp_scache
 smtp_tls_policy_maps = socketmap:inet:127.0.0.1:8461:postfix
 smtpd_tls_protocols = >=TLSv1.2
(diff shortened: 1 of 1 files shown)


missytake:
Whatever, this change is good enough to merge. 

image

PR #114, opened 2023-12-12
dovecot: unconditionally delete all mails after 40 days

missytake:
Does anyone have an idea how to test this?

### DIFF
--- a/chatmaild/src/chatmaild/ini/chatmail.ini.f
+++ b/chatmaild/src/chatmaild/ini/chatmail.ini.f
@@ -17,8 +17,8 @@ max_user_send_per_minute = 60
 # maximum mailbox size of a chatmail account
 max_mailbox_size = 100M
 
-# time after which mails are unconditionally deleted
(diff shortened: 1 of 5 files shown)


missytake:
This simply doesn't work yet. No idea why.

missytake:
Ah, it seems `doveadm expunge -A` doesn't work because of `iterate_disable=yes` in https://github.com/deltachat/chatmail/blob/main/cmdeploy/src/cmdeploy/dovecot/auth.conf#L2. We might need a different way to iterate over our users.

missytake:
This should work now. I'm not sure if we need/can have a useful test for it; it could involve creating two accounts, sending a message, running a modified find command with -mtime +0, and asserting the message *doesn't* arrive (which means waiting for a few seconds and praying it isn't just a delay I guess). I don't like the idea very much, it doesn't really test the find command in the cron job, only the find command in the test, so I'd just leave it out.

PR #117, opened 2023-12-12
Improving UX of setting up DNS entries

### LINKED ISSUE #105: Improve DNS UX during setup
As discussed:

- [x] `cmdeploy init` should recommend and print out two DNS entries the user needs to set, like this:  
```
A       $mail_domain.               <your server's IPv4 address>
CNAME   mta-sts.$mail_domain.       $mail_domain.
```
- [x] if those DNS entries don't exist yet, `cmdeploy run` should fail with an error before executing any deployment steps.
- [x] `cmdeploy dns` should check which DNS entries already exist, and only print out those which are not yet correct.
- [x] `cmdeploy dns` should print out AAAA records for `$mail_domain` and `mta-sts.$mail_domain`.
- [x] `cmdeploy dns` should print out PTR records and explain that you need to set them at the host provider, not DNS provider (if they are two different institutions).
- [x] `cmdeploy dns` should have a command line option to print a full DNS zone file instead.

missytake:
fix #105

### DIFF
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -15,6 +15,7 @@ from pathlib import Path
 
 from termcolor import colored
 from chatmaild.config import read_config, write_initial_config
+import cmdeploy.dns
 
 
 #
@@ -33,10 +34,29 @@ def init_cmd_options(parser):
 def init_cmd(args, out):
     """Initialize chatmail config file."""
     if args.inipath.exists():
-        out.red(f"Path exists, not modifying: {args.inipath}")
-        raise SystemExit(1)
-    write_initial_config(args.inipath, args.chatmail_domain)
-    out.green(f"created config file for {args.chatmail_domain} in {args.inipath}")
+        print(f"Path exists, not modifying: {args.inipath}")
+    else:
+        write_initial_config(args.inipath, args.chatmail_domain)
+        out.green(f"created config file for {args.chatmail_domain} in {args.inipath}")
+    ipaddress, _ = cmdeploy.dns.resolve(args.chatmail_domain)
+    mta_ipadress, _ = cmdeploy.dns.resolve("mta-sts." + args.chatmail_domain)
+    entries = 0
+    to_print = ["Now you should add %dnsentry% at your DNS provider:\n"]
+    if not ipaddress:
+        entries += 1
+        to_print.append(f"\tA\t{args.chatmail_domain}\t\t<your server's IPv4 address>")
+    if not mta_ipadress:
+        entries += 1
+        to_print.append(f"\tCNAME\tmta-sts.{args.chatmail_domain}\t{args.chatmail_domain}.")
+    if entries == 1:
+        singular = "this entry"
+    elif entries == 2:
+        singular = "these entries"
+    else:
+        return
+    to_print[0] = to_print[0].replace("%dnsentry%", singular)
+    for line in to_print:
+        print(line)
 
 
 def run_cmd_options(parser):
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -173,7 +173,7 @@ def dns_cmd(args, out):
                 if f" {typ} " in line:
                     domain, value = line.split(f" {typ} ")
                     current = dns.get(typ, domain.strip()[:-1])
-                    if current != value:
+                    if current != value.strip():
                         to_print.append(line)
             if " MX " in line:
                 domain, typ, prio, value = line.split()
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -3,7 +3,6 @@ Provides the `cmdeploy` entry point function,
 along with command line option and subcommand parsing.
 """
 import argparse
-import datetime
 import shutil
 import subprocess
 import importlib.resources
@@ -15,7 +14,7 @@ from pathlib import Path
 
 from termcolor import colored
 from chatmaild.config import read_config, write_initial_config
-from cmdeploy.dns import DNS
+from cmdeploy.dns import show_dns, check_necessary_dns
 
 
 #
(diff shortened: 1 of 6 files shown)


cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -33,10 +34,33 @@ def init_cmd_options(parser):
...
+        )
+    if entries == 1:
+        singular = "this entry"
+    elif entries == 2:
+        singular = "these entries"
+    else:
+        return
+    to_print[0] = to_print[0].replace("%dnsentry%", singular)
+    for line in to_print:
+        print(line)
```
hpk42, line 62: why not print out both lines unconditionally and avoid the singular/plural shenenigans? 
Maybe in green if it resolves correctly, red if not-exist/wrongly?

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -73,24 +117,104 @@ def read_dkim_entries(entry):
...
+        current = dns.get("TXT", domain.strip()[:-1]).replace('" "', '"\n "')
+        current = f"( {current} )"
+        if current.replace(";", "\\;") != data:
+            to_print.append(dkim_entry)
+    if to_print:
+        to_print.insert(
+            0, "You should configure the following DNS entries at your provider:\n"
+        )
+        to_print.append(
+            "\nIf you already configured the DNS entries, don't worry. It can take a while until they are public."
```
hpk42, line 213:
```suggestion
            "\nIf you already configured the DNS entries, wait a bit until the DNS entries propagate to the Internet." 
```

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -73,24 +117,104 @@ def read_dkim_entries(entry):
...
+        )
+        print("\n".join(to_print))
+    else:
+        out.green("Great! All your DNS entries are correct.")
```
hpk42, line 217:
```suggestion
        out.green("Great! All your DNS entries are correct 😅.")
```

missytake:
cmdeploy run could also just show the recommended DNS entries which init already does if they aren't there.

missytake:
> it looks good overall -- but some notes. `init` is now a network-using operations, and uses cloudflare/google implicitely i think when merging this here, we should also open an issue to move it to resolving locally, also along what link2xt commented.

It is now resolved on the server with the unbound DNS resolver instead of using centralized HTTP APIs.

> I haven't tested the new DNS workflow myself -- someone else than nami should do it (set up a new host) using the new workflows and see if it works.

c6 would be free :)

> i initially thought 'cmdeploy dns' to produce zonefile output on stdout, and write loggins/other things to 'stdinfo/stderr' so that shell-redirection works. It's kind of a standard pattern, no?

The stdout output is now used for helpful hints for users. Advanced users who can directly put a DNS zonefile into a name server can use the `--zonefile` argument.

missytake:
> Why this is merging into `deploy-config` branch instead of main, is it on purpose?

it was based on an open PR at the time, now it should merge into main. We didn't delete the `deploy-config` branch after merging #111, then the target branch doesn't change automatically when PRs are merged.

missytake:
btw, meanwhile a user [requested this feature](https://support.delta.chat/t/chatmail-self-hosted-deployment-never-get-success/2852/4?u=missytake), let's merge soon :sweat_smile:

cmdeploy/src/cmdeploy/tests/test_cmdeploy.py
```diff
@@ -27,6 +27,7 @@ def test_init(self, tmp_path):
         config = read_config(inipath)
         assert config.mail_domain == "chat.example.org"
 
+    @pytest.mark.xfail(reason="init doesn't exit anymore, check for CLI output instead")
```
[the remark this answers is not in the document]
missytake: exit, not exist. It used to do sys.exit(1) if it didn't overwrite the config. The exit code is 0 now in that case, so the test stopped working.

missytake:
`cmdeploy init` can be tested manually like this:

```
# test commands

cmdeploy init --config c8 c8.testrun.org
cmdeploy init --config lol loltry.org
cmdeploy init --config lef mail.lefherz.net
cmdeploy init --config c9ipv6 c9ipv6.testrun.org
cmdeploy init --config c9mta c9mta.testrun.org
cmdeploy init --config nonexist2 nonexist2.testrun.org
cmdeploy init --config nonexist nonexist.testrun.org
cmdeploy init --config c9 c9.testrun.org

# Case 0: c8.testrun.org

- VPS doesn't exist
- A @ entry is missing
- CNAME mta-sts entry is missing
- AAAA @ entry is missing

server dig fails, call local dig
print A and CNAME instructions
print "can't reach the server with $ssh"

# Case 1: loltry.org

- VPS exists
- A @ entry is missing
- CNAME mta-sts entry is missing
- AAAA @ entry is missing

call server dig, if that fails call local dig
print A and CNAME instructions

# Case 2: mail.lefherz.net

- VPS exists
- A @ entry is set
- CNAME mta-sts entry is missing
- AAAA @ entry is missing

call server dig, if that fails call local dig
print CNAME instructions

# Case 3: c9iv6.testrun.org

- VPS exists
- A @ entry is missing
- CNAME mta-sts entry is missing
- AAAA @ entry is set

call server dig, if that fails call local dig
print CNAME instructions

# Case 4: c9mta

- VPS exists
- A @ entry is missing
- CNAME mta-sts entry is set
- AAAA @ entry is missing

call server dig, if that fails call local dig
print A and CNAME instructions

# Case 5: nonexist2

- VPS doesn't exist
- A @ entry is missing
- CNAME mta-sts entry is set
- AAAA @ entry is set

call server dig, if that fails call local dig
print "all correct"
print "can't reach the server with $ssh"

# Case 6: nonexist

- VPS doesn't exist
- A @ entry is missing
- CNAME mta-sts entry is missing
- AAAA @ entry is set

call server dig, if that fails call local dig
print CNAME instructions
print "can't reach the server with $ssh"

Case 7: c9.testrun.org

- VPS exists
- A @ entry is missing
- CNAME mta-sts entry is set
- AAAA @ entry is set

call server dig, if that fails call local dig
print "all correct"
```

This could be adapted to pytest in parts, but it does require SSH access to some maybe existing VPS's so it's better tested manually imho.

PR #127, opened 2023-12-14
feat: one more paragraph to explain `chatmail`

### LINKED ISSUE #126: Define "chatmail"
https://nine.testrun.org/info.html talks about "chatmail" and users don't understand what it means: https://support.delta.chat/t/i-dont-know-what-chatmail-is/2850

Septias:
close #126

### DIFF
--- a/www/src/info.md
+++ b/www/src/info.md
@@ -3,6 +3,11 @@
 
 ## More information 
 
+`nine.testrun.org` provides a low-maintenance, resource efficient and 
+interoperable e-mail service for everyone. What's behind a `chatmail` is 
+effectively a normal e-mail address just like any other but optimized 
+for the usage in chats, especially DeltaChat.
+
 ### Choosing a chatmail address instead of using a random one
 
 In the Delta Chat account setup 


www/src/info.md
```diff
@@ -3,6 +3,11 @@
 
 ## More information 
 
+`nine.testrun.org` provides a low-maintenance, resource efficient and 
+interoperable e-mail service for everyone. What's behind a `chatmail` is 
+effectively a normal e-mail address just like any other but optimized 
+for the usage in chats, especially DeltaChat.
```
r10s, line 9: i find the "what's behind a 'chatmail'" part hard to understand.

is the gist the following?

```suggestion
`nine.testrun.org` provides a low-maintenance, resource efficient and 
interoperable e-mail service for everyone.
This is done by using "chatmail",
effectively a normal e-mail address just like any other but optimized 
for the usage in chats, especially DeltaChat.
```

hpk42: i'll merge and we can further modify then. thanks.

PR #150, opened 2023-12-22
Use dig +short option to simplify DNS parsing

link2xt:
Without this option parsing of answer was flaky
as for long records like
_submission._tcp.nine.testrun.org.
dig printed the result with a space rather
than tab as a separator and .split("\t") did not work.

This change makes the `dig` command print the answer in the form we need so there is no need for complex parsing other than taking the first line.

`-r` option is added to make sure options are not changed by .digrc in the root home directory.


missytake (approved):
Super nice! So reading man pages *is* useful after all.

PR #155, opened 2023-12-27
Update master.cf.j2

nudeldudel:
Add submission-header-cleanup to reduce the meta-data

Main critcal comments on chating via eMail is the mass of Meta-Data.

To activate submission-header-cleanup, just a few lines had to add to master.cf.

The filterfile ist found at `/etc/postfix/submission_header_cleanup` and looks like that:

```
### Entfernt Datenschutz-relevante Header aus E-Mails von MTUAs

/^Received:/            IGNORE
/^X-Originating-IP:/    IGNORE
/^X-Mailer:/            IGNORE
/^User-Agent:/          IGNORE
```


missytake:
Ah, great, we wanted to tackle that soon as well, thanks!

We can't really merge it like this right now, because we need to add a pyinfra operation to add the filterfile, and I'd also translate the comments to english. I can do that in a separate branch.

missytake:
I'd close this PR in favor of #156.

PR #161, opened 2024-01-04
Update nginx.conf.j2

nudeldudel:
For more SEO-friendly, www should have a CNAME and a canonical redirect into the nginx.conf.


nudeldudel:
This is a point. Yes SEO want this, why ever, but TLS is not nessesary,
because the 301 header feedback comes before TLS handshake.

Another cname would be nessesary.

That lifts up the site by ca. 10%.

Now Mehl.cloud is 88% SEO friendly. Last percent could be fetch by
individual Meta-description. That is too heavy, I think

link2xt ***@***.***> schrieb am Fr., 5. Jan. 2024, 17:07:

PR #164, opened 2024-01-06
Add www. subdomain

link2xt:
This was requested at #161

I have no idea about SEO, but makes sense that www.nine.testrun.org redirects to nine.testrun.org if someone types in the address with www in front.


missytake (approved):
Now `cmdeploy dns` works as well.

PR #166, opened 2024-01-07
Update autoconfig XML URL with RFC draft

link2xt:
Old page does not exist anymore and linking to web archive is not nice.


missytake (approved):

missytake:
We can do that, but maybe linking to the XML file itself makes more sense? It's easier for debugging. And admins usually don't need to know that it's a standard.

hpk42:
both a link to nine.testrun.org/.well-known/ autoconfig url and a link to the draft standard make sense.  But i am merging this one because it fixes a link which shouldn't be blocked.

PR #169, opened 2024-01-09
Hardening nginx.j2

nudeldudel:
The same as in Issue #77 , but for nginx now.

### DIFF
--- a/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
+++ b/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
@@ -18,7 +18,8 @@ http {
 	include /etc/nginx/mime.types;
 	default_type application/octet-stream;
 
-	ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3;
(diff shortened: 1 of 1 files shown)


hpk42:
Not sure there is much point in hardening nginx -- it's not required for chatmail usages from delta chat, and the distributions / debian defaults should be fine enough.  Thanks for the suggestion, though!

nudeldudel:
I don't agree completly. The complete server will be attacked, as it is a concurence to "the others". 

On the other site, there is no reason against this change. Or will user be blocked, becuase only can use weak ciphers....

If chatmail is audited by "experts", this will be an other point of critic.

hpk42:
so i guess it's ok to change the PR to only require TLS v1.2 and not change ciphers?  would be fine with that.

n3ph:
In the end, nginx is only used for the "welcome page" to register a new account. 
Since the registration process is usually done through the application (handheld, desktop), TLS version constraints are (probably/hopefully?) set by client implementation.

To narrow down used TLS versions (keep downgrade attacks in mind‼️), this change absolutely makes sense and I explicitly challenge closing this PR.

Best regards ✌🏼

PR #175, opened 2024-01-12
Add greeterbot to send example .xdc's to users

missytake:
better merge https://github.com/deltachat/chatmail/pull/157 before this one, otherwise users who try to write to xstore@testrun.org will get an error.


missytake:
Maybe we can add a config flag to disable this bot, not every chatmail admin wants this I guess.

chatmaild/src/chatmaild/greeterbot.py
```diff
@@ -0,0 +1,132 @@
...
+    def greet_users(self):
+        with self.db.read_connection() as conn:
+            users = conn.get_user_list()
+        new_users = users.difference(self.existing_users)
+        self.existing_users = users
+        time.sleep(20)  # wait until Delta is configured on the user side
+        for user in new_users:
+            for ci_prefix in ["ac1_", "ac2_", "ac3_", "ac4_", "ac5_", "ci-"]:
+                if user.startswith(ci_prefix):
+                    continue
```
missytake, line 86: > This will break our CI, online tests don't expect greeting bot showing up.

I thought of this, don't worry.

missytake:
It's deployed on c7.testrun.org, you can create an account there to try it out.

hpk42:
works for me, nice!  two notes: 

- it says "hello at try.webxdc.org" instead of c7.testrun.org 

- maybe better link to https://webxdc.org/apps in the last greeterbot
  message, instead of pointing to xstore (which may or may not
  be whitelisted on a chatmail instance, and also requires more taps/clicks 
  to get to the store view than the direct URL).

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -46,7 +46,7 @@ def is_allowed_to_create(config: Config, user, cleartext_password) -> bool:
         len(localpart) > config.username_max_length
         or len(localpart) < config.username_min_length
     ):
-        if localpart != "echo":
+        if localpart != "echo" and localpart != "hello":
```
hpk42, line 49: the pattern i prefer for this is: 
```suggestion
        if localpart not in ("echo", "hello"): 
```

chatmaild/src/chatmaild/greeterbot.py
```diff
@@ -0,0 +1,132 @@
...
+    ac.set_config("mvbox_move", "0")
+    ac.set_config("sentbox_watch", "0")
+    ac.set_config("bot", "1")
+    ac.set_config("mdns_enabled", "0")
+
+    if not ac.is_configured():
+        cleartext_password = "".join(
+            secrets.choice(ALPHANUMERIC_PUNCT)
+            for _ in range(chatmail_config.password_min_length + 3)
+        )
```
hpk42, line 48: i suggest to use `chatmaild.newemail.create_newemail_dict()['password']` to not add more places where passwords are generated.

chatmaild/src/chatmaild/greeterbot.py
```diff
@@ -0,0 +1,132 @@
...
+        self.domain = account.get_config("addr").split("@")[1]
+        with self.db.read_connection() as conn:
+            self.existing_users = conn.get_user_list()
+
+    def greet_users(self):
+        with self.db.read_connection() as conn:
+            users = conn.get_user_list()
+        new_users = users.difference(self.existing_users)
+        self.existing_users = users
+        time.sleep(20)  # wait until Delta is configured on the user side
```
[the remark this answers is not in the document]
missytake: There is also the time the user needs to choose a username and/or profile picture, that's more what those 20 seconds are for. Sure, the time is somewhat random.

missytake:
> Will it greet existing users as well or only new users?

only new users.

missytake:
We most likely don't want to run this on every chatmail instance; we will deploy it on try.webxdc.org.

PR #176, opened 2024-01-14
rspamd: give the reason to MTA when incoming mail is rejected

link2xt:
This is not secret but makes it easier for mail server admins to debug why chatmail does not accept their emails. If the server generates bounce messages, users will also see this and can redirect to their server support.
It also shows up in /var/log/rspamd/rspamd.log on chatmail server.


missytake (approved):
Okay, that does make sense. And I like how it's well documented and easy to modify for chatmail admins.

ralphtheninja:
Would it be possible to split e.g. the DKIM check so it can distinguish between missing and invalid signature? I.e. fail with "Missing DKIM signature" and fail with "Invalid DKIM signature"?

PR #178, opened 2024-01-15
Nicer /new URL for new accounts and redirect GET requests

### LINKED ISSUE #162: Going directly to https://nine.testrun.org/cgi-bin/newemail.py shows JSON in a browser
During testing of nine.testrun.org user scanned QR code from the website with a browser (IIRC it was a Brave browser) and ended up with a browser showing JSON directly, even though Delta Chat was installed on the phone. Clicking on the QR code in a browser worked.

We need to reproduce this and find a way to deal with the problem, probably redirecting to `dcaccount:` URL if we detect a browser request.

link2xt:
If user types in https://nine.testrun.org/new manually in the browser, at least Firefox and Brave suggest to open the app after following the redirect.

Fixes #162

Here is an example of opening in Mull (Firefox) private mode:
screen1

### DIFF
--- a/cmdeploy/src/cmdeploy/genqr.py
+++ b/cmdeploy/src/cmdeploy/genqr.py
@@ -6,7 +6,7 @@ import io
 
 
 def gen_qr_png_data(maildomain):
-    url = f"DCACCOUNT:https://{maildomain}/cgi-bin/newemail.py"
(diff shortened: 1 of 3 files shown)


missytake (approved):
Could be "new_account" or something more verbose, but doesn't matter much.

PR #180, opened 2024-01-16
Stop requesting DMARC reports

link2xt:
Nobody reads these XML reports
and we know our DKIM is valid
when `cmdeploy dns` is happy.


hpk42 (approved):
sounds good, and one less thing admins need to monitor/worry about.

missytake (approved):
Yeah.

PR #184, opened 2024-01-17
WIP: generate dkim_selector with cmdeploy init

missytake:
Probably we should not generate it in `cmdeploy init`, but instead check if there is already a DKIM key.


missytake:
> What happens though if you and me deploy to nine.testrun.org from time to time? We will have different selectors and keep switching between them, should be fine?

should be fine as long as we take care to publish the corresponding DNS records.

missytake:
Let's only do it on staging.testrun.org for now.

missytake:
This is not necessary anymore, as staging.testrun.org isn't reset on every run.

PR #186, opened 2024-01-19
Replace rspamd with OpenDKIM

### LINKED ISSUE #179: Replace rspamd SPF/DKIM/DMARC with a filter checking DKIM and From: domain alignment
Currently we use rspamd with [force actions](https://rspamd.com/doc/modules/force_actions.html) module to reject all emails that don't have a valid DKIM signature (#20, #176). If email has a valid DKIM signature, checking SPF and DMARC does not add anything and even resolving additional DNS records for this is not worth it.

Unless we have to revert this DKIM check because it causes too much trouble, we can replace `rspamd` with our own SMTP filter in Python using https://pypi.org/project/dkimpy/ It is maintained at https://git.launchpad.net/dkimpy/ and can verify as well as sign DKIM. So we can sign DKIM using this library in filtermail and check DKIM in a new incoming filter that we will need anyway if we want to start doing push notifications.

We can still keep `rspamd` for DKIM signing or restore OpenDKIM, DKIM-signing is independent. `rspamd` is good enough for this, it even signs Autocrypt headers out of the box.

### LINKED ISSUE #182: Enforce equality of RFC5322.From (IMF From) domain and DKIM domain
Currently we reject emails that fails DKIM check:
https://github.com/deltachat/chatmail/blob/3665d957a7243695a3db4d45918118e7eed74add/cmdeploy/src/cmdeploy/rspamd/force_actions.conf#L9

We also reject emails failed by DMARC policy:
https://github.com/deltachat/chatmail/blob/3665d957a7243695a3db4d45918118e7eed74add/cmdeploy/src/cmdeploy/rspamd/force_actions.conf#L41

However, apparently `dkim` module does not check that DKIM signature domain is [aligned](https://www.rfc-editor.org/rfc/rfc7489#section-3.1) with RFC5322.From domain. This is [the only good] part of DMARC standard and is therefore checked by DMARC module.

However, DMARC module checks very little even if `p=reject` policy is published for the `From` header domain. It will pass if either SPF or DKIM passes, while we want to enforce valid DKIM signatures. It will also pass if there is no strict reject policy or if there is simply no DMARC record.

We need a way to check that there is a DKIM signature corresponding to the `From:` header of the IMF aka RFC5322.From regardless of DMARC. Then we don't even need to resolve SPF or DMARC records. If there is no way to do this with `rspamd`, seems the solution would be to implement our own checker, see #179

link2xt:
OpenDKIM configuration
has two Lua scripts defining strict DKIM policy.

screen.lua filters out signatures that do not correspond
to the From: domain so they are not even checked.
final.lua rejects mail if it is not outgoing
and has no valid DKIM signatures.

OpenDKIM is configured as a milter on port 25 smtpd
to check DKIM signatures
and on mail reinjecting smtpd
to sign outgoing messages with DKIM signatures.

Closes #179
Closes #182

### DIFF
--- a/cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
+++ b/cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
@@ -48,7 +48,7 @@ def test_reject_missing_dkim(cmsetup, maildata, from_addr):
     recipient = cmsetup.gen_users(1)[0]
     msg = maildata("plain.eml", from_addr=from_addr, to_addr=recipient.addr).as_string()
     with smtplib.SMTP(cmsetup.maildomain, 25) as s:
-        with pytest.raises(smtplib.SMTPDataError, match="missing DKIM signature"):
+        with pytest.raises(smtplib.SMTPDataError, match="No valid DKIM signature"):
             s.sendmail(from_addr=from_addr, to_addrs=recipient.addr, msg=msg)
 
 


missytake:
I received the following DKIM signature from c20.testrun.org:

```
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/simple; d=c20.testrun.org;
	s=opendkim; t=1705684900;
	bh=9nI2wlDG0Ykq3gVRJxWi2HowGwYlGYJr4KFoLXtD3bE=;
	h=Subject:From:To:Date:In-Reply-To:References:From;
	b=BT9br06BVJXpKq/hG4TimZaXcOyoouXjQng4Uckxj+omtutf+lvL1Rvf8ttPaSD3d
	 HW71e3tW26ouvU+ttX4hmGAuvoAAWpM2eONtwNkxKAgpIA3BryV3j8+wzMlWYrH8rO
	 wTiObK2/EVyW/sfvdWHOAO+C4KW9t+iRVk50WUw+Yk7ybkJrSrumopG07mRQ/z4PQw
	 V4SDcicpadm3TqRpKhuD4D7dU5tFoLlXYkqb4cBoKAmPUAw0N1kUeKuq7WHGVetMXS
	 kUmDIVmseT7MWNJFC8zn12bs9AGNfnJuub7HO3t6YvUJ1hX4J4rVnj3GJTIq7s1fuO
	 imB3Bzqqy2WXw==
```

does that mean the autocrypt header was not signed?

missytake (commented):
Seems to work nicely :) one thing I noticed is that senders get a nice error message if their server doesn't have working DKIM, but opendkim doesn't log the rejection in journalctl. Can we add that?

c9.testrun.org is a test host without DKIM if you want to test it.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -494,15 +493,15 @@ def deploy_chatmail(config_path: Path) -> None:
...
     systemd.service(
-        name="Start and enable rspamd",
-        service="rspamd.service",
+        name="Start and enable OpenDKIM",
+        service="opendkim.service",
         running=True,
         enabled=True,
-        restarted=rspamd_need_restart,
+        restarted=opendkim_need_restart,
     )
```
missytake, line 505: Hm, starting the opendkim service during cmdeploy fails for me:

```
Jan 23 17:06:42 c1 systemd[1]: Starting opendkim.service - OpenDKIM Milter...
Jan 23 17:06:42 c1 opendkim[359568]: opendkim: milter socket must be specified
Jan 23 17:06:42 c1 opendkim[359568]:         (use "-?" for help)
Jan 23 17:06:42 c1 systemd[1]: opendkim.service: Control process exited, code=exited, status=78/CONFIG
Jan 23 17:06:42 c1 systemd[1]: opendkim.service: Failed with result 'exit-code'.
Jan 23 17:06:42 c1 systemd[1]: Failed to start opendkim.service - OpenDKIM Milter.
```

missytake:
> @missytake I pushed another commit that apparently fixes the bug. Deployed ccc20.testrun.org from main, then deployed this PR.
> 
> Could you try the same thing on another domain?

sounds good, will try tomorrow... my eyes hurt already^^

hpk42 (approved):
lgtm, a nice simplifying improvement.  Didn't test it myself though and also am not sure about the one "you need to run it twice" issue.  That probably needs to be addressed before merging.

missytake (approved):
Works :) I first deployed main on c4 and then deployed this PR. A test still expected the old error message, but apart from that it worked.

One thing which could be improved is: printing "please run `cmdeploy dns` again, your DKIM setup changed" if `apt.packages(packages="rspamd", present=False).changed == True`. That would help admins notice that they need to update their DKIM record, I guess. But it's no blocker.

PR #190, opened 2024-01-20
Dovecot: setup METADATA

link2xt:
There is no dictionary to set additional attributes, but admin email can already be retrieved:

```
? GETMETADATA "" (/shared/admin)
* METADATA "" (/shared/admin {27} mailto:root@c20.testrun.org)
? OK Getmetadata completed (0.001 + 0.000 secs).
```


hpk42 (approved):
lgtm

PR #200, opened 2024-01-24
postfix: require that login matches envelope FROM

### LINKED ISSUE #151: postfix: master.cf references undefined variables mua_*
Postfix log has this:
```
Dec 24 00:11:47 c20 configure-instance.sh[3359]: postconf: warning: /etc/postfix/master.cf: undefined parameter: mua_client_restrictions
Dec 24 00:11:47 c20 configure-instance.sh[3359]: postconf: warning: /etc/postfix/master.cf: undefined parameter: mua_helo_restrictions
```

link2xt:
Testing that envelope FROM matches From: header
already happens in filtermail
and tested with `test_reject_forged_from`.

The most important part here is
`reject_sender_login_mismatch` check
documented in
<https://www.postfix.org/postconf.5.html#reject_sender_login_mismatch>.

Fixes #151


missytake:
`cmdeploy/src/cmdeploy/tests/online/test_1_basic.py::test_reject_forged_from` fails for the wrong reasons now, we have to submit the mail on port 25 and then check for the correct error message.

missytake (approved):
nice! works now.

missytake:
I really love our test framework.

Issue #185, opened 2024-01-18
Setup IMAP METADATA process


link2xt:
[IMAP METADATA](https://www.rfc-editor.org/rfc/rfc5464.html) extension can be used to pass arbitrary values between the client and the server.

[Dovecot METADATA](https://doc.dovecot.org/configuration_manual/imap_metadata/) needs a [dictionary backend](
https://doc.dovecot.org/configuration_manual/dict/#dict). The most flexible one is the dictionary proxy protocol, the same one as we already use for authentication.

As a first thing we can publish admin email as `/shared/admin` entry and test that it can be retrieved over plain IMAP: https://www.rfc-editor.org/rfc/rfc5464.html#section-6.2.2

Later we can extend this to publish iroh DERP proxy URL, let the client store configuration settings to configure their own account (message deletion timer, password etc.) and pass device token for apple push notifications.

This issue can be closed once we have a working constant entry for `/shared/admin` served by a dictionary server.

PR #201, opened 2024-01-24
Enable push_notification plugin

link2xt:
This PR enables `push_notification` Dovecot plugin and adds IMAP METADATA backend server in Python to store device tokens for notifications.

For the [driver](https://doc.dovecot.org/configuration_manual/push_notification/#drivers) we use [Lua](https://doc.dovecot.org/configuration_manual/push_notification/#lua-lua) and notify localhost daemon (also acting as METADATA dictionary, see #185) over HTTP about MessageNew events.

When a new message arrives Lua script "sets" `/private/messagenew` metadata to trigger sending notification to https://github.com/deltachat/notifiers via https://notifications.delta.chat/notify endpoint to deliver notification for Apple devices.
Notification happens when `/private/messagenew` is set. When a new message arrives to INBOX Lua plugin for push_notification will do `SETMETADATA INBOX /private/messagenew ""` by calling https://doc.dovecot.org/admin_manual/lua/#mail_user.metadata_set which will trigger sending previously set devicetoken to `notifiers`.

Subscribing to notifications works by doing `SETMETADATA` on the folder:
```
? SETMETADATA INBOX (/private/devicetoken foobar)
```

Core PR where setting device ID via metadata is going to be implemented: https://github.com/deltachat/deltachat-core-rust/pull/5286

---

Lua gets `mail_user` in https://doc.dovecot.org/configuration_manual/push_notification/#dovecot_lua_notify_begin_txn
Then it is possible to get any necessary metadata from https://doc.dovecot.org/admin_manual/lua/#mail_user.metadata_get
and if https://doc.dovecot.org/configuration_manual/push_notification/#dovecot_lua_notify_event_message_new is called, give notification to the local daemon via HTTP so it can forward to https://github.com/deltachat/notifiers.

### DIFF
--- a/cmdeploy/src/cmdeploy/dovecot/push_notification.lua
+++ b/cmdeploy/src/cmdeploy/dovecot/push_notification.lua
@@ -1,0 +1,32 @@
+function dovecot_lua_notify_begin_txn(user)
+  return user
+end
+
+function contains(v, needle)
+  for _, keyword in ipairs(v) do
+    if keyword == needle then
+      return true
+    end
+  end
+  return false
+end
+
+function dovecot_lua_notify_event_message_new(user, event)
+  local mbox = user:mailbox(event.mailbox)
+  mbox:sync()
+
+  if user.username ~= event.from_address then
+    -- Incoming message
+    if not contains(event.keywords, "$Auto") then
+      -- Not an Auto-Submitted message, notifying.
+
+      -- Notify METADATA server about new message.
+      mbox:metadata_set("/private/messagenew", "")
+    end
+  end
+
+  mbox:free()
+end
+
+function dovecot_lua_notify_end_txn(ctx, success)
+end
--- a/chatmaild/pyproject.toml
+++ b/chatmaild/pyproject.toml
@@ -10,6 +10,7 @@ dependencies = [
   "iniconfig",
   "deltachat-rpc-server",
   "deltachat-rpc-client",
+  "requests",
 ]
 
 [tool.setuptools]
@@ -20,6 +21,7 @@ where = ['src']
 
 [project.scripts]
 doveauth = "chatmaild.doveauth:main"
+chatmail-metadata = "chatmaild.metadata:main"
 filtermail = "chatmaild.filtermail:main"
 echobot = "chatmaild.echo:main"
 chatmail-metrics = "chatmaild.metrics:main"
--- a/chatmaild/src/chatmaild/chatmail-metadata.service.f
+++ b/chatmaild/src/chatmaild/chatmail-metadata.service.f
@@ -1,0 +1,10 @@
+[Unit]
+Description=Chatmail dict proxy for IMAP METADATA
+
+[Service]
+ExecStart={execpath} /run/dovecot/metadata.socket vmail {config_path}
+Restart=always
+RestartSec=30
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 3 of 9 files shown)


cmdeploy/src/cmdeploy/dovecot/push_notification.lua
```diff
@@ -0,0 +1,30 @@
...
+end
+
+function dovecot_lua_notify_begin_txn(user)
+  local ctx = {messages={}, user=user}
+  return ctx
+end
+
+function dovecot_lua_notify_event_message_new(ctx, event)
+  local mbox = ctx.user:mailbox(event.mailbox)
+  mbox:sync()
```
hpk42, line 13: is this `sync()` neccessary?

hpk42 (approved):
Didn't study the dovecot docs now but it all looks good!

r10s:
> This is mostly ready

yeah! thanks for pushing PUSH :)

>  but need to avoid notifying BCC-self (outgoing) messages.

we probably also want to avoid notifying webxdc-status updated, read-receipts and sync-messages - just saying, i cannot judge it from the existing code :)

hpk42:
On Tue, Feb 27, 2024 at 07:13 -0800, bjoern wrote:
> > This is mostly ready
> 
> yeah! thanks for pushing PUSH :)
> 
> >  but need to avoid notifying BCC-self (outgoing) messages.
> 
> we probably also want to avoid notifying webxdc-status updated, read-receipts and sync-messages - just saying, i cannot judge it from the existing code :)

Do i understand correctly that anything the notifiers-server sends to apple/device
will appear as a notification like "N new messages " on a user device
and that might need to be clicked before the app 
actually starts and can retrieve messages, right? 
Of so then read-receipts, sync-messages and (for now) also webxdc-status updates 
should not cause such user-visible message, indeed.

r10s:
> Do i understand correctly that anything the notifiers-server sends to apple/device
will appear as a notification like "N new messages " on a user device
and that might need to be clicked before the app
actually starts and can retrieve messages, right?

yes, this is my understanding and what was also discussed with @zeitschlag iirc. Apple will just display the notification as is, by default, there is no chance for deltachat-ios to filter or do whatever locally on the device - deltachat-ios won't be running in most situations and will not be started by just a notification being received.

with chance, deltachat-ios can later on (by some local wakeup) replace the generic notification by a more informative one silently - but at that point, the original PUSH was already processed, may have played a sound etc.

r10s:
> We can filter out Auto-submitted messages, but there is no way for the server to know if the message is a webxdc status update or not currently I think.

i just had a look: webxdc status update as well have `Auto-Submitted: auto-generated`

Simon-Laux:
> with chance, deltachat-ios can later on (by some local wakeup) replace the generic notification by a more informative one silently - but at that point, the original PUSH was already processed, may have played a sound etc.

Isn't what we want silent push or a push message that we can "decrypt" before displaying?
what's wrong with relying on dc to wake up and process the message? did I miss some discussion that we need a loud real push notification?

standard background might be too unreliable and disabled on low battery, is it that?

some links:
- replace remote notifications: https://developer.apple.com/documentation/usernotifications/modifying-content-in-newly-delivered-notifications, might need some core api to download only the new message with core, the whole background fetch for all accounts will probably take too long for apple/iOS in many cases (Attachments, slow connection)
- (iOS 14+), support focus and contact profile image and calls: what we should eventually implement in the future: https://developer.apple.com/documentation/usernotifications/handling-communication-notifications-and-focus-status-updates

PR #205, opened 2024-01-31
tests: add test that we don't leak email addresses via VRFY

missytake:
It seems we already return 252 for every VRFY command, so we neither confirm nor deny that a mailbox exists on the server. This is the desired behavior. I nonetheless added a test to avoid regressions.

This means our postfix is sufficiently hardened, I think?

replaces #168

### DIFF
--- a/cmdeploy/src/cmdeploy/tests/online/test_0_login.py
+++ b/cmdeploy/src/cmdeploy/tests/online/test_0_login.py
@@ -1,6 +1,7 @@
 import pytest
 import threading
 import queue
+import socket
 
 from chatmaild.config import read_config
 from cmdeploy.cmdeploy import main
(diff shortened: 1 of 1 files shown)


missytake:
> Anything speaks against setting https://www.postfix.org/postconf.5.html#disable_vrfy_command anyway and verifying that an error is returned rather than the same "252 2.0.0" response?

The 252 response *is* the error. See https://www.rfc-editor.org/rfc/rfc2821#section-7.3:

"If a site disables these commands for security reasons, the SMTP server MUST return a 252 response, rather than a code that could be confused with successful or unsuccessful verification.

Returning a 250 reply code with the address listed in the VRFY command after having checked it only for syntax violates this rule. Of course, an implementation that "supports" VRFY by always returning 550 whether or not the address is valid is equally not in conformance."

PR #215, opened 2024-02-14
run expunge cronjob as correct user

missytake:
This also adds a test which admins can run to verify the cronjob is working correctly. I'm a bit conflicted about the test as it doesn't verify if a system works correctly *right now*, only that it worked correctly last night... but it will work on staging at least, as staging isn't old enough yet for the bug to appear :upside_down_face: 

fix https://github.com/deltachat/chatmail/issues/210


hpk42 (approved):
nice!

PR #233, opened 2024-03-10
expunge: reset maildirsize after expunging old mails

### LINKED ISSUE #232: Quota is not updated when old messages are removed
Current cronjob for cleaning old mails removes messages from maildir directly:
https://github.com/deltachat/chatmail/blob/4bca7891a2b2214c81a708bf4202dfea8c30f3a3/cmdeploy/src/cmdeploy/dovecot/expunge.cron.j2

This does not update the quota automatically according to https://doc.dovecot.org/configuration_manual/quota/quota_maildir/:
> You can’t rely on Dovecot noticing external changes to Maildir and updating maildirsize accordingly. This happens eventually when quota is being recalculated, but it may take a while. Quota recalculation also currently doesn’t trigger quota warning executions.

A way to fix this would be to remove messages via `doveadm`:
https://serverfault.com/questions/769208/doveadm-delete-messages-older-than-date

missytake:
closes #232

### DIFF
--- a/cmdeploy/src/cmdeploy/dovecot/expunge.cron.j2
+++ b/cmdeploy/src/cmdeploy/dovecot/expunge.cron.j2
@@ -8,3 +8,4 @@
 # or only temporary (but then they shouldn't be around after {{ config.delete_mails_after }} days anyway).
 2 0 * * * vmail find /home/vmail/mail/{{ config.mail_domain }} -path '*/tmp/*' -mtime +{{ config.delete_mails_after }} -type f -delete
 2 0 * * * vmail find /home/vmail/mail/{{ config.mail_domain }} -path '*/.*/tmp/*' -mtime +{{ config.delete_mails_after }} -type f -delete
+3 0 * * * vmail find /home/vmail/mail/{{ config.mail_domain }} -name 'maildirsize' -type f -delete


cmdeploy/src/cmdeploy/dovecot/expunge.cron.j2
```diff
@@ -8,3 +8,4 @@
 # or only temporary (but then they shouldn't be around after {{ config.delete_mails_after }} days anyway).
 2 0 * * * vmail find /home/vmail/mail/{{ config.mail_domain }} -path '*/tmp/*' -mtime +{{ config.delete_mails_after }} -type f -delete
 2 0 * * * vmail find /home/vmail/mail/{{ config.mail_domain }} -path '*/.*/tmp/*' -mtime +{{ config.delete_mails_after }} -type f -delete
+3 0 * * * vmail find /home/vmail/mail/{{ config.mail_domain }} -name 'maildirsize' -type f -delete
```
hpk42, line 11: alex wrote in the issue: 
> Another thing that I just tested is simply removing maildirsize file, it worked and resulted in quota update. It is not nice because the quota has to be recalculated, but if we only do it when removing messages, may be fine.

It's probably not a big deal but i think writing a python script (maybe chatmaild.expunge_messages) and only removing maildirsize if files were actually deleted should be easy enough and also goes a bit of the way of making this per-user configurable maybe for 1.46.

PR #239, opened 2024-03-20
add persistent metadata/token storage

### LINKED ISSUE #238: Persist push notification tokens in imap metadata process
Right now, https://github.com/deltachat/chatmail/blob/main/chatmaild/src/chatmaild/metadata.py does not persist tokens so any reload of the metadata unit will wipe all tokens and it then takes a while, and ios devices will meanwhile not get woken up again ... only when they are started again will they re-register and populate the metadata.py ram-storage.

hpk42:
fixes #238 

note that i use "marshal", the python interpreter fast marshalling of basic data types. A dict[string->string] is pretty safe to marshal and demarshal, so nothing more fancy is needed.  Also tested that python3.6 and python3.10 can read/write the marhsalled data.

### DIFF
--- a/chatmaild/src/chatmaild/chatmail-metadata.service.f
+++ b/chatmaild/src/chatmaild/chatmail-metadata.service.f
@@ -2,7 +2,7 @@
 Description=Chatmail dict proxy for IMAP METADATA
 
 [Service]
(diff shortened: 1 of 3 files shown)


missytake (approved):
Looks good!

Issue #235, opened 2024-03-19
CI fails at cmdeploy test


hagenest:
4 cmdeploy tests are currently failing. The last commit that did pass was e6a3fab6aac6f7830d0666bf829ede8b895121cc.

The failing tests are:
- [x] test_2_deltachat.py::test_echobot
- [x] TestEndToEndDeltaChat::test_read_receipts_between_instance
- [x] TestEndToEndDeltaChat::test_exceed_quota: https://github.com/deltachat/chatmail/pull/245
- [x] test_2_deltachat.py::test_hide_senders_ip_address
- [x] tests using CHATMAIL_DOMAIN2

hpk42:
could you mention your invocation line and attach log of the test run including all the failures?

hagenest:
well, the last CI run is here and includes invocation and logs: https://github.com/deltachat/chatmail/actions/runs/8322003073/job/22769103076

The invocation was `venv/bin/pytest cmdeploy/src/ -n4 -rs -x -vrx --durations=5 --slow`

missytake:
The `CHATMAIL_DOMAIN2` tests seem to work :) I tried it on the #245 branch, with c2.testrun.org:

```
$ CHATMAIL_DOMAIN2=staging.testrun.org pytest --slow; beep
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.9.18, pytest-8.0.1, pluggy-1.4.0
Deltachat core=v1.136.2 sqlite=3.39.4 journal_mode=wal
rootdir: /home/user/code/chatmail
plugins: cmdeploy-0.2, deltachat-1.136.2, deltachat-rpc-client-1.133.2, xdist-3.5.0
collected 58 items                                                                                                                                                                                                                          

chatmaild/src/chatmaild/tests/test_config.py ..                                                                                                                                                                                       [  3%]
chatmaild/src/chatmaild/tests/test_doveauth.py ........                                                                                                                                                                               [ 17%]
chatmaild/src/chatmaild/tests/test_filtermail.py ........                                                                                                                                                                             [ 31%]
chatmaild/src/chatmaild/tests/test_metadata.py ........                                                                                                                                                                               [ 44%]
chatmaild/src/chatmaild/tests/test_metrics.py .                                                                                                                                                                                       [ 46%]
chatmaild/src/chatmaild/tests/test_newmail.py ..                                                                                                                                                                                      [ 50%]
[...]

PR #242, opened 2024-03-21
fix some test and formatting/ruff issues

hpk42:
addresses #235 a default `cmdeploy test` without the env-var passes  on c1 

Does not fully resolve test-runs as specifying CHATMAIL_DOMAIN2 to c1 or nine fails some tests because currently nine/c1 reject DKIM-sigs from c2-accounts.


missytake (approved):
ah, we need to use the `_evtracker` for `wait_next_incoming_message` now? good catch

PR #243, opened 2024-03-24
Store raw tokens instead of dictionaries in metadata

link2xt:
Follow-up to #239 before we deploy it to nine

### DIFF
--- a/chatmaild/src/chatmaild/metadata.py
+++ b/chatmaild/src/chatmaild/metadata.py
@@ -1,6 +1,6 @@
 import pwd
 
-import pathlib
+from pathlib import Path
 from queue import Queue
 from threading import Thread
 from socketserver import (
@@ -13,7 +13,6 @@ import sys
 import logging
 import os
 import requests
-import marshal
 
 
 DICTPROXY_LOOKUP_CHAR = "L"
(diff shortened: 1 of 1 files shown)


hpk42:
If we want to evolve per-user settings (eg when to expire/remove messages)
then we will need more than just a single setting. 
Also, having a dictionary quite directly maps imap key-value metadata
so i think the simplification here is not that useful
and requires more work if we add more keys.

hpk42:
On Sun, Mar 24, 2024 at 11:25 -0700, link2xt wrote:
> `SETMETADATA` requests can be sent concurrently, in which case we will overwrite the whole dictionary.

Wasn't there something with dovecot that serializes all metadata-operations or do i misremember? 
If indeed we have concurrency then the current way is insufficient
and i think going for guid-dir and store files like `token` etc. is
the easiest as it also makes accessing it from other processes trivial. 
I btw tried to find a way to use the vmail-dir of a user but couldn't easily find one.

hpk42:
meanwhile @link2xt and me discussed to store metadata-settings into a directory structure with each setting being a separate file -- preferably living inside the per-user maildir

hpk42:
it seems that both SETMETADATA and GETMETADATA transmit the e-mail address to our dict-proxy, 
in the begin-transaction and in the lookup, respectively. 
```
B13\tlkj123poi@c2.testrun.org
S13\tpriv/43f5f508a7ea0366dff30200c15250e3/devicetoken\tl1kj23lk123
C13

Lpriv/43f5f508a7ea0366dff30200c15250e3/devicetoken\tlkj123poi@c2.testrun.org
```
so we probably don't need any hacks.

hpk42:
#254 is merged, closing this

PR #245, opened 2024-03-25
tests: fix quota test after log line changed

missytake:


hpk42 (approved):
Looks good.but I'd let the test accept both spellings because we might disable sieve sometime and then need to rediscover the issue

PR #258, opened 2024-03-28
re-enable running the CI in pull requests, but not concurrently

### LINKED ISSUE #246: Run CI for every commit in a PR, but only one CI run at a time
Only running the CI after merging a PR leads to problems like #235. Instead we should:

- run the CI on every PR commit, but ensure that only one workflow can run during the same time
- re-set the staging.testrun.org VPS during the workflow (except acme + dkim state)

missytake:
fix #246

### DIFF
--- a/.github/workflows/test-and-deploy.yaml
+++ b/.github/workflows/test-and-deploy.yaml
@@ -4,12 +4,17 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


missytake:
The CI fails because of https://github.com/deltachat/chatmail/issues/217

PR #261, opened 2024-03-29
Run chatmail-metadata and doveauth as vmail

### LINKED ISSUE #260: chatmail-metadata and doveauth are running as root
It should run as `vmail`.
Because it currently runs as root, the result is that files it creates are owned by `root`:
```
root@nine /home/vmail/mail/nine.testrun.org # ls -lah | grep pending
drwxr-xr-x  2 root  root  4.0K Mar 29 05:17 pending_notifications
```

Same problem with `/home/vmail/passdb.sqlite`, it is owned by root.

We can solve the problem by running `chown -R vmail:vmail /home/vmail` on every deploy and changing the `User` in systemd units.

link2xt:
Also sandbox chatmail-metadata

Fixes #260

### DIFF
--- a/cmdeploy/src/cmdeploy/service/chatmail-metadata.service.f
+++ b/cmdeploy/src/cmdeploy/service/chatmail-metadata.service.f
@@ -2,9 +2,11 @@
 Description=Chatmail dict proxy for IMAP METADATA
 
 [Service]
-ExecStart={execpath} /run/dovecot/metadata.socket vmail /home/vmail/mail/{mail_domain}
+ExecStart={execpath} /run/chatmail-metadata/metadata.socket /home/vmail/mail/{mail_domain}
 Restart=always
 RestartSec=30
+User=vmail
+RuntimeDirectory=chatmail-metadata
 
 [Install]
 WantedBy=multi-user.target
--- a/cmdeploy/src/cmdeploy/service/doveauth.service.f
+++ b/cmdeploy/src/cmdeploy/service/doveauth.service.f
@@ -2,9 +2,11 @@
 Description=Chatmail dict authentication proxy for dovecot
 
 [Service]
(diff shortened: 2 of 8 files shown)


cmdeploy/src/cmdeploy/service/chatmail-metadata.service.f
```diff
@@ -5,6 +5,41 @@
 ExecStart={execpath} /run/dovecot/metadata.socket vmail /home/vmail/mail/{mail_domain}
 Restart=always
 RestartSec=30
+User=vmail
+CapabilityBoundingSet=
```
hpk42, line 9: please add a comment here with the invocation that created these auto-generated security settings (i presume they are). 
Also i wonder if these security-settings shouldn't be done in a separate PR applied to all services?

cmdeploy/src/cmdeploy/service/chatmail-metadata.service.f
```diff
@@ -5,6 +5,41 @@
 ExecStart={execpath} /run/dovecot/metadata.socket vmail /home/vmail/mail/{mail_domain}
 Restart=always
 RestartSec=30
+User=vmail
```
hpk42, line 8: if we do this, do we still need to pass in `vmail` to the metadata script and do the "chown" shenenigans in https://github.com/deltachat/chatmail/blob/main/chatmaild/src/chatmaild/metadata.py ?

cmdeploy/src/cmdeploy/service/doveauth.service.f
```diff
@@ -5,6 +5,7 @@
 ExecStart={execpath} /run/dovecot/doveauth.socket vmail /home/vmail/passdb.sqlite {config_path}
 Restart=always
 RestartSec=30
+User=vmail
```
hpk42, line 8: same here, doveauth.py probably does not need "chown" anymore?

hpk42 (approved):
lgtm, nice

PR #263, opened 2024-03-29
better preserve notification order, using a queue again

hpk42:
turns out it's just a few lines change and notification order is preserved and it's also better on IO a bit (but not a problem in any case)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged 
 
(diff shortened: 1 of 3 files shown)


missytake:
CI fails because it is not rebased on top of https://github.com/deltachat/chatmail/pull/262

missytake:
CI fails for the right reason now (I think).

hpk42:
closed in favor of #265

PR #264, opened 2024-03-30
Sandbox chatmail-metadata

link2xt:
# Based on PR #261


hpk42:
if systemd-level isolation is worthwhile, i think we need a systematic approach for all our systemd-services. 
Bit it probably makes sense to discuss security/hardening a bit more generally -- at least all our custom systemd-services now run under user-priviliges and not root anymore.   Closing this particular metadata-systemd pr for now.

PR #269, opened 2024-04-10
Document email authentication requirements

link2xt:


missytake (approved):
It's a good summary, but it requires high knowledge about how authentication in the email system works. Maybe we could put a small paragraph in the beginning which requires what the goal of our configuration is? Then it is easier to understand the necessary pre-knowledge. Maybe:

```
Chat users expect that if they receive a message from `alice@example.org`
the message was actually sent
from someone with the passphrase to the `alice@example.org` account.
But the email system
[doesn't guarantee this](https://proton.me/blog/what-is-email-spoofing).
With these best practices we try to improve on this situation:
```

Issue #249, opened 2024-03-25
Build deb with dovecot delay patch


hagenest:
We need a deb with dovecot with the patch for #72 included.

I first tried doing this at build.opensuse.org - but building debs with tons of additional backports from debian doesn't seem to be supported there - or at least not at all documented.

We now have (at least) 2 options:
- Build in the OBS anyway, but with dovecot-main + patch
  - Pro: We automatically get a PPA, relatively easy
  - Con: We build a bleeding-edge package for older dependencies in debian, this could lead to much more work in the long run
- Build the deb in the GitHub CI => Host a PPA on page.deltachat
  - Pro: We can easily fork the package from Debian
  - Con: We have to provide a PPA by ourself.

IMO option 2 would be the preferable option, at least until we decide to move chatmail to nixos :laughing: 
Will implement the building part now, we have to see how we best provide a repository then

hagenest:
any ideas how we best provide a repository?

hagenest:
This may sound stupid, but after your last message I searched again for some kind of documentation and now I got it. I tried to build the package like described here: https://www.sidorenko.io/post/2016/05/build-packages-with-openbuild-service/

This did not work - but I now found this page (which for some reason doesn't seem indexed by search engines): https://en.opensuse.org/openSUSE:Build_Service_Debian_builds

And this explanation actually works :) So yeah, I'm now all in favor of the first option again.
Apologies for being very confidently incorrect when opening this issue :sweat_smile:

hagenest:
Hooray, we have a deb and repo: https://download.opensuse.org/repositories/home:/deltachat/Debian_12/

https://build.opensuse.org/package/show/home:deltachat/

Here is the explanation for adding the repo, I'll add instructions tomorrow: https://en.opensuse.org/openSUSE:Build_Service_Debian_builds#Adding_repository_signing_key

hpk42:
https://build.opensuse.org/package/show/home:deltachat/ gives 404
and i still don't quite get what is the source for https://download.opensuse.org/repositories/home:/deltachat/Debian_12/

the input flowing into the build-process there should be committed into a separate "deltachat/debian-builds" one  with a proper README -- along with being explicit what kind of permissions/passwords are neccessary. 
Ideally, a 3rd party should be able to build their own debian-packages (without opensuse) and use that for their chatmail service.

hagenest:
[...]

PR #270, opened 2024-04-11
Add OBS repository for dovecot without delay

### LINKED ISSUE #72: 500 ms delay between sending a message over SMTP and getting IDLE notified
I have measured round trip time sending messages between two accounts: https://github.com/deltachat/deltachat-core-rust/pull/4974#issuecomment-1806517224

It takes roughly 1.1 seconds to send a message to the echo bot and receive a reply with two Delta Chat accounts. Slightly more than 0.5 seconds (516 or 517 ms) is spent between a mail is sent over SMTP and Dovecot notifies the client in IDLE mode that the message has arrived.

There is a suspicious constant `#define NOTIFY_DELAY_MSECS 500` in Dovecot `mailbox-watch.c`:
https://github.com/dovecot/core/blob/93a53a9d590f0220de28600f36a969cc38c3148b/src/lib-storage/mailbox-watch.c#L12

`mailbox_watch_add` is called from storage implementation. We are using Maildir (may consider switching to dbox, but Maildir is good enough), so in our case it is here:
https://github.com/dovecot/core/blob/93a53a9d590f0220de28600f36a969cc38c3148b/src/lib-storage/index/maildir/maildir-storage.c#L632-L645

Here is a `mailbox_notify_changes` which dynamically dispatches to a particular storage implementation:
https://github.com/dovecot/core/blob/93a53a9d590f0220de28600f36a969cc38c3148b/src/lib-storage/mail-storage.c#L2473-L2484

And finally, here is IMAP IDLE command implementation using `mailbox_notify_changes` to subscribe for notifications:
https://github.com/dovecot/core/blob/93a53a9d590f0220de28600f36a969cc38c3148b/src/imap/cmd-idle.c#L289

So it seems `NOTIFY_DELAY_MSECS 500` directly affects how IDLE callback is called when storage notices changes. And it corresponds with the measurements, it is always slightly more than 500 ms delay.

And here is a commit adding it 14 years ago:
https://github.com/dovecot/core/commit/56fb5d09955b6097f77b341717fd9b70e9f13e7a

It reads:
> mailbox_notify_changes(): Delay sending notifications for 500 msecs.
>
> If the notification is done immediately, IDLE may not notice the change
> because it's not finished yet.

hagenest:
This PR adds our new OBS repository for our patched dovecot version. Closes #72 and #249.


cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -522,6 +522,25 @@ def deploy_chatmail(config_path: Path) -> None:
...
+        src = importlib.resources.files(__package__).joinpath("obs-home-deltachat.gpg"),
+        dest = "/etc/apt/keyrings/obs-home-deltachat.gpg",
+        user="root",
+        group="root",
+        mode="644",
+    )
+
+    files.line(
+        name = "Add DeltaChat OBS home repository to sources.list",
+        path = "/etc/apt/sources.list",
```
missytake, line 537:
```suggestion
        path = "/etc/apt/sources.list.d/opensuse-build-service.list",
```

Let's move this to the subdirectory :) that's what it's good for.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -522,6 +522,25 @@ def deploy_chatmail(config_path: Path) -> None:
         packages="postfix",
     )
 
+    # Add our OBS repository for dovecot_no_delay
+    files.put(
+        name = "Add Deltachat OBS GPG key to apt keyring",
+        src = importlib.resources.files(__package__).joinpath("obs-home-deltachat.gpg"),
+        dest = "/etc/apt/keyrings/obs-home-deltachat.gpg",
+        user="root",
+        group="root",
```
missytake, line 531: I think `user` and `group` can be left out? by default it should be root already, as the root user is used for the deploy.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -522,6 +522,25 @@ def deploy_chatmail(config_path: Path) -> None:
...
+        mode="644",
```
missytake, line 532: Is this necessary? depends on root's umask, I guess. But doesn't hurt.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -522,6 +522,25 @@ def deploy_chatmail(config_path: Path) -> None:
...
+    )
+
+    files.line(
+        name = "Add DeltaChat OBS home repository to sources.list",
+        path = "/etc/apt/sources.list",
+        line = "deb [signed-by=/etc/apt/keyrings/obs-home-deltachat.gpg] https://download.opensuse.org/repositories/home:/deltachat/Debian_12/ ./",
+        ensure_newline = True,
+    )
+
+    apt.update(name="apt update")
```
missytake, line 542: We do an `apt update` in line 486 anyway, maybe leave this out and just move the two other operations above that?

missytake (approved):
Wow! I just logged into staging.testrun.org, messaged echo, and it came back immediately. Also really nice that we can now use staging.testrun.org to preview PRs.

hagenest:
ARM users take the build for ARM. The OBS documentation was outdated, there was a runner for ARM Debian the entire time

cmdeploy/src/cmdeploy/__init__.py
hagenest: well, I just copied it from the code below, for consistency :D don't know if it was necessary there

cmdeploy/src/cmdeploy/__init__.py
hagenest: done

cmdeploy/src/cmdeploy/__init__.py
missytake: Yeah, then let's just merge it and try out in a separate PR if we can leave these out.

cmdeploy/src/cmdeploy/__init__.py
missytake: let's just merge it and try out in a separate PR if we can leave these out.

hpk42:
Great but changelpg entry is missing

hagenest:
oops, forgot. pushed a changelog to main, hope that's ok @hpk42

PR #278, opened 2024-05-02
add XCHATMAIL marker so delta chat can recognize it's a chatmail server

hpk42:
helps https://github.com/deltachat/deltachat-core-rust/issues/5380


hpk42:
i added a test and then merged

PR #284, opened 2024-05-05
metadata: add support for /shared/vendor/deltachat/irohrelay

link2xt:
This is needed for peerchannels: https://github.com/deltachat/deltachat-core-rust/pull/5346

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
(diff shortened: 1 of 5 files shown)


hpk42 (commented):
looks good but better add a test to `test_metadata.py` -- currently only the iroh_relay == None path is tested.  and it would be good to see the metadata keys spelled out.

hpk42 (approved):
to be clear: i am fine to merge it but prefer there to be a test.

PR #286, opened 2024-05-06
Configure more lints and switch from `black` to `ruff format`

link2xt:


hpk42 (approved):

hpk42:
looks good to me but might mess now with other PRs?

PR #289, opened 2024-05-07
Ok, this is the important step to hare the server.

nudeldudel:
The French instance https://acesttoi.fr has been hardened in this way from the beginning, and it was necessary .

To use cmdeploy probably, using of ~/.ssh/conifg seams to be sensfull, and using ssh with key not with password is recommended.


hpk42:
this is not a complete PR -- did you manually install those files after you ran "cmdeploy run"?

nudeldudel:
I won't change cmddeploy.

There are some changes, that should be make during the installationen. At
least the change of the ssh-port is very important.

This change can be done by 'tee' or something else over the command - line.
The path should usualy be: /etc/ssh/sshd_config. If dropbear is used and
installed correctly, the changes in this file should be produced, too.

But there is - still - the problem with the use of ssh by cmddeploy. After
the the change, the new port is to give on every connection by using the
parameter -p 220XX with ssh.

As I don't want to delitate into cmddeploy, I only pushed the sshd_config
with in the changes to the original. So the change-protokol is more
important then the file itself.



holger krekel ***@***.***> schrieb am Sa., 11. Mai 2024,
22:52:

hpk42:
I don't understand how you suggest to merge this, sorry. 
This repo aims to make things automatically be setup via "cmdeploy" commands. 
We try to keep the "out of cmdeploy" tasks to a minimum -- 
currently the DNS settings and adapting the web page. 
So if we want to harden ssh in some way, 
i am afraid we need to think how to integrate it.

nudeldudel:
Ok, I am not so fit in find and replace by command line, but will try to
add a new command to cmdeploy which calls "harden" to do theses things
additionally if they are wanted.



holger krekel ***@***.***> schrieb am So., 12. Mai 2024,
18:22:

Issue #287, opened 2024-05-06
Be more clear about root password during setup


roughnecks:
HI,

not really a bug, but I wouldn't know where else to post this.

When you deploy chatmail it asks for root's password. It should be mentioned in the setup guide that it's not the root account password, but its private key's password - and so, that you have to previously set up an ssh key for the root account on the remote server.

hagenest:
has been in adressed #291

PR #291, opened 2024-05-17
Recommend authentication via ssh key with ed25519 algorithm (#231)

### LINKED ISSUE #231: Recommend ed25519 keys in the README
This is needed due to a bug https://github.com/paramiko/paramiko/issues/2191

hagenest:
this should fix #231 (and arguably #287) or would you prefer a longer explanation? I felt a long paragraph would be overkill in comparision to the rest of the short documentation.

I haven't written into the changelog, as this is doesn't touch the code, is this the right way?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ after which the initially specified password is required for using them.
 
 ## Deploying your own chatmail server 
 
-To deploy chatmail on your own server, you must have set-up ssh authentication and need to use an ed25519 key, due to an [upstream bug in paramiko](https://github.com/paramiko/paramiko/issues/2191). You also need to add your private key to the local ssh-agent to avoid having to type the password for your key multiple times.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,6 +15,8 @@ after which the initially specified password is required for using them.
 
 ## Deploying your own chatmail server 
 
+To deploy chatmail on your own server, you must have set-up ssh authentication and need to use an ed25519 key, due to an [upstream bug in paramiko](https://github.com/paramiko/paramiko/issues/2191). You also need to add your private key to the local ssh-agent to avoid having to type the password for your key multiple times.
```
missytake, line 18:
```suggestion
To deploy chatmail on your own server, you must have set-up ssh authentication and need to use an ed25519 key, due to an [upstream bug in paramiko](https://github.com/paramiko/paramiko/issues/2191). You also need to add your private key to the local ssh-agent, because you can't type in your password during deployment.
```

Otherwise some might think "oh, I'm fine with entering my password multiple times, whatever".

hagenest: makes sense, didn't know it wasn't possible at all, did not want to test it :D

missytake (approved):
Nice :)

PR #296, opened 2024-05-18
DKIM-sign Content-Type and oversign all signed headers

link2xt:
Oversigning (including header name in DKIM-Signature
more times that it appears in the mail) prevents
adding more headers with the same name
without invalidating DKIM signature.

We don't want middleboxes to insert a second From header,
adding Cc field to mails that don't have one etc.


missytake (approved):
Does it make sense to write a test for this?

PR #300, opened 2024-05-18
change delete_mails_after default to 20

### LINKED ISSUE #298: delete_mails_after = 40 is too high
@hagenest just ran into the 100MB limit.

Also, GDPR has a frist of 1 month, which is why it makes sense to keep mails less than that if we want to have a minimalist privacy policy. Maybe 20 days?

Should this be the new default or should we just change it for nine.testrun.org?

hagenest:
closes #298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- change default for delete_mails_after from 40 to 20 days
+  ([#300]https://github.com/deltachat/chatmail/pull/300)
+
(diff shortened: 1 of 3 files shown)


hagenest:
yeah, turns out the test values need to changed as well, lol

missytake:
oops, I meant to rebase. Let me fix this

missytake (approved):
Looks good :) only a changelog item is missing. I think it makes sense to notify admins of existing instances that we changed the default, but it's on them if they want to follow our new recommendation.

hagenest:
true, don't know, why I forgot. done

PR #304, opened 2024-05-18
Move cronjobs to systemd-timer

### LINKED ISSUE #282: METRICS doesn't work
As I installed the new instance, acesttoi.fr, metrics doesn't work. I did no changes to the main branch. 

Browser shows "nginx 404"….

hagenest:
closes #282

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- run metrics generation with systemd-timer instead of cron
+  ([#304](https://github.com/deltachat/chatmail/pull/304))
+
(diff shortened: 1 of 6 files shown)


hagenest:
not working yet, https://staging.testrun.org/metrics returns a 404

hagenest:
so, this is currently deployed on c2.testrun.org and it does not work - the metrics are in the journal, but the metrics file doesn't get created. any ideas what I did wrong?

hagenest:
It's possible to complete this someday, if we decide that we don't want to use cron anymore. For now https://github.com/deltachat/chatmail/pull/310 should fix the issue too

PR #312, opened 2024-05-30
Add config option for ipv6 usage

### LINKED ISSUE #306: Steps for Chatmail when IPv6 support is disabled in OS
Hello,

We had a situation with our VPS's host because they aren't letting customers set a PTR for IPv6.
So the choice was pretty simple, disable IPv6 entirely on chatmail.

First I edited `/etc/default/grub` and added `ipv6.disable=1` to:

```
GRUB_CMDLINE_LINUX_DEFAULT=
GRUB_CMDLINE_LINUX=
```

while preserving the other settings. Just add the new option at the end of those configurations (space separated options)

Then I rebooted the server and noticed I couldn't connect anymore to chatmail.
Dovecot and nginx were down because both tried to listen to the now unsupported IPv6 address family.

Fix for nginx is easy, just remove the two `listen [::]:443` lines in /etc/nginx/nginx.conf and restart the service.
Fix for Dovecot was to add a new line to /etc/dovecot/dovecot.conf so that it only listens to IPv4:

`listen = *`

Last, when I tried to run `mailq` I also got this warning from Postfix, which I hadn't noticed before:

```
mailq: warning: inet_protocols: disabling IPv6 name/address support: Address family not supported by protocol
postqueue: warning: inet_protocols: disabling IPv6 name/address support: Address family not supported by protocol
Mail queue is empty
```

To fix that I edited /etc/postfix/main.cf, changing from:

`inet_protocols = all`

to

`inet_protocols = ipv4`

Again, restart dovecot and postfix when done.
Everyone seem extremely silent this evening on our server, but I believe everything is working alright, DC says it's connected and mailq is empty after sending a few messages myself..

hagenest:
fixes #306


hagenest:
ok, so after testing on c3.testrun.org, it seems that manually disabling ipv6 in grub does lead all tests to fail, even with the config changes. Maybe one of them isn't working right, need to investigate.

hagenest:
TODO: - name to disable_ipv6
- add disclaimer, because our ci doesn't test this

missytake:
Hm, have you seen https://github.com/deltachat/chatmail/compare/main...roughnecks:wpn-chatmail:no-ipv6 ? that's quite an elegant solution... but not conditional on the chatmail.ini variable of course. But maybe we should use jinja templating and the config files instead of meddling with `files.line`?

hagenest:
yeah, templating would be better, I agree.

hagenest:
thing is, I think we have to edit files that aren't part of the templating yet, but probably for the best

missytake:
> thing is, I think we have to edit files that aren't part of the templating yet, but probably for the best

all files in https://github.com/deltachat/chatmail/compare/main...roughnecks:wpn-chatmail:no-ipv6 are jinja2 templates, you can recognize that at the `.j2` file ending

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -375,17 +385,26 @@ def _configure_dovecot(config: Config, debug: bool = False) -> bool:
     return need_restart
 
 
-def _configure_nginx(domain: str, debug: bool = False) -> bool:
+def _configure_nginx(domain: str, debug: bool = False, config = Config) -> bool:
```
hagenest, line 388: I'm not sure what I have to do here to pass config into this function. Adding it like that leads to `AttributeError: 'str' object has no attribute 'disable_ipv6'`

hagenest: @missytake any ideas?

missytake: First of all, I think you mean `config: Config` not `config = Config`; the former is specifying a type, the latter is setting a default value :)

And you need to pass the config to the `_configure_nginx()` function, afaict that doesn't happen yet? So I wonder why `config` is a string object instead of a class object here, that doesn't make much sense.

chatmaild/src/chatmaild/config.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, inipath, params):
         self.passthrough_recipients = params["passthrough_recipients"].split()
         self.filtermail_smtp_port = int(params["filtermail_smtp_port"])
         self.postfix_reinject_port = int(params["postfix_reinject_port"])
+        self.disable_ipv6 = params.get("disable_ipv6")
```
missytake, line 23:
```suggestion
        self.disable_ipv6 = True if params.get("disable_ipv6") == "True" else False
```

Let's make a boolean out of it?

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -317,6 +326,7 @@ def _configure_dovecot(config: Config, debug: bool = False) -> bool:
         mode="644",
         config=config,
         debug=debug,
+        listen_ipv4_only=listen_ipv4_only,
```
missytake, line 329:
```suggestion
        listen_ipv4_only="listen = *" if config.disable_ipv6 else "",
```

Then you can do nice little things like this, and even leave out the 3 lines added above...

cmdeploy/src/cmdeploy/__init__.py
missytake: In general you can take a look at https://github.com/deltachat/chatmail/pull/285/files and how I added a config value there, I think there it worked

hagenest: Problem is, I tried adding config: Config first and that leads to `TypeError: _configure_nginx() missing 1 required positional argument: 'domain'`.
I did it the same way as in the PR you linked, and it works in the other functions in __init__.py.

I can't find any other mention of the _configure_nginx() function in the entire project, so I don't get where it should get this value.

chatmaild/src/chatmaild/config.py
hagenest: Originaly this was a boolean like that - I changed it back to a string because of the weird string related issue above

cmdeploy/src/cmdeploy/__init__.py
missytake: It is called here, with a string: https://github.com/deltachat/chatmail/blob/hagi/%23306-disable-ipv6/cmdeploy/src/cmdeploy/__init__.py#L595
you should pass the whole config here, instead of only the domain. Luckily it's really only called once.

hagenest:
so, deploying the server with the ipv6 works on c2 and I can use it, even after removing ipv6 in hetzner. But many of the tests fail, I suppose they weren't made with ipv6 in mind. I propose that we add the disable ipv6 as a kind of unsupported action instead of rewriting all tests

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -375,17 +387,26 @@ def _configure_dovecot(config: Config, debug: bool = False) -> bool:
     return need_restart
 
 
-def _configure_nginx(domain: str, debug: bool = False) -> bool:
+def _configure_nginx(config: Config, domain: str, debug: bool = False) -> bool:
```
missytake, line 390: Hm, `domain` == `config.mail_domain`, so I think it isn't very pretty to pass it to this method, instead it should be used from the config object in the whole method. Can you change this?

missytake:
As about half (I think?) of existing chatmail instances have no ipv6, not sure if calling it "unsupported" makes sense. We could have a second CI machine which has no IPv6; it's resource-wise not very nice, sure, but maybe it makes sense here? @hpk42

cmdeploy/src/cmdeploy/__init__.py
hagenest: done

cmdeploy/src/cmdeploy/__init__.py
hagenest: done

chatmaild/src/chatmaild/config.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, inipath, params):
         self.passthrough_recipients = params["passthrough_recipients"].split()
         self.filtermail_smtp_port = int(params["filtermail_smtp_port"])
         self.postfix_reinject_port = int(params["postfix_reinject_port"])
-        self.disable_ipv6 = params.get("disable_ipv6")
+        self.disable_ipv6 = True if params.get("disable_ipv6") == "True" else False
```
missytake, line 23:
```suggestion
        self.disable_ipv6 = True if params.get("disable_ipv6").lower() == "true" else False
```

On second thought, we can't expect admins to case their variables correctly.

missytake (approved):
Great!

missytake:
lint still complains, other than that it could be merged and given to the community.

missytake:
> As about half (I think?) of existing chatmail instances have no ipv6, not sure if calling it "unsupported" makes sense. We could have a second CI machine which has no IPv6; it's resource-wise not very nice, sure, but maybe it makes sense here?

We decided for now to setup a second CI instance for this.

hagenest:
I suppose dropping IP adress checks like suggested in [#336](https://github.com/deltachat/chatmail/issues/336) might affect this?

hpk42:
hum, the original issue #306 addressed by this PR is not really a valid issue anymore, no?

hpk42:
could you describe in user-facing terms what precisely "disable-ipv6" delivers? maybe best in a changelog entry that is missing btw.

missytake:
> could you describe in user-facing terms what precisely "disable-ipv6" delivers? maybe best in a changelog entry that is missing btw.

done :)

hpk42:
still wondering if there is any motiviation besides https://github.com/deltachat/chatmail/issues/306 
for adding a new config setting. See also https://github.com/deltachat/chatmail/issues/306#issuecomment-2254562282 in particular.  IOW, i am a bit wary of doubling CI resources and adding a config setting when the need could be addressed differently, after all.  I do think chatmail servers should work on ipv4-only or ipv6-only hosts, though.

missytake:
deltachat.jurkiewicz.chat and chatmail.woodpeckersnest.space don't have an AAAA record, bcc.chat got one meanwhile. New admins who don't have an IPv6 will show up for sure. I'm not 100% certain that dovecot (and nginx?) will fail if no IPv6 address is available, it probably also depends on the reason; but it's what caused #306.

I mean, we could re-introduce detecting IP addresses and do it automagically instead of introducing a config option :upside_down_face: 

Maybe I'm just frustrated because a lot of effort already went into this PR^^ that would be a bad reason to merge it of course. I never encountered the problem myself, this PR mostly did woodpeckersnest's fixes in a way that we can have it as a config option, that's the reason why we think it works (and the green CI of course). But there is proof that it fixed something for somebody in the past :upside_down_face:

hpk42:
k, also discussed with link2xt who also thinks it makes sense to merge this. 
So let's merge this -- changelog needs a fix.

PR #316, opened 2024-06-03
DNS: don't check DNS on cmdeploy init anymore

### LINKED ISSUE #288: don't check DNS records with cmdeploy init
even "cmdeploy init" which should be a purely local command now does dns checking. 
i think only "run" should do it implicitely (if even that) and definitely show a message before starting dns checks -- otherwise there is an unexplianed multi-second delay as if it's waiting for stdin to provide something or so.

missytake:
fix #288

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- don't check necessary DNS records on cmdeploy init anymore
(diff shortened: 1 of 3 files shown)


hagenest:
ah, I'm stupid. I was looking at this yesterday and thought the import statement in cmdeploy.py was the culprit and was wondering how to best refactor this :upside_down_face:

PR #320, opened 2024-06-11
Improve filtermail checks for encrypted messages

link2xt:
Ensure that first part only contains "Version: 1"
and second part only contains base64 payload
enclosed in "-----BEGIN PGP MESSAGE-----"
and "-----END PGP MESSAGE-----".

Second commit removes support for unencrypted MDNs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## untagged
 
-- improve filtermail checks for encrypted messages
+- improve filtermail checks for encrypted messages and drop support for unencrypted MDNs
   ([#320](https://github.com/deltachat/chatmail/pull/320))
 
 ## 1.3.0 - 2024-06-06
--- a/chatmaild/src/chatmaild/filtermail.py
+++ b/chatmaild/src/chatmaild/filtermail.py
@@ -66,34 +66,6 @@ def check_encrypted(message):
     return True
 
 
-def check_mdn(message, envelope):
-    if len(envelope.rcpt_tos) != 1:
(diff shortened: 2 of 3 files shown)


chatmaild/src/chatmaild/filtermail.py
```diff
@@ -23,43 +30,40 @@ def check_encrypted(message):
...
-    # and then MDNs are just encrypted mail and can pass the border
-    # to other instances.
+            suffix = "\r\n-----END PGP MESSAGE-----\r\n\r\n"
+            if not payload.endswith(suffix):
+                return False
+            payload = payload.removesuffix(suffix)
 
+            for line in payload.split("\r\n"):
+                if NONBASE64_REGEX.match(line):
+                    return False
```
hpk42, line 63: this could be a multi-line regex match to avoid splitting and iterating the lines.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- improve filtermail checks for encrypted messages and drop support for unencrypted MDNs
```
hagenest, line 5: Correct me, if I'm wrong, but this does affect read receipts on chatmail, right? Maybe we should make this clearer, before we're getting bug reports?
Can't really tell, if most of our users know what MDN means - I certanily didn't a few months ago.

hagenest (approved):

hpk42 (approved):
lgtm -- i assume the staging CI deploy tests fail for unrelated reasons.

PR #321, opened 2024-06-11
Reject DKIM signatures that do not cover the whole message body

### LINKED ISSUE #297: Reject messages with partial DKIM signatures
Messages with DKIM-signatures that have `l=` tag are not fully protected. We should reject such signatures.

See `opendkim-lua` man page, there a functions `odkim.sig_bodylength` and `odkim.sig_canonlength` that can help with this. If `sig_canonlength` is less than `sig_bodylength`, then signature should be rejected.

link2xt:
Closes #297

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- Reject DKIM signatures that do not cover the whole message body.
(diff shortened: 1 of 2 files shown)


hpk42 (approved):
lgtm -- test would be nice but at least changelog entry needs to rebased.

PR #323, opened 2024-06-12
Check that OpenPGP has only PKESK and SEIPD packets

link2xt:


missytake (approved):
Nice! The complicated function with a lot of static constants is even documented :))

PR #325, opened 2024-06-19
filtermail: do not allow ASCII armor without actual payload

link2xt:
Last line is removed as "optional checksum",
so it can contain anything.
Make sure that there is at least some actual payload besides this line.

### DIFF
--- a/chatmaild/src/chatmaild/tests/test_filtermail.py
+++ b/chatmaild/src/chatmaild/tests/test_filtermail.py
@@ -167,3 +167,19 @@ UN4fiB0KR9JyG2ayUdNJVkXZSZLnHyRgiaadlpUo16LVvw==\r
 """
 
     assert check_armored_payload(payload) == True
+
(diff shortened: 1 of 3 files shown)


chatmaild/src/chatmaild/tests/test_filtermail.py
```diff
@@ -167,3 +167,19 @@ def test_check_armored_payload():
...
+"""
+    assert check_armored_payload(payload) == False
+
+    payload = """-----BEGIN PGP MESSAGE-----\r
+\r
+=njUN
+-----END PGP MESSAGE-----\r
+\r
+"""
+    assert check_armored_payload(payload) == False
```
missytake, line 185: What about this?

```
-----BEGIN PGP MESSAGE-----\r
\r
-----END PGP MESSAGE-----\r
```

missytake (approved):

hpk42 (approved):
changelog would be nice but also it's minor.

PR #327, opened 2024-06-20
Automatically remove accounts n days after last login

### LINKED ISSUE #295: Remove old accounts
At least `ci-` accounts can be removed. Currently running "Fix file owner in /home/vmail" steps takes a long time because it has to go through a lot of directories.

We also should drop "Fix file owner in /home/vmail" step eventually.

hagenest:
fixes #295

### DIFF
--- a/chatmaild/src/chatmaild/doveauth.py
+++ b/chatmaild/src/chatmaild/doveauth.py
@@ -106,7 +106,7 @@ def lookup_passdb(db, config: Config, user, cleartext_password):
         if userdata:
             # Update last login time.
             conn.execute(
-                "UPDATE users SET last_login=? WHERE addr=?", (int(time.time()), user)
+                "UPDATE users SET last_login=? WHERE addr=?", (round(int(time.time()) // 86400), user) # only save the rounded down date of login
             )
 
             userdata["home"] = f"/home/vmail/mail/{config.mail_domain}/{user}"
--- a/chatmaild/src/chatmaild/doveauth.py
+++ b/chatmaild/src/chatmaild/doveauth.py
@@ -106,7 +106,7 @@ def lookup_passdb(db, config: Config, user, cleartext_password):
         if userdata:
             # Update last login time.
             conn.execute(
-                "UPDATE users SET last_login=? WHERE addr=?", (round(int(time.time()) // 86400), user) # only save the rounded down date of login
+                "UPDATE users SET last_login=? WHERE addr=?", (round(int(time.time()) // 86400), user)
             )
 
             userdata["home"] = f"/home/vmail/mail/{config.mail_domain}/{user}"
--- a/chatmaild/src/chatmaild/doveauth.py
+++ b/chatmaild/src/chatmaild/doveauth.py
@@ -106,7 +106,8 @@ def lookup_passdb(db, config: Config, user, cleartext_password):
         if userdata:
             # Update last login time.
             conn.execute(
-                "UPDATE users SET last_login=? WHERE addr=?", (round(int(time.time()) // 86400), user)
+                "UPDATE users SET last_login=? WHERE addr=?",
+                (round(int(time.time()) // 86400), user),
             )
 
             userdata["home"] = f"/home/vmail/mail/{config.mail_domain}/{user}"
--- a/chatmaild/src/chatmaild/doveauth.py
+++ b/chatmaild/src/chatmaild/doveauth.py
@@ -107,7 +107,7 @@ def lookup_passdb(db, config: Config, user, cleartext_password):
             # Update last login time.
             conn.execute(
                 "UPDATE users SET last_login=? WHERE addr=?",
-                (round(int(time.time()) // 86400), user),
+                (int(time.time() // 86400), user),
             )
 
             userdata["home"] = f"/home/vmail/mail/{config.mail_domain}/{user}"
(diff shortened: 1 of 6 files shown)


chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -20,6 +20,9 @@
 # days after which mails are unconditionally deleted
 delete_mails_after = 20
 
+# days after which accounts are deleted if there were no login
+delete_accounts_after = 30
```
missytake, line 24: maybe 21 could be the default? This is exactly the GDPR limit iirc. Then we wouldn't have to do any manual action if users want to delete their account, we could just tell them to delete it in their app.

hagenest: In that case,let's do 20, like with the mails, I'd say

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -106,7 +106,8 @@ def lookup_passdb(db, config: Config, user, cleartext_password):
         if userdata:
             # Update last login time.
             conn.execute(
-                "UPDATE users SET last_login=? WHERE addr=?", (int(time.time()), user)
+                "UPDATE users SET last_login=? WHERE addr=?",
+                (round(int(time.time()) // 86400), user),
```
hpk42, line 110: the `round` seems superflous

chatmaild/src/chatmaild/ini/chatmail.ini.f
hpk42: the GDPR limit is rather 30 days IIRC, so 25 or so should be fine.

chatmaild/src/chatmaild/rm_accounts.py
```diff
@@ -0,0 +1,54 @@
...
+    UnixStreamServer,
+)
+
+from .config import Config, read_config
+from .database import Database
+
+
+def remove_users(db: Database, cutoff_date: int):
+    db.connect()
+    try:
```
hpk42, line 21: the typical pattern is rather "with db.write_transaction()" like in other places. You don't need explicit committing and finally-clauses then.

chatmaild/src/chatmaild/rm_accounts.py
```diff
@@ -0,0 +1,54 @@
...
+        db.commit_changes()
+    finally:
+        db.close()
+
+
+def remove_user_data(db: Database, cutoff_date: int, dir: Path):
+    """Collects all users where last_login < cutoff_date and deletes corresponding directories."""
+    db.connect()
+
+    try:
```
hpk42, line 33: same here, use with-statement.

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -20,6 +20,9 @@
 # days after which mails are unconditionally deleted
 delete_mails_after = 20
 
+# days after which accounts are deleted if there were no login
```
hpk42, line 23:
```suggestion
# days after which accounts are deleted if nobody logged in
```

chatmaild/src/chatmaild/rm_accounts.py
```diff
@@ -0,0 +1,54 @@
...
+import sys
+import time
+import shutil
+from pathlib import Path
+from socketserver import (
+    StreamRequestHandler,
+    ThreadingMixIn,
+    UnixStreamServer,
+)
+
```
hpk42, line 14: remove unused imports.

chatmaild/src/chatmaild/rm_accounts.py
```diff
@@ -0,0 +1,54 @@
...
+    """Collects all users where last_login < cutoff_date and deletes corresponding directories."""
+    db.connect()
+
+    try:
+        select_query = "SELECT user FROM users WHERE last_login <?"
+        cursor = db.execute_query(select_query, (cutoff_date,))
+        usernames = cursor.fetchall()
+
+        for username in usernames:
+            user_dir = dir / username[0]
```
hpk42, line 39: don't use `dir` -- it's a builtin name and you are shadowing it. 
Also `vmail_basedir` would be more telling.

chatmaild/src/chatmaild/rm_accounts.py
```diff
@@ -0,0 +1,54 @@
...
+            if user_dir.exists() and user_dir.is_dir():
+                shutil.rmtree(user_dir)
```
hpk42, line 41: probably better with "ignore_errors=True"

hpk42 (commented):
apart from the issues mentioned in the comments, also tests are missing.

chatmaild/src/chatmaild/ini/chatmail.ini.f
hagenest: set it to 25

chatmaild/src/chatmaild/doveauth.py
hagenest: true

hpk42 (commented):
PR still  has a  number of problems -- see my comments. 

Note that i am tackling the goal of this PR now myself, because it requires adding tests and touching several code bits to ensure we don't accidently remove users.  Also, we need this PR rather sooner than later because nine needs to be properly cleaned up still.

hpk42:
k, since #350 is merged, this PR is redundant.

PR #329, opened 2024-06-24
fix headline ordering numbers, typo

r10s:
before, the order was 2 - 3.1 - 3.2 - 3
i think, the gist was to have subheadlines under "2."; this is fixed by this PR.

moreover, the PR contains a small typo fix.


missytake (approved):
Thanks, good catch!

PR #335, opened 2024-07-01
Increase number of logged in IMAP sessions to 50000

link2xt:


hpk42 (approved):
lgtm, thanks for investigating and fixing.

hpk42:
yes, let's maybe go for 40K (but also dovecot has a client-limit of 20K, which needs to get increased i think)

PR #342, opened 2024-07-05
Add script to remove old seen messages

link2xt:


cmdeploy/src/cmdeploy/dovecot/remove-seen.py
```diff
@@ -0,0 +1,41 @@
...
+            if seen and now > stat.st_mtime + 2 * 24 * 3600:
+                removed_bytes += size
+                mailpath.unlink(missing_ok=True)
+
+        if removed_bytes > 0:
+            (p / "maildirsize").unlink(missing_ok=True)
+
+
+if __name__ == "__main__":
+    main()
```
hpk42, line 41: when #344 is in, this should be rewritten using "doveadm --f flow quota get -A" to get all quotas from users 
and then `doveadm expunge  -A mailbox INBOX SAVEDBEFORE 1d \( SEEN \)` for each user that is above 80%

hpk42: here is a rough script that does it: 
```
import subprocess


def get_near_quota_users():
    res = subprocess.check_output(["doveadm", "-f", "flow", "quota", "get", "-A"]).decode("ascii")
    for line in res.split("\n"):
        parts = line.split()
        if not parts:
            continue
        if parts[4].lower() != "type=storage":
            continue
        addr = parts[0]
        c, percent = parts[-1].split("=")
        assert c == r"%"
        if int(percent) > 80:
            yield addr 

def expunge_seen_messages_from_near_quota_users():
    for addr in get_near_quota_users():
        print("trying to expunge seen messages", addr)
        subprocess.run(["doveadm", "expunge", "-u", addr, "mailbox", "INBOX", "(", "SEEN", ")"], check=True)
        subprocess.run(["doveadm", "purge", "-u", addr])

expunge_seen_messages_from_near_quota_users()
```

hpk42:
i think after we introduced "delete-server=at-once" as Android default we don't need to remove "old seen" messages anymore.

PR #343, opened 2024-07-05
Remove sieve to enable hardlink deduplication in LMTP

### LINKED ISSUE #339: dovecot: LMTP hardlinking is not working
LMTP is supposed to deliver message once for the first recipient, then hardlink or copy the message to other recipients. As in typical chatmail setup all mailboxes reside on the same filesystem and owned by the same user and group, delivered mails should be hardlinked. However, it is not the case, `stat` command shows "Links: 1" for all delivered mails.

Apparently it is the problem caused by Sieve which is used for message notifications: https://dovecot.dovecot.narkive.com/qfq7puPs/maildir-hardlinks-with-sieve

link2xt:
Remove sieve to enable hardlink deduplication in LMTP

    LMTP does not deduplicate messages
    if sieve plugin is used.

    We don't check for Auto-Submitted header anymore
    as iOS application has a notification service
    and should not display "You have a new message".

Fixes #339

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@
 - filtermail: do not allow ASCII armor without actual payload
   ([#325](https://github.com/deltachat/chatmail/pull/325))
 
+- Remove sieve to enable hardlink deduplication in LMTP
+  ([#343](https://github.com/deltachat/chatmail/pull/343))
+
+
 ## 1.3.0 - 2024-06-06
 
 - don't check necessary DNS records on cmdeploy init anymore
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -338,20 +338,6 @@ def _configure_dovecot(config: Config, debug: bool = False) -> bool:
     )
     need_restart |= lua_push_notification_script.changed
 
-    sieve_script = files.put(
-        src=importlib.resources.files(__package__).joinpath("dovecot/default.sieve"),
-        dest="/etc/dovecot/default.sieve",
-        user="root",
-        group="root",
-        mode="644",
-    )
-    need_restart |= sieve_script.changed
-    if sieve_script.changed:
-        server.shell(
-            name="compile sieve script",
-            commands=["/usr/bin/sievec /etc/dovecot/default.sieve"],
-        )
-
     files.template(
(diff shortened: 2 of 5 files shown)


chatmaild/src/chatmaild/metadata.py
```diff
@@ -117,7 +118,25 @@ def handle_dovecot_request(msg, transactions, notifier, metadata, iroh_relay=Non
...
+                    "-u",
+                    addr,
+                    "--",
+                    "hdr.auto-submitted",
+                    "uid",
+                    value,
+                ],
+                capture_output=True,
+                check=True,
+            )
```
hpk42, line 134: what happens if  "doveadm fetch" fails? Will this this not error out completely because `check==True` ?

chatmaild/src/chatmaild/metadata.py
```diff
@@ -117,7 +118,25 @@ def handle_dovecot_request(msg, transactions, notifier, metadata, iroh_relay=Non
         if keyname[0] == "priv" and keyname[2] == metadata.DEVICETOKEN_KEY:
             metadata.add_token_to_addr(addr, value)
         elif keyname[0] == "priv" and keyname[2] == "messagenew":
-            notifier.new_message_for_addr(addr, metadata)
+            proc = subprocess.run(
+                [
+                    "doveadm",
```
hpk42, line 123: creating a new subprocess for each incoming message is probably expensive but then the mime-parser of doveadm might be faster than python's.  However, it may make sense to find out the location of `doveadm` and specify an absolute path here.

chatmaild/src/chatmaild/metadata.py
```diff
@@ -117,7 +118,30 @@ def handle_dovecot_request(msg, transactions, notifier, metadata, iroh_relay=Non
...
+                    ],
+                    capture_output=True,
+                    check=True,
+                )
+                notify = proc.stdout not in [
+                    b"hdr.auto-submitted: auto-generated\n",
+                    b"hdr.auto-submitted: auto-replied\n",
+                ]
+            except FileNotFoundError:
+                # doveadm not available, don't notify
```
hpk42, line 141: better do something like `doveadm_path = subprocess.check_output(["which", "doveadm"], text=True).strip()` at module level and use that.

hpk42 (approved):
very nice to get rid of sieve. might also help reduce  IO load.  lint needs fixing before merge.

PR #345, opened 2024-07-06
remove fix-file-owner step 

hpk42:
which takes forever on servers with many mail directories

it's unclear why this is still needed and should be fixed differently in there is an actual problem.


missytake (approved):
As the CI passes, apparently it isn't needed during setup :)

PR #347, opened 2024-07-08
Add nonci_accounts metric

link2xt:
Calculating this with PromQL is not easy
due to interpolation.

Also add HELP and TYPE metadata for each metric.


hpk42 (approved):
lgtm if it works ;)

PR #349, opened 2024-07-08
doveauth: log when a new account is created

link2xt:


hpk42 (approved):

hagenest (approved):
Looks good👍

I wonder if we want to have different log levels someday

edit: seems like Holger was faster :D

PR #356, opened 2024-07-10
don't report on xfail, it's useless

hpk42:


hpk42:
> This `-rx` actually did overwrite `-rs` above?

not, it's cumulative ... see https://docs.pytest.org/en/7.1.x/how-to/output.html#producing-a-detailed-summary-report

PR #357, opened 2024-07-11
Multiplex HTTPS, IMAP and SMTP on port 443

link2xt:
Services are distinguished based on ALPN.
For example,
    openssl s_client -connect example.org:443 -alpn smtp
gives SMTP connection and
    openssl s_client -connect example.org:443 -alpn imap
gives IMAP connection.

### DIFF
--- a/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
+++ b/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
@@ -1,3 +1,5 @@
+load_module modules/ngx_stream_module.so;
+
 user www-data;
 worker_processes auto;
 pid /run/nginx.pid;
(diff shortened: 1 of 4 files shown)


cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
```diff
@@ -8,6 +10,21 @@ events {
 	# multi_accept on;
 }
 
+stream {
+        map $ssl_preread_alpn_protocols $proxy {
+            default 127.0.0.1:8443;
+            ~\bsmtp\b 127.0.0.1:465;
+            ~\bimap\b 127.0.0.1:993;
```
hpk42, line 17: does this mean that establishing SSL is done by imap/postfix still?

hpk42, line 17: can we use the /etc/service names of  "imaps" and "smtps" here? i'd prefer that.

hpk42 (approved):
very nice!

PR #358, opened 2024-07-11
fix to remove xfailing test

hpk42:


missytake (approved):
Nice, this was bugging me for so long...

PR #360, opened 2024-07-12
acmetool: reload nginx after requesting new cert

missytake:
This should fix reports of outdated TLS certs in NGINX, while postfix already has the more recent one.

Related: https://github.com/deltachat/pyinfra-acmetool/pull/2

Maybe we should update the version of pyinfra-acmetool we use in chatmail with the improvements we made in the package we use in the rest of our infrastructure.


hpk42 (approved):

link2xt:
I opened a PR that removes this cronjob:
https://github.com/chatmail/relay/pull/708

This change should not have been needed because there was a hook to restart nginx in place when this was committed.

PR #361, opened 2024-07-12
CI: don't let commits in other PRs interrupt CI runs

missytake:


missytake:
This didn't work, because when another PR started a workflow it didn't have to wait; it started, reset the VPS, making the workflow from this PR fail because the CI machine was killed.

missytake:
Now it works:

1. First we started https://github.com/deltachat/chatmail/actions/runs/9905900660/job/27366388792 (workflow 1) from #546.
2. then I pushed https://github.com/deltachat/chatmail/pull/361/commits/9006ea10f638461369c64d7178a279ec427ce015, which triggered https://github.com/deltachat/chatmail/actions/runs/9905934539/job/27366494427 (workflow 2). It didn't start to execute, but waited for workflow 1 instead.
3. then I pushed the last commit, which killed workflow 2, but then continued to wait for workflow 1.

hpk42 (approved):
nice. recommend to squash-merge.

PR #366, opened 2024-07-13
filtermail: do not inject addresses into fromat string

link2xt:


hpk42 (approved):

missytake:
What is the advantage of this change? Asking for educational purposes :)

PR #369, opened 2024-07-13
Accept encrypted messages that use hcp_minimal

dkg:
in draft-ietf-lamps-header-protection-22, hcp_minimal recommends "[...]" as the obscured Subject header.  In the pending draft -23 (hopefully released this week, going into a working group last call), the same HCP will be renamed to hcp_baseline, but it still recommends the use of "[...]" for the obscured Subject header.

### DIFF
--- a/chatmaild/src/chatmaild/tests/plugin.py
+++ b/chatmaild/src/chatmaild/tests/plugin.py
@@ -71,11 +71,11 @@ def maildata(request):
 
     assert datadir.exists(), datadir
 
-    def maildata(name, from_addr, to_addr):
+    def maildata(name, **kwargs):
         # Using `.read_bytes().decode()` instead of `.read_text()` to preserve newlines.
         data = datadir.joinpath(name).read_bytes().decode()
 
-        text = data.format(from_addr=from_addr, to_addr=to_addr)
+        text = data.format(**kwargs)
         return BytesParser(policy=policy.default).parsebytes(text.encode())
 
     return maildata
--- a/chatmaild/src/chatmaild/tests/test_filtermail.py
+++ b/chatmaild/src/chatmaild/tests/test_filtermail.py
@@ -54,10 +54,16 @@ def test_filtermail_no_encryption_detection(maildata):
 
 
 def test_filtermail_encryption_detection(maildata):
-    msg = maildata("encrypted.eml", from_addr="1@example.org", to_addr="2@example.org")
-    assert check_encrypted(msg)
+    for subject in ("...", "[...]"):
+        msg = maildata(
+            "encrypted.eml",
+            from_addr="1@example.org",
+            to_addr="2@example.org",
+            replaced_subject=subject,
+        )
+        assert check_encrypted(msg)
 
-    # if the subject is not "..." it is not considered ac-encrypted
+    # if the subject is not a known encrypted subject value, it is not considered ac-encrypted
     msg.replace_header("Subject", "Click this link")
     assert not check_encrypted(msg)
 
@@ -72,7 +78,7 @@ def test_filtermail_unencrypted_mdn(maildata, gencreds):
     """Unencrypted MDNs should not pass."""
     from_addr = gencreds()[0]
     to_addr = gencreds()[0] + ".other"
-    msg = maildata("mdn.eml", from_addr, to_addr)
+    msg = maildata("mdn.eml", from_addr=from_addr, to_addr=to_addr)
 
     assert not check_encrypted(msg)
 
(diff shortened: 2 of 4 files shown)


hpk42:
sounds good -- could you add a test similar to https://github.com/deltachat/chatmail/blob/main/chatmaild/src/chatmaild/tests/test_filtermail.py#L56 ? 
It's probably best if you modify https://github.com/deltachat/chatmail/blob/main/chatmaild/src/chatmaild/tests/mail-data/encrypted.eml#L3  to take the parameter from the test, and run the test with these two subjects. 

As to running (offline) tests there is no documented way, but here is the flow: 
```
scripts/initenv.sh
source venv/bin/activate
pytest chatmaild
```

this should run any of the abovementioned tests. Some will be skipped if you have no root-level "chatmail.ini" but that's fine. 
- As to running the tests for "cmdeploy test" you would need a chatmail.ini 
Let us know if you have troubles setting up a "pytest chatmaild" run to verify things locally.

hpk42 (approved):
thanks, lgtm.

hpk42:
tests fail though :) (it's the CI / isolated chatmaild tests) that are relevant.  Did this pass for you locally?

chatmaild/src/chatmaild/tests/test_filtermail.py
```diff
@@ -54,10 +54,14 @@ def test_filtermail_no_encryption_detection(maildata):
...
 def test_filtermail_encryption_detection(maildata):
-    msg = maildata("encrypted.eml", from_addr="1@example.org", to_addr="2@example.org")
-    assert check_encrypted(msg)
-
-    # if the subject is not "..." it is not considered ac-encrypted
+    for subject in {"...", "[...]"}:
+        msg = maildata(
+            "encrypted.eml", from_addr="1@example.org", to_addr="2@example.org",
+            replaced_subject=subject
+        )
```
hpk42, line 61: for this call to work you need to modify https://github.com/deltachat/chatmail/blob/main/chatmaild/src/chatmaild/tests/plugin.py#L74 as well.  arguably we could use `**kwargs` but that's not the case currently.

hpk42: i'd probably let `maildata()` accept an optional `subject` parameter, and pass it on to the format() call  (formatting is IIRC fine with extra parameters that are not used in a template-string).

hpk42: please do try to run "pytest chatmaild" locally -- this gives more direct feedback compared to CI.

dkg: the tests didn't pass for me locally (i understood your recommendation to test but didn't have an environment set up that i felt comfortable running the venv you described).  I'm working on sorting that out.

hpk42: truth be told we so far only cared for ./scripts/*.sh invocations -- so far nobody tried to land a PR to chatmaild except the three core maintainers :)
we are also using "black" and "ruff" for formatting.  This is done with "cmdeploy fmt" or "venv/bin/cmdeploy fmt" if you use the environment created by `./scripts/initenv.sh`

dkg:
the tests pass locally for me now, i don't understand why i'm not seeing the two failing CI builds above, but i think i can fix them anyway.

chatmaild/src/chatmaild/tests/plugin.py
```diff
@@ -71,11 +71,11 @@ def maildata(request):
 
     assert datadir.exists(), datadir
 
-    def maildata(name, from_addr, to_addr):
+    def maildata(name, **kwargs):
```
hpk42, line 74: i suggest to be more conservative here: 
```suggestion
    def maildata(name, from_addr, to_addr,  **kwargs):
```

dkg:
just pushed a final complaint from the linter

hpk42:
code is good but "ruff chatmaild" complains.

hpk42:
image

dkg:
I'm reading the CI reports, thanks :)  you don't need to cc them here :)

hpk42:
ah, can you try something?  
if you install "tox" (io assumed it's packaged for debian) and then run "cd chatmaild ; tox"?

hpk42:
> I'm reading the CI reports, thanks :) you don't need to cc them here :)

ah, i thought you couldn't see them.

dkg:
some of these CI failures appear to be flaky things that are not under my control:

20240713_12h27m49s_grim

hpk42:
yip, your PR is not allowed to access some secrets used for live-deploying and testing to staging2.testrun.org 
in invited you as outside collaborator. then it all works, and CI-runs don't need maintainer approval.

dkg:
so, i accepted your request, and then triggered the job to re-run, and it still seems to fail in the same way.   :shrug: I'll leave this in your capable hands ☺

hpk42:
k, maybe @nami or @link2xt can look into the precise github-rules there, i am less involved with github/CI magic myself. 
thanks for caring and congrats on a merged PR :)

PR #370, opened 2024-07-13
Add additional known obscured subjects from k-9/thunderbird-android.

dkg:
These additional subjects were extracted from the thunderbird-android source (which is inherited from k-9).

The extraction was done with:

```
git clone https://github.com/thunderbird/thunderbird-android/
cd thunderbird-android/legacy/ui/legacy/src/main/res
grep string\ name=\"encrypted_subject values-*/strings.xml | cut -f2 -d'>' | cut -f1 -d'<' | sort -u | sed -e 's/^/    "/' -e 's/$/",/'
```

(i did need to clean up one line's escaping to pass the linter's expectations)

See also https://github.com/thunderbird/thunderbird-android/issues/8011


chatmaild/src/chatmaild/filtermail.py
```diff
@@ -14,6 +14,64 @@
...
+    "დაშიფრული წერილი",
+    "הודעה מוצפנת",
+    "پیام رمزنگاری‌شده",
+    "رسالة مشفّرة",
+    "എൻക്രിപ്റ്റുചെയ്‌ത സന്ദേശം",
+    "加密邮件",
+    "已加密的訊息",
+    "暗号化されたメッセージ",
+}
+
```
hpk42, line 74: hum, we decided against going for all of these strings at one point. 

Shouldn't K-9 mail rather change to `[...]`?

hpk42: the thing is that a localized "encrypted subject" 

a) leaks the mail sender's locale 
b) mail receiver locale/language might not be able to make sense of sender's localized "encrypted subject"

hpk42:
if you want CI to pass, you need to rebase on main.  I meanwhile fixed some issues only revealed after a full CI run -- after merging your last PR.

chatmaild/src/chatmaild/filtermail.py
dkg: yes, k-9 should change, for the reasons you describe and more. i've asked them to do that at https://github.com/thunderbird/thunderbird-android/issues/8011

However, even if they adopt that change, existing deployments won't change until there is an upgrade path and it actually happens.

If chatmail wants to be interoperable with the deployed base, it should also change, as described here.

Once there are widely-deployed instances of k9 and thunderbird-android that do the sensible standardized thing, then maybe chatmail could consider reverting this.  But in the meantime, this seems like a reasonable measure for interoperability.

hpk42:
On Mon, Jul 15, 2024 at 05:02 -0700, dkg wrote:
> @dkg commented on this pull request.
> 
> > +    "Κρυπτογραφημένο μήνυμα",
> +    "Зашифроване повідомлення",
> +    "Зашифрованное сообщение",
> +    "Зашыфраваны ліст",
> +    "Криптирано съобщение",
> +    "Шифрована порука",
> +    "დაშიფრული წერილი",
> +    "הודעה מוצפנת",
> +    "پیام رمزنگاری‌شده",
> +    "رسالة مشفّرة",
> +    "എൻക്രിപ്റ്റുചെയ്‌ത സന്ദേശം",
> +    "加密邮件",
> +    "已加密的訊息",
> +    "暗号化されたメッセージ",
> +}
> +
> 
> yes, k-9 should change, for the reasons you describe and more. i've asked them to do that at https://github.com/thunderbird/thunderbird-android/issues/8011

thanks! 

> However, even if they adopt that change, existing deployments won't change until there is an upgrade path and it actually happens.
> 
> If chatmail wants to be interoperable with the deployed base, it should also change, as described here.

Delta Chat users on chatmail are interoperable with K9 users on non-chatmail servers
as incoming K9 PGP-encrypted messages should be fine: Delta Chat will ignore the 
"cleartext" subject and not show it to users at all. 

It is true that K9 users can not use chatmail servers to send out encrypted messages
which is however not a supported use case currently and does not affect interoperability
in the above sense. 

As cketti has thankfully done https://github.com/thunderbird/thunderbird-android/pull/8014
i think it's better to close this PR here. 
It should then be at least possible to use K9 with chatmail addresses. 
Delta Chat itself will btw evolve to use `[...]` subjects instead of `...`
so it will all be pretty much aligned.

hpk42 (approved):
while i don't regard supporting k9 (or any other MUA's) usage with chatmail servers a recommended workflow it's easy enough to accept this PR.  link2xt is also +0 on it.

PR #371, opened 2024-07-13
fix maildata handling after prematurely merging #369

hpk42:
this should fix what wasn't quite right yet about #369 but wasn't detected because the CI runs not well for outside-collaborators.

### DIFF
--- a/chatmaild/src/chatmaild/tests/plugin.py
+++ b/chatmaild/src/chatmaild/tests/plugin.py
@@ -71,11 +71,11 @@ def maildata(request):
 
     assert datadir.exists(), datadir
 
-    def maildata(name, **kwargs):
+    def maildata(name, from_addr, to_addr, subject="..."):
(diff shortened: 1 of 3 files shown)


chatmaild/src/chatmaild/tests/plugin.py
```diff
@@ -71,11 +71,11 @@ def maildata(request):
 
     assert datadir.exists(), datadir
 
-    def maildata(name, **kwargs):
+    def maildata(name, from_addr, to_addr, subject="..."):
```
missytake, line 74: shouldn't it be `[...]`?

missytake (approved):

chatmaild/src/chatmaild/tests/plugin.py
hpk42: the default before  was  "..."  so i just kept it at that.

PR #355, opened 2024-07-10
streamline DNS checks further

hpk42:
this is an attempt to help with and close  #317  
and also to get rid of "ip-address" guessing, thereby closing #336 

It moves chatmail.zone generation to jinja2 to be able to use conditional-statements
and should allow ipv4 and ipv6-only setups respectively (but not tested in real life).

PR #372, opened 2024-07-15
refine DNS-checking and add automated tests 

hpk42:
supersedes #355 

There are now phases of DNS checks: 

- initial DNS entries that must be present for "cmdeploy run" 

- required DNS entries for operating within the chatmail server network 

- recommended DNS settings for interoperability and security hardening 

"cmdeploy dns" will return 0 as exitcode (all ok) if all required DNS settings are correct but will print warnings for any missing recommended one 

We might want to shift the SPF record from recommended into required.


cmdeploy/src/cmdeploy/remote_funcs.py
```diff
@@ -30,76 +31,105 @@ def get_systemd_running():
...
-            continue
+        query_value = query_dns(zf_typ, zf_domain)
+        if zf_value != query_value:
+            assert zf_typ in ("A", "AAAA", "CNAME", "CAA", "SRV", "MX", "TXT"), zf_line
+            if required:
+                required_diff.append(zf_line)
+            else:
+                recommended_diff.append(zf_line)
+
+    return required_diff, recommended_diff
```
missytake, line 101: Wow, this is beautiful. I still think with horror of my previous spaghetti code

cmdeploy/src/cmdeploy/tests/test_dns.py
```diff
@@ -0,0 +1,117 @@
...
+        return func(**kwargs)
+
+
+class TestZonefileChecks:
+    def test_check_zonefile_all_ok(self, cm_data, mockdns_base):
+        zonefile = cm_data.get("zftest.zone")
+        parse_zonefile_into_dict(zonefile, mockdns_base)
+        required_diff, recommended_diff = remote_funcs.check_zonefile(zonefile)
+        assert not required_diff and not recommended_diff
+
```
missytake, line 90: With this PR, setting recommended DNS records on with a zonefile is sliding out of the part which is tested by CI, as `cmdeploy dns` doesn't fail for them anymore. This is probably not a huge problem, but we should notice it consciously.

In the past, the CI set all DNS records on ns.testrun.org, and then ran `cmdeploy dns` to verify them, and return had to be zero (with three tries). Now I wonder if we would notice if one of the recommended DNS records isn't set as expected anymore, as CI could still be green. Note that there can be complications while setting DNS records, e.g. formatting of DKIM records. OTOH, most of them happen in web interfaces, which are not tested anyway.

missytake (approved):
Looks great!

cmdeploy/src/cmdeploy/tests/test_dns.py
hpk42: good point.  Let's maybe do this is a separate PR -- an option "--all" that gives return code 1 if any dns record does not match.

PR #373, opened 2024-07-15
refactor: use f-string in logging where it is easy

link2xt:
% is only interpreted if there are two or more arguments: <https://docs.python.org/3/library/logging.html#logging.Logger.debug> So it is safe to pass a single argument with already formatted string.


hpk42 (approved):
lgtm

PR #375, opened 2024-07-17
nginx: use numbers for upstream ports

link2xt:
Otherwise nginx fails when user actually tries to connect, logs have errors such as
`invalid port in upstream "127.0.0.1:imaps"`
and
`invalid port in upstream "127.0.0.1:submissions"`.


missytake (approved):
:tada:

hpk42 (approved):
strange but ok

PR #381, opened 2024-07-28
introduce max_message_size config option

### LINKED ISSUE #268: Two more variables into the chatmail.ini
As chatmail is recommended for small groups, the compliance into this group can it make affordable to change the maximal size of the message and the time to delete an inactive account, too. 

Please add these two variables into chatmail.ini if it is possible:

max_size =              # maximal size for a message in megabyte
time_to_delete =    # an inactive acoount in days

Thank you!

hpk42:
fixes #268 and also fixes a format-string typo/error on the side.

### DIFF
--- a/chatmaild/src/chatmaild/config.py
+++ b/chatmaild/src/chatmaild/config.py
@@ -20,6 +20,7 @@ class Config:
         self.mail_domain = params["mail_domain"]
         self.max_user_send_per_minute = int(params["max_user_send_per_minute"])
         self.max_mailbox_size = params["max_mailbox_size"]
+        self.max_message_size = int(params.get("max_message_size", "31457280"))
         self.delete_mails_after = params["delete_mails_after"]
         self.delete_inactive_users_after = int(params["delete_inactive_users_after"])
         self.username_min_length = int(params["username_min_length"])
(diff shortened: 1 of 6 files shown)


chatmaild/src/chatmaild/config.py
```diff
@@ -73,7 +74,7 @@ def write_initial_config(inipath, mail_domain, overrides):
         new_line = line.strip()
         if new_line and new_line[0] not in "#[":
             name, value = map(str.strip, new_line.split("=", maxsplit=1))
-            value = overrides.pop(name, value)
+            value = extra.pop(name, value)
```
hpk42, line 77: this is an unrelated fix which was needed for the added test.

cmdeploy/src/cmdeploy/dns.py
```diff
@@ -15,7 +15,7 @@ def get_initial_remote_data(sshexec, mail_domain):
 def check_initial_remote_data(remote_data, print=print):
     mail_domain = remote_data["mail_domain"]
     if not remote_data["A"] and not remote_data["AAAA"]:
-        print("Missing A and/or AAAA DNS records for {mail_domain}!")
+        print(f"Missing A and/or AAAA DNS records for {mail_domain}!")
```
hpk42, line 18: this is also unrelated but doing an extra PR is too much.

missytake (approved):

PR #382, opened 2024-07-28
run filtermail as dedicated user

### LINKED ISSUE #322: filtermail should run as an unprivileged user
It is currently not doing much and should be secure, but also does not need access to any data. It should probably run as a separate user without the ability to write anywhere to the disk using systemd sandboxing. Since it is an SMTP server listening on TCP port, it does not even need to create or access unix sockets.

hpk42:
fixes #322


missytake (approved):
What is this fix needed for?

PR #388, opened 2024-07-30
Setup mtail

link2xt:
This runs `mtail` at localhost, so you can query `http://127.0.0.1:3903/metrics` and get metrics collected by parsing the log files to get some counters of the number of sent/received mails.

This can then be plugged into Prometheus. Setting up Prometheus is outside of the scope of cmdeploy.

What is missing is a setting in the .ini file to set the host and port. Listening on 127.0.0.1 is not very useful, normally you would have a wireguard network for metric collection and make mtail listen on the IP address from this network.

Another problem currently is that deployment fails while trying to copy `mtail` binary if it is already running. We should stop `mtail` service before replacing the binary somehow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,14 @@
 
 ## untagged
 
-There is a new setting in `chatmail.ini`: `mtail_address`.
-This defines the address on which [`mtail`](https://google.github.io/mtail/)
-exposes its metrics collected from the logs.
-If you want to collect the metrics with Prometheus,
-setup a private network (e.g. WireGuard interface)
-and assign an IP address from this network to the host.
-If you do not plan to collect metrics,
-keep this setting unset.
-
-- add mtail
+- add mtail support (new optional `mail_address` ini value)
+  This defines the address on which [`mtail`](https://google.github.io/mtail/)
+  exposes its metrics collected from the logs.
+  If you want to collect the metrics with Prometheus,
+  setup a private network (e.g. WireGuard interface)
+  and assign an IP address from this network to the host.
+  If you do not plan to collect metrics,
+  keep this setting unset.
   ([#388](https://github.com/deltachat/chatmail/pull/388))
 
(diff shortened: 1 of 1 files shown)


hpk42:
> Another problem currently is that deployment fails while trying to copy `mtail` binary if it is already running. We should stop `mtail` service before replacing the binary somehow.

but this makes it hard to merge right away, no?  

also, there is an "mtail" debian package which is probably better to use than downloading from google server and unpacking etc.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -441,6 +441,44 @@ def check_config(config):
...
+
+    # Using our own systemd unit instead of `/usr/lib/systemd/system/mtail.service`.
+    # This allows to read from journalctl instead of log files.
+    files.template(
+        src=importlib.resources.files(__package__).joinpath("mtail/mtail.service.j2"),
+        dest="/etc/systemd/system/mtail.service",
+        user="root",
+        group="root",
+        mode="644",
+        address="127.0.0.1",  # TODO read from config
```
hpk42, line 458: i think this needs to happen still in this PR -- if the chatmail.ini does not configure mtail address + port, then we should not setup or at least not start mtail i guess.

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 ## untagged
 
+There is a new required setting in `chatmail.ini`: `mtail_address`.
+This defines the address on which [`mtail`](https://google.github.io/mtail/)
+exposes its metrics collected from the logs.
+If you want to collect the metrics with Prometheus,
+setup a private network (e.g. WireGuard interface)
+and assign an IP address from this network to the host.
+If unsure, set this setting to `127.0.0.1`.
```
hpk42, line 11: why is it a required setting?  It should be fine to leave it empty, and then no mtail is setup or run?

hpk42 (approved):

PR #403, opened 2024-08-15
Move MX record to its own subdomain

link2xt:


missytake (approved):

hpk42:
changelog entry is missing.

hpk42 (changes requested):
Please provide a reason for this change.  i can guess but it's not clear to me, sorry. 
Also, would the letsencrypt certs not need to be adapted for "mx.maildomain"?

missytake:
> Please provide a reason for this change. i can guess but it's not clear to me, sorry. Also, would the letsencrypt certs not need to be adapted for "mx.maildomain"?

One reason is that we can then add proxy VPSs for IMAP+SMTP usage only, but not as MX.

PR #406, opened 2024-09-02
add a note on deletion of accounts

### LINKED ISSUE #405: Mention delete_inactive_users_after in info.html (and privacy.html?)
This would help with Google Play Store requirements about data retention information.

hpk42:
fixes #405 

also lowers the inactive account deletion to 90 days by default.

### DIFF
--- a/chatmaild/src/chatmaild/ini/chatmail.ini.f
+++ b/chatmaild/src/chatmaild/ini/chatmail.ini.f
@@ -23,8 +23,8 @@ max_message_size = 31457280
 # days after which mails are unconditionally deleted
 delete_mails_after = 20
 
-# days after which users without a login are deleted (database and mails)
-delete_inactive_users_after = 100
+# days after which users without a successful login are deleted (database and mails)
+delete_inactive_users_after = 90
 
 # minimum length a username must have
 username_min_length = 9
--- a/www/src/info.md
+++ b/www/src/info.md
@@ -43,6 +43,20 @@ The first login sets your password.
 - You can store up to [{{ config.max_mailbox_size }} messages on the server](https://delta.chat/en/help#what-happens-if-i-turn-on-delete-old-messages-from-server).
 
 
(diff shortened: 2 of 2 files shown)


chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -23,8 +23,8 @@
 # days after which mails are unconditionally deleted
 delete_mails_after = 20
 
-# days after which users without a login are deleted (database and mails)
-delete_inactive_users_after = 100
+# days after which users without a successful login are deleted (database and mails)
+delete_inactive_users_after = 90
```
missytake, line 27: We could change this only in https://github.com/deltachat/sysadmin-pyinfra/blob/main/servers/nine.testrun.org/chatmail.ini#L27, it doesn't need to be the new default IMHO - but no blocker.

www/src/info.md
```diff
@@ -43,6 +43,20 @@ The first login sets your password.
...
+If you remove a {{ config.mail_domain }} profile from within the Delta Chat app, 
+then the according account on the server, along with all associated data,
+is automatically deleted {{ config.delete_inactive_users_after }} days afterwards. 
+
+If you use multiple devices 
+then you need to remove the according chat profile from each device
+in order for all account data to be removed on the server side. 
+
+If you have any further questions or requests regarding account deletion
+please send a message from your account to {{ config.privacy_mail }}.
```
missytake, line 57: If `privacy_mail` is unset it will look weird, OTOH chatmail admins really should have a privacy policy and can just remove the sentence otherwise.

missytake: But we could make an if clause around the paragraph.

missytake (approved):
Thanks!

www/src/info.md
hpk42: Maybe it's better to show a broken message than just removing the sentence.  This way the user can become aware that they don't have any handle on who runs the server ;)

PR #408, opened 2024-09-06
add support for specifying whole domains for passthrough recipients

hpk42:
this is to be used for whitelisting "@cheogram" for XMPP-interactions for webxdc apps.


chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -39,7 +39,7 @@
 passthrough_senders =
 
 # list of e-mail recipients for which to accept outbound un-encrypted mails
-# (space-separated)
+# (space-separated, item may start with "@" to whitelist whole recipient domains)
```
missytake, line 42:
```suggestion
# (space-separated, item may start with "@" to allow whole recipient domains)
```
https://www.ietf.org/archive/id/draft-knodel-terminology-09.html

missytake (approved):
Looks reasonable :)

hpk42:
the problem is that Delta Chat on a chatmail server does not know that the chatmail server allows unencrypted mail, so DC itself will block it.  How to address this?

missytake:
> the problem is that Delta Chat on a chatmail server does not know that the chatmail server allows unencrypted mail, so DC itself will block it. How to address this?

will it? afaict the client sends the message, receives the SMTP error, and then displays it nicely: 
screenshot of message details of a failed message from a nine account to echo@example.org

If it doesn't receive the SMTP error everything should be fine, right?

hpk42:
On Mon, Oct 14, 2024 at 05:15 -0700, missytake wrote:
> > the problem is that Delta Chat on a chatmail server does not know that the chatmail server allows unencrypted mail, so DC itself will block it. How to address this?
> 
> will it? afaict the client sends the message, receives the SMTP error, and then displays it nicely: 
> ![screenshot of message details of a failed message from a nine account to ***@***.***(https://github.com/user-attachments/assets/7ccfe2c4-df12-4d96-906e-8e73ce7f4440)
> 
> If it doesn't receive the SMTP error everything should be fine, right?

There are cases where Android does not even attempt to send a message
but directly displays a "blocking" warning, see
https://github.com/deltachat/deltachat-android/pull/3177

(in short: put in Android ***@***.***" into saved-messages
then tap on the highlighted message)

PR #411, opened 2024-09-13
Set CAA record flags to 0

### LINKED ISSUE #410: Set CAA record flags to 0
Cloudflare only allows to set flags to 0:
letsencrypt

We currently set "Issuer Critical Flag" (bit 0) so flags is 128. Given that we only use "issue" tag which is defined in https://datatracker.ietf.org/doc/html/rfc6844 together with the "issue critical" flag itself, there is not much point setting it. We need to somehow find a CA that does not understand "issue" but understand critical flag to abort issuing the certificate.

Looking at github.com, cloudflare.com, slashdot.org, google.com - all of them set flags to 0.
fastmail.com sets flags to 0 for some records and to 128 for some other.

https://letsencrypt.org/docs/caa/ says "**If you like**, you can set the flags to the integer 128, indicating that the “critical bit” has been set, and that CAs should immediately halt and not issue a certificate if they don’t recognize the contents of the tag field."

link2xt:
Close #410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 - drop hispanilandia passthrough address
   ([#401](https://github.com/deltachat/chatmail/pull/401))
 
+- set CAA record flags to 0
+
 
 ## 1.4.1 2024-07-31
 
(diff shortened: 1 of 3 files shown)


missytake (approved):
If we change it like this, the next time someone who has already set up chatmail runs `cmdeploy dns`, it will recommend to set their perfectly fine 128 record to 0... that's a bit unfortunate, but probably fine. I don't think it will happen very often.

PR #412, opened 2024-09-13
DNS: fix checking for required DNS records

missytake:


missytake:
I have no idea why those two CI checks fail. Locally they pass for me. As it's just lint, I propose to ignore them and merge this PR nonetheless.

PR #413, opened 2024-09-22
Add IMAP capabilities instead of overwriting them

link2xt:
I wanted to add `COMPRESS=DEFLATE`,
but it should be added only for sessions
that are logged in because `COMPRESS`
command does not work before logging in.

Dovecot already does it correctly
if we don't overwrite the capability string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@
 
 - set CAA record flags to 0
 
+- add IMAP capabilities instead of overwriting them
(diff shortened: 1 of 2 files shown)


hpk42:
the overhead seems to be roughly 350 bytes -- it's not super-huge i agree, but also i don't see the point of adding ~300 bytes unneccessarily for each imap connection.

hpk42 (approved):

PR #417, opened 2024-10-10
Document how to migrate the server

link2xt:


hpk42:
This looks good but what about installing the "firewall" scripts with cmdeploy-run?  Something like /use/local/bin/chatmail-* commands?  Would be nicer and we can also adapt.it if needed.

README.md
```diff
@@ -187,3 +187,225 @@ to MAIL FROM with
...
+        ct state established accept
+    }
+    chain forward {
+        type filter hook forward priority filter;
+    }
+    chain output {
+        type filter hook output priority filter;
+    }
+}
+```
```
hpk42, line 225: how about we add a `cmdeploy run --disable-mx`  switch that makes sure the server is setup but not reachable/accepting SMTP/IMAP?  (not sure if simply not running postfix/dovecot might be enough?)

README.md
```diff
@@ -187,3 +187,225 @@ to MAIL FROM with
 and rejects incorrectly authenticated emails with [`reject_sender_login_mismatch`](reject_sender_login_mismatch) policy.
 `From:` header must correspond to envelope MAIL FROM,
 this is ensured by `filtermail` proxy.
+
+## Migrating chatmail server to a new host
+
+If you want to migrate your chatmail server to a new host,
+follow these steps:
+
+1. Block all ports except 80 and 22 with firewall on a new server.
```
hpk42, line 196:
```suggestion
1. On new server: block all ports except 80 and 22 with firewall on a new server.
```

README.md
```diff
@@ -187,3 +187,225 @@ to MAIL FROM with
...
+3. Setup the new server with `cmdeploy`.
+
+   This step is similar to initial setup.
+   However, because ports Delta Chat uses are blocked,
+   new server will not become usable immediately.
+   If other servers try to deliver messages to your new server they will fail,
+   but normally email servers will retry delivering messages
+   for at least a week, so messages will not be lost.
+
+4. Firewall all ports except `ssh` (22) on the old server.
```
hpk42, line 254:
```suggestion
4. On old server: firewall all ports except `ssh` (22) 
```

README.md
```diff
@@ -187,3 +187,225 @@ to MAIL FROM with
...
+   it will lose track of the IMAP message UID
+   and miss messages that arrived during migration.
+
+   Same for SMTP port 25, you want it blocked during migration so no new mails arrive
+   while the server is moving.
+
+5. Use `rsync -avz` over SSH to copy /home/vmail/mail from the old server to the new one
+   preserving file permissions and timestamps.
+
+6. Unblock ports used by Delta Chat and SMTP message exchange.
```
hpk42, line 270:
```suggestion
6. On new server: unblock ports used by Delta Chat and SMTP message exchange.
```

README.md
```diff
@@ -187,3 +187,225 @@ to MAIL FROM with
...
+    chain forward {
+        type filter hook forward priority filter;
+    }
+    chain output {
+        type filter hook output priority filter;
+    }
+}
+```
+Execute `nft -f /etc/nftables.conf` as root to apply the changes.
+
```
hpk42, line 300: i'd add a 

`step 7. Switch off old server` and maybe advise for "cmdeploy test" to run successfully first, and wait to hear from users if they are any issues.

README.md
```diff
@@ -187,3 +187,225 @@ to MAIL FROM with
...
+## Setting up a reverse proxy
+
+chatmail server does not depend on the client IP address
```
hpk42, line 303:
```suggestion
A chatmail server does not depend on the client IP address
```

README.md
```diff
@@ -187,3 +187,225 @@ to MAIL FROM with
...
+for its operation, so it can be run behind a reverse proxy.
+This will not even affect incoming mail authentication
+as DKIM only checks cryptographic signature
```
hpk42, line 306:
```suggestion
as DKIM only checks the cryptographic signature
```

hpk42 (approved):
this all looks good -- would be fine to merge preferably after addressing/accepting some of my suggested changes.

README.md
missytake: I agree... not everyone knows how to adjust nftables. Admins will find an existing configuration at `/etc/nftables.conf`, modified by the ISP, their past self, or a team member. Adjusting nftables will much more easily lead to misconfiguration than simply stopping postfix + dovecot directly after initial setup in a `cmdeploy run --disable-mx` command.

If we leave out step 1 in favor of `cmdeploy run --disable-mx`, we could also swap (current) step 2 & 3 to keep the "registration doesn't work" timespan shorter.

README.md
```diff
@@ -187,3 +187,113 @@ to MAIL FROM with
...
+3. Setup the new server with `cmdeploy`.
+
+   This step is similar to initial setup.
+   However, because ports Delta Chat uses are blocked,
+   new server will not become usable immediately.
+   If other servers try to deliver messages to your new server they will fail,
+   but normally email servers will retry delivering messages
+   for at least a week, so messages will not be lost.
+
+4. Firewall all ports except `ssh` (22) on the old server.
```
missytake, line 254: Same here, I think just stopping dovecot + postfix is easier for admins than adjusting nftables and achieves the same result. And the know-how is more wide-spread among self-taught admins than adjusting firewalls.

README.md
```diff
@@ -187,3 +187,113 @@ to MAIL FROM with
...
+   and miss messages that arrived during migration.
+
+   Same for SMTP port 25, you want it blocked during migration so no new mails arrive
+   while the server is moving.
+
+5. Use `rsync -avz` over SSH to copy /home/vmail/mail from the old server to the new one
+   preserving file permissions and timestamps.
+
+6. Unblock ports used by Delta Chat and SMTP message exchange.
+   For that you can modify `/etc/nftables.conf` as follows:
```
missytake, line 271: Again here, simply starting the services is much easier. Recommending firewall settings assumes a lot about the environment the chatmail server is running in; it might run on the same OS as other services which require custom configuration, it might be co-managed by other people, ...

missytake (commented):
I left some comments to argue why I'd recommend stopping/starting services over firewall adjustments... but not a merge blocker.

README.md
missytake: > If we leave out step 1 in favor of `cmdeploy run --disable-mx`, we could also swap (current) step 2 & 3 to keep the "registration doesn't work" timespan shorter.

no we can't, as step 2 is required for step 3's acmetool setup to work.

missytake: Hm, if we simply copy `/var/lib/acme` in the beginning it should work.

PR #418, opened 2024-10-14
Document setting up DNAT

link2xt:
Moved out of #417 with fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,3 +187,115 @@ to MAIL FROM with
 and rejects incorrectly authenticated emails with [`reject_sender_login_mismatch`](reject_sender_login_mismatch) policy.
 `From:` header must correspond to envelope MAIL FROM,
 this is ensured by `filtermail` proxy.
+
+## Setting up a reverse proxy
+
+A chatmail server does not depend on the client IP address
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -187,3 +187,115 @@ to MAIL FROM with
...
+of the message and does not use the IP address as the input.
+
+For example, you may want to self-host your chatmail server
+and only use hosted VPS to provide a public IP address
+for client connections and incoming mail.
+You can also setup multiple reverse proxies
+for your chatmail server in different networks
+to ensure your server is reachable even when
+one of the IPs becomes inaccessible due to
+hosting or routing problems.
```
hpk42, line 206: would mention the need to setup tailscale or zerotier like tunneling, so that the public ip-address server can internally reach the NATted server.

README.md
```diff
@@ -187,3 +187,115 @@ to MAIL FROM with
...
+
+Note that your server still needs
+to be able to make outgoing connections on port 25
+to send messages outside.
```
hpk42, line 210: what is missing to make the public ip-address proxy act as a gateway for the internal actual server, so that it can do connects to port 25 of other public internet servers?

hpk42 (approved):

PR #420, opened 2024-10-15
docs: more details for the repo overview

### LINKED ISSUE #419: bring repository structure docs up-to-date
- explain that chatmaild can be installed separately
- explain how `cmdeploy run` applies config values, while chatmaild reads them from the deployed chatmail.ini at runtime
- document all chatmaild services
- document how push notifications are passed from dovecot via a lua `push_notification.lua` to the chatmail-metadata service, which passes it onto notifications.delta.chat

missytake:
fix #419 

I decided against mentioning that you can modify the chatmail.ini on the server and restart a chatmaild service to make it use the new value on runtime. Admins don't need to know that there is even a copy of the config file on the server; editing it is a bad idea anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ This repository has four directories:
 - [cmdeploy](https://github.com/deltachat/chatmail/tree/main/cmdeploy)
   is a collection of configuration files
   and a [pyinfra](https://pyinfra.com)-based deployment script.
+
 - [chatmaild](https://github.com/deltachat/chatmail/tree/main/chatmaild)
   is a python package containing several small services
   which handle authentication,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,31 +80,84 @@ scripts/cmdeploy bench
...
-
-as well as custom services that are integrated with these two:
-
-- `chatmaild/src/chatmaild/doveauth.py` implements
+This repository has four directories:
+
+- [cmdeploy](https://github.com/deltachat/chatmail/tree/main/cmdeploy)
+  is a collection of configuration files
+  and a [pyinfra](https://pyinfra.com)-based deployment script.
+- [chatmaild](https://github.com/deltachat/chatmail/tree/main/chatmaild)
```
hpk42, line 88: could you use a blank line to separate the list items? It doesn't make a difference in rendering but when reading the text source.

hpk42 (approved):
big improvement.  Have some suggestions for streamlining but will better do this in a followup PR when i get to it.

PR #423, opened 2024-10-15
fix: set acme_account_url even if some DNS records are not set

link2xt:
perform_initial_checks may exit early
and not add `acme_account_url` if required DNS
records are not found.
In this case other `cmdeploy run` fails
with KeyError.

To avoid this, `acme_account_url` should always be set.

Unlike DNS checks, running acmetool
may not fail due to network errors,
so it is more reliable and should be checked first.


missytake (approved):
This basically means `acme_account_url` is set to "" on first deployment, and `dkim_entry` to `None`,  which is fine because it isn't actually useful during first `cmdeploy run`, only for `cmdeploy dns` which calls the same function?

Issue #414, opened 2024-09-24
cmdeploy issues in a NAT'd environment


feld:
I stumbled upon this while trying to deploy to a fresh Debian Bookworm VM behind NAT last night and had to work around it. This scenario breaks some assumptions that are built into chatmail:

If you try to deploy in a NAT'd environment where the hostname of the server is a private/internal hostname and not the same as the A/AAAA record and you'll find cmdeploy uses the chatmail.ini value for `mail_domain`, so you have to override that in a `~/.ssh/config` so it will connect to the right place.

After you've overcome that hurdle, the next issue is that acmetool appears to have some built-in sanity checks to make sure it's reachable before attempting to issue the certificate, so it tries to reach http://`mail_domain`:80. This will fail because of hairpin routing (request hits your firewall/router and has to come right back out the same interface) which is generally denied for security reasons. You can fix this by adding an entry to your `/etc/hosts` file to declare your `mail_domain` is reachable at 127.0.0.1

At this point I can confirm that 80, 443, and all the IMAP/SMTP ports are open to the public internet.  However, acmetool keeps failing to issue a certificate and I don't know why. I didn't capture those logs (sorry, it was very late and I just wanted this to work), but it was the typical failed-HTTP-validation scenario. I was looking in `/run/acme/acme-challenge` and nothing ever showed up in there, so I'm not surprised the validation was failing. Very strange all around.

I worked around this by disabling acmetool during deployment and using my own automation with Chef to deploy my certificate to the same location acmetool was going to use.

The chatmail cmdeploy could use some work to allow providing your own certificates and to not make some of these assumptions, or at least permit overriding what you're connecting to for deployment.

hpk42:
There is #439 which introduced an ssh-option. this simplifies the first problem, right?

hpk42:
could you maybe close this issue here and make a new one if needed, and also provide more information so we can work on a fix, addressing it?

PR #428, opened 2024-10-16
cmdeploy: flag to disable postfix + dovecot for migration

missytake:
Migration is still not super nice because it required me to do a lot of back-and-forth in my local /etc/hosts file, and keep track of "which certificates are on which server now, where does DNS point already, and for who..."

But this made it work in principle.

Note that postfix will still be reachable for a while during the initial `cmdeploy run --disable-mail` command on the new server, so if DNS is already pointing to it, some messages will be sucked into the void. I will write a migration guide which avoids this.

### DIFF
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -479,10 +479,11 @@ def deploy_mtail(config):
     )
 
 
-def deploy_chatmail(config_path: Path) -> None:
+def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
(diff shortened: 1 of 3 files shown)


hpk42 (approved):
lgtm

hpk42:
would it help to introduce an "ssh-host" option either as command line or via chatmail.ini file? 
See also https://github.com/deltachat/chatmail/issues/414

missytake:
> would it help to introduce an "ssh-host" option either as command line or via chatmail.ini file? See also #414

yes.... but out of scope for this PR.

Issue #195, opened 2024-01-22
DKIM key rotation


link2xt:
We cannot expect admins to rotate the key or change DNS records after setting up a server because, but there should be some way to do it. E.g. a command `cmdeploy dns` can always suggest to add next key and then if corresponding DKIM public key is already deployed make the server switch to it. Then just running `cmdeploy dns` from time to time and deploying DKIM records it suggests will rotate the keys at least eventually.

For reference: https://www.m3aawg.org/DKIMKeyRotation

kescherCode:
Delegating the record would, however, cause issues with DNSSEC-protected zones, no?

As someone who runs their own nameservers, I'd love to be able to set up this rotation within this setup using something like rfc2136 updates.

PR #429, opened 2024-10-16
Guide on how to migrate chatmail to a new host

missytake:
This guide doesn't require knowing about firewalls, but utilizes the `cmdeploy run --disable-mail` command from #428 and the `cmdeploy run --ssh-host` command from #439. Should be merged after those.

supercedes #417

Tested by migrating c2 back and forth, especially the second time worked like a charm :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - increase `request_queue_size` for UNIX sockets to 1000.
   ([#437](https://github.com/deltachat/chatmail/pull/437))
 
+- add argument to `cmdeploy run` for specifying
+  a different SSH host than `mail_domain`
+  ([#439](https://github.com/deltachat/chatmail/pull/439))
+
 - query autoritative nameserver to bypass DNS cache
   ([#424](https://github.com/deltachat/chatmail/pull/424))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- add guide to migrate chatmail to a new server
+  ([#429](https://github.com/deltachat/chatmail/pull/429))
+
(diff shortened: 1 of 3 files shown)


missytake:
I wonder whether we should also preserve the DKIM key, as it might lead to rejected messages if multiple keys are around for the same dkim selector.

hpk42 (commented):
To be honest, it all looks workable but also rather error-prone/tedious, especially the constant overwriting of host identity keys. 
I think we need a "cmdeploy --ssh-host=sshname" option because then one can just add "old" and "new" to `/etc/hosts` and do "cmdeploy --ssh-host=old run" etc, and skip all the hostkey-overwriting. It should also help with the NAT-issue and should not be hard to do.

missytake:
> > I wonder whether we should also preserve the DKIM key, as it might lead to rejected messages if multiple keys are around for the same dkim selector.
> 
> I think we should generate a new DKIM selector for each deployment. Using current date like 20241017 for the selector name. Ideally DKIM selector to key mapping should be possible to cache indefinitely, so reusing DKIM selectors is not a good idea.

Do you consider #195 a blocker for merging this?

missytake:
> To be honest, it all looks workable but also rather error-prone/tedious, especially the constant overwriting of host identity keys. I think we need a "cmdeploy --ssh-host=sshname" option because then one can just add "old" and "new" to `/etc/hosts` and do "cmdeploy --ssh-host=old run" etc, and skip all the hostkey-overwriting. It should also help with the NAT-issue and should not be hard to do.

With https://github.com/deltachat/chatmail/pull/429/commits/c9b574b87d7ab3c0b05a913eb27d8b7317ceb1c2 we're down from 16 to 9 steps :)))

hpk42 (approved):
overall looks great, some minor comments only. 
Haven't run it myself though. 
Changelog entry is missing.

PR #434, opened 2024-10-22
Deploy iroh relay

link2xt:


cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -479,6 +479,38 @@ def deploy_mtail(config):
     )
 
 
+def deploy_iroh_relay() -> None:
+    server.shell(
+        name="Download iroh-relay",
+        commands=[
+            "(echo '8af7f6d29d17476ce5c3053c3161db5793cb2ac49057d0bcaf689436cdccbeab  /usr/local/bin/iroh-relay' | sha256sum -c) || curl -L https://github.com/n0-computer/iroh/releases/download/v0.27.0/iroh-relay-v0.27.0-x86_64-unknown-linux-musl.tar.gz | gunzip | tar -x -f - ./iroh-relay -O >/usr/local/bin/iroh-relay",
```
hpk42, line 486: what is the checksum check really doing?  the check does not happen after the file is downloaded -- so we are basically running an unchecked binary?

cmdeploy/src/cmdeploy/iroh-relay.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Iroh relay
+
+[Service]
+ExecStart=/usr/local/bin/iroh-relay --config-path /home/iroh-relay/iroh-relay.toml
+WorkingDirectory=/home/iroh-relay/
+Restart=on-failure
+RestartSec=5s
```
hpk42, line 8: does this run as root and if so, does it need to?

cmdeploy/src/cmdeploy/iroh-relay.toml
```diff
@@ -0,0 +1,5 @@
+enable_relay = true
+http_bind_addr = "[::]:3340"
+enable_stun = true
+enable_metrics = false
+metrics_bind_addr = "127.0.0.1:9092"
```
hpk42, line 5: this address must be set even if "enable_metrics" is false?

hpk42 (approved):
looks good.  But i'd keep an opt-out `iroh_relay` setting. i.e. upgrading automatically uses `iroh.{mail_domain}` but if you set `iroh_relay = ` to anything that the systemd-service is not enabled, and Delta Chat users see this setting if it contains something not-empty.

cmdeploy/src/cmdeploy/dns.py
```diff
@@ -6,19 +6,22 @@
...
 
-def check_initial_remote_data(remote_data, print=print):
+def check_initial_remote_data(remote_data, require_iroh, *, print=print):
     mail_domain = remote_data["mail_domain"]
     if not remote_data["A"] and not remote_data["AAAA"]:
         print(f"Missing A and/or AAAA DNS records for {mail_domain}!")
     elif remote_data["MTA_STS"] != f"{mail_domain}.":
         print("Missing MTA-STS CNAME record:")
         print(f"mta-sts.{mail_domain}.   CNAME  {mail_domain}.")
+    elif require_iroh and remote_data["IROH"] != f"{mail_domain}.":
```
missytake, line 22:
```suggestion
    elif require_iroh and remote_data["IROH"] != f"iroh.{mail_domain}.":
```
This needs to compare against the DNS record being `iroh.example.org`, right?

missytake (commented):
Looks good! One thing:

missytake:
> Can also wait for [n0-computer/iroh#2847](https://github.com/n0-computer/iroh/pull/2847) so we don't need a writable config.

I think this is not so important. I'll set myself a reminder to `chmod -w` in 2 months or so.

PR #439, opened 2024-10-27
cmdeploy: add argument to specify different SSH host than mail_domain

### LINKED ISSUE #432: make `cmdeploy run $sshname` work
`cmdeploy run` uses the `mail_domain` from chatmail.ini; for #429 and other purposes it would be good to specify an alternative host.

missytake:
fix #432


hpk42 (approved):

hpk42:
great, but changelog is missing (and important so some people notice it, and we have it in release description)

PR #448, opened 2024-10-31
acmetool: only ask for iroh.$mail_domain cert if iroh_relay isn't set

### LINKED ISSUE #447: cmdeploy run should refuse to run if iroh subdomain isn't set
- expected behavior: instead of failing at the acmetool step which could affect the letsencrypt rate limit, if I have a necessary subdomain like `iroh.$mail_domain` not set, it should refuse to run.
- actual behavior: it just runs. actually not even `cmdeploy dns` warns us about the missing subdomain atm.

missytake:
fix #447

### DIFF
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -528,11 +528,12 @@ def deploy_iroh_relay(config) -> None:
     )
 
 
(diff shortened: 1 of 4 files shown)


missytake:
This is deployed on nine for emergency reasons. So kind of tested already even :upside_down_face:

hpk42 (approved):

PR #454, opened 2024-11-09
Require TLS 1.3 on client-facing ports

link2xt:
I tested with -tls1_2 option
of openssl s_client
that TLS 1.2 connections
are no longer possible
on any ports except port 25.

Port 25 requires at least TLS 1.2
for encrypted connections.

### DIFF
--- a/cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
+++ b/cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
@@ -209,7 +209,7 @@ ssl = required
 ssl_cert = </var/lib/acme/live/{{ config.mail_domain }}/fullchain
 ssl_key = </var/lib/acme/live/{{ config.mail_domain }}/privkey
 ssl_dh = </usr/share/dovecot/dh.pem
-ssl_min_protocol = TLSv1.2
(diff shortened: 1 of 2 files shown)


feld:
I'm not up to speed on how widespread TLSv3 is for SMTP right now. Could this break the ability to send messages to other mail servers that don't support TLSv3?

hpk42:
I think "client-facing" means here Delta <-> Chatmail server (submission and imap ports)
and not MTA SMTP (port 25) server-to-server connections.

missytake (commented):
I'm a bit hesitant to approve this; some people with outdated desktop clients might have followed our instructions in https://delta.chat/en/2023-12-13-chatmail to setup a nine account. Yes, it's an unlikely scenario, but for them it would break.

And I don't see why it's necessary? Wouldn't any more recent official client use TLSv1.3 anyway? No need to force it server-side.

missytake (approved):
It's fine, we don't support Windows 7 anymore since way before 1.46, the release which supported instant onboarding.

Amzd:
This did break my Desktop (macOS) client 1.46.8 with core v1.142.12 but updating fixed it.

PR #462, opened 2024-12-17
DNS: add --all to cmdeploy dns

### LINKED ISSUE #425: DNS check should report all missing records at once
With the current repo state this happens, CNAME records during initial checks are reported one-by-one:
```
$ scripts/cmdeploy run
[ssh] login to c35.testrun.org
Collecting initial DNS settings........../bin/sh: 1: acmetool: not found
..
Missing MTA-STS CNAME record:
mta-sts.c35.testrun.org.   CNAME  c35.testrun.org.
```

After adding mta-sts:
```
$ scripts/cmdeploy run
[ssh] login to c35.testrun.org
Collecting initial DNS settings........../bin/sh: 1: acmetool: not found
..
Missing www CNAME record:
www.c35.testrun.org.   CNAME  c35.testrun.org.
```

Besides that, there is a `/bin/sh: 1: acmetool: not found` error that is printed in the middle of the output.

missytake:
fix #425

### DIFF
--- a/cmdeploy/src/cmdeploy/dns.py
+++ b/cmdeploy/src/cmdeploy/dns.py
@@ -40,7 +40,7 @@ def get_filled_zone_file(remote_data):
     return zonefile
 
 
-def check_full_zone(sshexec, remote_data, out, zonefile) -> int:
+def check_full_zone(sshexec, remote_data, out, zonefile, all) -> int:
     """Check existing DNS records, optionally write them to zone file
     and return (exitcode, remote_data) tuple."""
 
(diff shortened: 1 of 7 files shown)


cmdeploy/src/cmdeploy/dns.py
```diff
@@ -49,16 +49,20 @@ def check_full_zone(sshexec, remote_data, out, zonefile) -> int:
...
     )
 
+    returncode = 0
     if required_diff:
         out.red("Please set required DNS entries at your DNS provider:\n")
         for line in required_diff:
             out(line)
-        return 1
-    elif recommended_diff:
+        print()
```
hpk42, line 57: this should be 'out("")' or so.

PR #463, opened 2024-12-20
DNS: always show recommended DNS records

missytake:
There is not really a reason not to show the recommended DNS records on `cmdeploy dns`. We can re-introduce the `--all` option or something similar when we think of one.


hpk42 (approved):
lgtm

PR #481, opened 2025-01-29
Replace subject with [...] for outgoing mail

link2xt:
`authclean` cleanup server is used by
reinjecting smtpd running on localhost:10025 by default. It runs after filtermail
and currently removes `Received` header
to avoid leaking IP address.
Can as well be used to replace `Subject` lines
with `Subject: [...]`.
If there are multiple `Subject` lines,
all of them should be replaced.

This allows us to avoid dealing with
localized subjects, including SecureJoin
messages `vc-request` and `vg-request`
which can have Subject lines like
Subject: =?utf-8?q?Nachricht_von_nrn178fi4=40nine=2Etestrun=2Eorg?=


missytake (approved):
Great :) would be nice to have a test which ensures that the subject is overridden...

PR #485, opened 2025-02-14
Simplify SPF record

link2xt:
There is no need to explicitly specify domain for `a` rule.


missytake (approved):
Tried it out with learndmarc.com, works like a charm.

only annoying thing is that if anyone runs `cmdeploy dns` again who configured it already in the past will get a warning to fix what isn't broken.

PR #488, opened 2025-02-16
Remove MTA-STS daemon

### LINKED ISSUE #202: Replace MTA-STS with strict TLS policy
We currently run [`postfix-mta-sts-resolver`](https://pypi.org/project/postfix-mta-sts-resolver/).

MTA-STS is prone to downgrade attacks by blocking HTTPS and with this complicated policy we may still deliver or accept messages without transport encryption.

I suggest that we require TLS for port 25 except for nauta.cu which is known to not support STARTTLS on port 25 (but supports DKIM).

[Postfix documentation](https://www.postfix.org/TLS_README.html) has an opinion:
"You can ENFORCE the use of TLS, so that the Postfix SMTP server announces STARTTLS and accepts no mail without TLS encryption, by setting "smtpd_tls_security_level = encrypt". According to RFC 2487 this MUST NOT be applied in case of a publicly-referenced Postfix SMTP server. This option is off by default and should only seldom be used."

If we drop MTA-STS for TLS policy discovery we still want to keep using it for [MX Host Validation](https://www.rfc-editor.org/rfc/rfc8461#section-4.1) to ensure it is impossible to hijack MX record resolution and make us deliver email to attacker's MX server. This is the difficult part, we may have to write our own policy server that keeps checking MX records against MTA-STS.

Related issue, I suggest that we drop TLS-RPT: #181 (EDIT: it is merged)

Chasquid implements TOFU-like policy, not allowing downgrade: https://blitiri.com.ar/p/chasquid/sec-levels/

We may also rethink requiring TLS1.2 when doing opportunistic encryption: https://github.com/deltachat/chatmail/pull/97#issuecomment-1912621455

Postfix has documentation about TLS and TLS forward secrecy:
- https://www.postfix.org/TLS_README.html
- https://www.postfix.org/FORWARD_SECRECY_README.html

We should also think about using `dane` (opportunistic DANE) as the default, current default is `may`.

link2xt:
Closes #202

Outgoing connections now default to `verify` level. See http://www.postfix.org/postconf.5.html#smtp_tls_security_level for possible levels and check that I selected the correct level.

For incoming connections we do not enforce anything as it does not increase security.

### DIFF
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -224,42 +224,21 @@ def _configure_opendkim(domain: str, dkim_selector: str = "dkim") -> bool:
     return need_restart
 
 
-def _install_mta_sts_daemon() -> bool:
-    need_restart = False
+def _uninstall_mta_sts_daemon() -> None:
+    # Remove configuration.
+    files.file("/etc/mta-sts-daemon.yml", present=False)
 
-    config = files.put(
-        name="upload postfix-mta-sts-resolver config",
-        src=importlib.resources.files(__package__).joinpath(
-            "postfix/mta-sts-daemon.yml"
-        ),
-        dest="/etc/mta-sts-daemon.yml",
-        user="root",
-        group="root",
-        mode="644",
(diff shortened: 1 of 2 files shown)


missytake (approved):
Great :)

PR #500, opened 2025-02-22
Remove unnecessary opendkim milter from reinjecting smtpd

link2xt:
OpenDKIM milter is only needed on port 25,
it seems to do nothing on reinjecting port.


missytake:
CI runs into timeout (I retried once, it happened again), so I guess running the tests manually and look for the reason is necessary.

Issue #489, opened 2025-02-16
Remove mails faster because users keep running out of quota


link2xt:
For chatmail accounts the only options for "Delete Messages from Server" setting are "Automatic" and "At once after Download".

For multi-device setups users should use "Automatic", but they still keep running out of quota:
https://support.delta.chat/t/warning-about-providers-storage-but-delete-from-server-is-enabled/3575

In default configuration messages are deleted after 20 days:
https://github.com/deltachat/chatmail/blob/49907c78a36bbef1cf38d5f4b632e6d5306311e5/chatmaild/src/chatmaild/ini/chatmail.ini.f#L23-L24

A simple solution to keep users running out of quota is to delete large messages faster, e.g. after 7 days if the message is larger than 2 MB.

thob:
I would also favor an option to set the automatic cleansing to like 5, 10, 15, 10 days – in my use case - notifications with images attached – I don't have large attachments but rather many messages containing smaller images.

ccclxxiii:
closed by #504

PR #504, opened 2025-03-03
delete big messages after 7 days

adbenitez:
tackles #489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- cmdeploy dovecot: delete big messages after 7 days
+  ([#504](https://github.com/chatmail/server/pull/504))
+
 - mtail: fix getting logs from STDIN
   ([#502](https://github.com/chatmail/chatmail/pull/502))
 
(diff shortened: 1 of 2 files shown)


missytake (approved):
I did a quick superficial check (without `-delete`) which kind of files this returns on nine, looks like the syntax is fine.

We could probably go over 200k for now, balanced quality images (up to 640 KiB in our settings) should be fine. Let's see how this shows in nine's monitoring?

hpk42:
needs a changelog entry, but looks fine to merge.

adbenitez:
I added changelog entry

ThatLurker:
Hmm could this be made configurable in the config?
So that all the people self hosting can incrase or even lower it if needed.

adbenitez:
Yes, but you could also modify the value directly in your local clone of the repo, this is likely bound to change to a smarter approach (ex. Only remove if inbox is 80% full and space is needed and other heuristics)

r10s:
not sure, it was kind of a hotfix, and will probably not the last one.

i also woule not go into exact details for the info, more sth that gives operators some room, eg "Messages are unconditionally removed 20 days after arriving on the server. much sooner when using single device or if quota may exceed otherwise"

PR #505, opened 2025-03-04
update Let's encrypt Subscriber Agreement

tpokorra:
to avoid the problem:

```
failed to request certificate: terms of service agreement is required, but user declined
```


missytake (approved):
Thanks a lot!

PR #507, opened 2025-03-05
Update __init__.py

troilus:
tested it works on armv7l


missytake (approved):
Thanks for testing! I think we can just merge this, the CI/we can't test this easily anyway.

missytake:
> If iroh-relay does not start, then we should instead say that iroh-relay is not supported on armv7l and make the user disable the relay in the config and make sure deployment works.

@troilus can you check with `systemctl status iroh-relay`?

troilus:
> If iroh-relay does not start, then we should instead say that iroh-relay is not supported on armv7l and make the user disable the relay in the config and make sure deployment works.

iroh-relay-v0.28.1-aarch64-unknown-linux-musl.tar.gz

missytake:
Closing this for now - feel free to re-open it after confirming it works :)

PR #512, opened 2025-03-08
Remove cleanup service from submission ports

link2xt:
It does not work because of `smtpd_proxy_filter`
forwarding the message to filtermail
and we cleanup the message once
filtermail reinjects it on port 10025.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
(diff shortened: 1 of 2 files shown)


missytake (approved):
This only removes things which don't work anyway?

This means that we only clean Received headers on outgoing messages, not incoming ones, right?

Issue #515, opened 2025-03-11
cmdeploy run issues [Contabo]


feld:
A user reported problems trying to deploy a server on a Debian 12 Contabo VPS and I ordered one to reproduce the problems. Here is what I found doing a clean deployment.

1. The README should tell users to set the `www` and `mta-sts` CNAMEs before the first run of cmdeploy to avoid having to hit those errors.

2. Contabo ships broken DNS servers in `/etc/resolv.conf`. This really messes with the deploy in several ways. I think it would be wise to immediately overwrite the system DNS recursors with known good ones (google/8.8.8.8, opendns, etc. something that just works until Unbound is running)

3. Extremely early in the cmdeploy run it needs to start checking and managing some sysctls. Contabo ships `net.ipv6.conf.all.disable_ipv6=1` which prevents Unbound from working after it is installed because it cannot bind to `::1`.  We can just drop a file into `/etc/sysctl.d/` and actually there are some important things that should be changed for hardening the TCP/IP stack that can be discussed at another time. Having a functional IPv6 stack is important regardless of whether or not we intend to use it because of these little issues.

4. `opendkim-genkey` fails because it is being run as user `opendkim` and `/usr/sbin` will not be in the PATH. The command should be updated to use the absolute path: `/usr/sbin/opendkim-genkey`

```
    if not host.get_fact(File, f"/etc/dkimkeys/{dkim_selector}.private"):
        server.shell(
            name="Generate OpenDKIM domain keys",
            commands=[
                f"opendkim-genkey -D /etc/dkimkeys -d {domain} -s {dkim_selector}"
            ],
            _use_su_login=True,
            _su_user="opendkim",
        )
```

After these changes are made the deployment work successfully.

feld:
For the sysctls: Contabo is dropping their settings into the bottom of `/etc/sysctl.d/99-sysctl.conf` so an override needs to be in a file that is loaded later, like `/etc/sysctl.d/9999-chatmail.conf'. The service to restart is `systemd-sysctl.service`. I have tested it works correctly with this naming convention to override the earlier declaration.

adbenitez:
> In general I think we should also advise against using contabo in the README. These errors didn't seem to come up consistently for all contabo VPSs, right? So we can catch some of them, but I'm not convinced we should clean up their mess of breaking Debian 12.

if you are going to do so, also mention that Contabo limits traffic volume in port 25 and eventually you might need to try to convince support of rising limit (which I have failed so far)

[...]

PR #518, opened 2025-03-11
DNS: add 9.9.9.9 to resolv.conf

missytake:
According to #515, some providers misconfigure /etc/resolv.conf. With 9.9.9.9, at least one working DNS resolver is there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@
 - unbound: check that port 53 is not occupied by a different process
   ([#537](https://github.com/chatmail/server/pull/537))
 
+- unbound: before unbound is there, use 9.9.9.9 for resolving
+  ([#518](https://github.com/chatmail/relay/pull/518))
+
 - Limit the bind for the HTTPS server on 8443 to 127.0.0.1 
   ([#522](https://github.com/chatmail/server/pull/522))
   ([#532](https://github.com/chatmail/server/pull/532))
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -592,6 +592,12 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
         ensure_newline=True,
     )
 
+    if host.get_fact(Port, port=53) != "unbound":
+        files.line(
+            name="Add 9.9.9.9 to resolv.conf",
+            path="/etc/resolv.conf",
+            line="nameserver 9.9.9.9",
+        )
     apt.update(name="apt update", cache_time=24 * 3600)
     apt.upgrade(name="upgrade apt packages", auto_remove=True)
 


adbenitez (approved):

missytake:
> How does this interact with installing unbound? We are installing our own resolver on 127.0.0.1 that does DNSSEC checks only to add 9.9.9.9 next to it and send plaintext requests there?

Let me re-run the CI to double-check, but I assume that the unbound installation overwrites /etc/resolv.conf again and our temporary addition of 9.9.9.9 gets removed as soon as we don't need it anymore

missytake:
> > How does this interact with installing unbound? We are installing our own resolver on 127.0.0.1 that does DNSSEC checks only to add 9.9.9.9 next to it and send plaintext requests there?
> 
> Let me re-run the CI to double-check, but I assume that the unbound installation overwrites /etc/resolv.conf again and our temporary addition of 9.9.9.9 gets removed as soon as we don't need it anymore

The assumption was correct:

```
$ ssh staging2.testrun.org cat /etc/resolv.conf  # after adding 9.9.9.9, before the install unbound step
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "resolvectl status" to see details about the actual nameservers.

nameserver 2a01:4ff:ff00::add:2
nameserver 2a01:4ff:ff00::add:1
nameserver 185.12.64.1
nameserver 9.9.9.9
$ ssh staging2.testrun.org cat /etc/resolv.conf  # after the unbound installation
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "resolvectl status" to see details about the actual nameservers.

nameserver 2a01:4ff:ff00::add:2
nameserver 2a01:4ff:ff00::add:1
nameserver 185.12.64.1
```

missytake:
(I'll run a second deploy on staging2.testrun.org to make sure 9.9.9.9 isn't re-added on future runs, when unbound isn't installed freshly)

feld:
My only advice is that if you **append** the DNS server to the end of the resolv.conf it will still slow down a lot of operations when there are broken DNS servers because glibc will try the DNS servers in the resolv.conf in order, but it is hardcoded to try the server 5 times with a 5 second timeout [1]. This means each failure requires you to wait 25 seconds per broken DNS server.

The cmdeploy will check for 4 DNS records early (A, AAAA, mta-sts CNAME, www CNAME) so the deployment is slowed down by at least 200 seconds just for this operation, not including the DNS lookups required to install packages.


[1] https://github.com/bminor/glibc/blob/master/resolv/resolv.h#L68-L73

missytake:
If we write "nameserver 127.0.0.1" to /etc/resolvconf/resolv.conf.d/base after installing unbound, this doesn't make all the other entries disappear, right? I wonder what adds them.

feld:
systemd has overcomplicated all of this, sadly

https://www.freedesktop.org/software/systemd/man/latest/systemd-resolved.service.html#/etc/resolv.conf

missytake:
Now 9.9.9.9 is only used if unbound isn't there yet, and then a temporary addition to /etc/resolv.conf should be fine :) unbound will later overwrite this.

> @missytake But this still looks broken, requests should go to 127.0.0.1, not some 185.12.64.1 :/

This is out of scope for this PR I'd say? hard to say why this happens.

missytake:
Bases on https://github.com/chatmail/relay/pull/537 and should be merged after that I'd say.

hpk42 (approved):
lgtm but i'd prefer Mark or L to confirm.

PR #521, opened 2025-03-14
Chatmaild metrics: fix counting logic

feld:
Ensure the candidates for possible accounts are directories with a "cur" subdir to ensure it is a real maildir

### DIFF
--- a/chatmaild/src/chatmaild/metrics.py
+++ b/chatmaild/src/chatmaild/metrics.py
@@ -11,6 +11,8 @@ def main(vmail_dir=None):
     ci_accounts = 0
 
     for path in Path(vmail_dir).iterdir():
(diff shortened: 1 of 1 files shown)


hpk42:
you also need to change the tests (see isolated chatmaild tests CI)

chatmaild/src/chatmaild/metrics.py
```diff
@@ -11,6 +11,8 @@ def main(vmail_dir=None):
     ci_accounts = 0
 
     for path in Path(vmail_dir).iterdir():
+        if not path.is_dir() or not (path / "cur").is_dir():
```
hpk42, line 14: it's sufficient to directly test the target dir i think 

```suggestion
        if not (path / "cur").is_dir():
```

hpk42:
see #534

PR #522, opened 2025-03-15
[Nginx] Limit the bind for the HTTPS server on 8443 to 127.0.0.1

feld:
I don't see any reason why 8443 should need to be exposed or what benefit there is to binding to IPv6 here when it's always proxied to on 127.0.0.1


missytake:
Can you add a CHANGELOG entry? We forget it all the time

PR #524, opened 2025-03-16
fix: send SNI when connecting to outside servers

### LINKED ISSUE #523: Postfix does not send SNI on port 25
Example of a server that fails returns valid certificate when running `openssl s_client -connect example.org:25 -starttls smtp -verify_hostname example.org -verify_return_error`, but wildcard certificate of the hoster when adding `-noservername` to the command.

Postfix then complains:
```
server certificate verification failed for example.org[A.B.C.D]:25: num=62:hostname mismatch
```

https://www.postfix.org/TLS_README.html says "When usable TLSA records are obtained for the remote SMTP server the Postfix SMTP client sends the SNI TLS extension in its SSL client hello message. This may help the remote SMTP server live up to its promise to provide a certificate that matches its TLSA records."
Unclear why it does not always send SNI, should be somehow configured or patched to do it.

This is with Postfix 3.7.11-0+deb12u1

We should probably set https://www.postfix.org/postconf.5.html#smtp_tls_servername to `hostname`.

link2xt:
Otherwise email providers which allow to bring your own domain and use the same IP addresses for all customers
send wildcard certificate instead of the correct one and Postfix refuses to connect with an error

    server certificate verification failed for example.org[A.B.C.D]:25: num=62:hostname mismatch

Closes #523

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- Send SNI when connecting to outside servers
+  ([#524](https://github.com/chatmail/server/pull/524))
+
 - Pass through `original_content` instead of `content` in filtermail
   ([#509](https://github.com/chatmail/server/pull/509))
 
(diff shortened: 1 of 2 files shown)


hpk42 (approved):
Lgtm

PR #533, opened 2025-03-25
Remove DKIM-Signature from incoming mail after checking

### LINKED ISSUE #530: Remove DKIM signature after checking it
DKIM signature is currently checked by OpenDKIM milter attached to `smtpd` running on port 25. Once it is checked, it is not useful for anything. You cannot change it later because the key may change (see #195). As [RFC 6376](https://www.rfc-editor.org/rfc/rfc6376) states it: "message archiving is not a design goal".

As far as I see from the diagram in https://www.postfix.org/MILTER_README.html#plumbing, `cleanup` happens after `smtpd` is done with the message. `cleanup` is essentially a daemon that puts the message into the queue, so we can remove all unnecessary headers there, including DKIM signature.

This will save users some traffic and server operators some disk space because DKIM headers are quite large compared to other headers.

link2xt:
Closes #530

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/incoming_header_cleanup
+++ b/cmdeploy/src/cmdeploy/postfix/incoming_header_cleanup
@@ -1,0 +1,1 @@
+/^DKIM-Signature:/            IGNORE
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- Remove `DKIM-Signature` from incoming mails after verifying
+  ([#530](https://github.com/chatmail/server/pull/530))
+
 - Send SNI when connecting to outside servers
   ([#524](https://github.com/chatmail/server/pull/524))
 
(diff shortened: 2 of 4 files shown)


cmdeploy/src/cmdeploy/postfix/incoming_header_cleanup
```diff
@@ -0,0 +1 @@
+/^DKIM-Signature:/            IGNORE
```
hpk42, line 1: Shouldn't this also scrub Authentication-Results then?

hpk42 (approved):
lgtm (modulo my review comment)

feld:
Does this actually work? I tried applying this to my server but I still see the DKIM-Signature header in the maildir files on disk

feld:
Local mails didn't get DKIM-Signature stripped and this broke federation for me

hpk42:
this PR is pretty stale and broken.  Better to start a new PR attempting to fix #530

PR #537, opened 2025-03-29
cmdeploy: disable nsd so it doesn't block port 53

missytake:
oops, we never created an issue for this. But in some cases cmdeploy fails to install unbound because nsd is already running on port 53. I made sure that this PR also works if nsd isn't installed and the nsd service doesn't exist.

### DIFF
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -587,6 +587,12 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
     # Run local DNS resolver `unbound`.
     # `resolvconf` takes care of setting up /etc/resolv.conf
     # to use 127.0.0.1 as the resolver.
+    systemd.service(
+        name="Disable nsd if it's running, so it doesn't block port 53",
+        service="nsd.service",
+        running=False,
+        enabled=False,
+    )
     apt.packages(
         name="Install unbound",
         packages=["unbound", "unbound-anchor", "dnsutils"],


hpk42 (commented):
i don't quite understand why NSD can be running. That's not the case in a standard debian 12 install, is it?  If someone runs additional services, does it make sense to disable them "just in case"?

feld:
I think it would be better to detect if some process other than unbound is binding to port 53 and throw an error message. People should not be running other software on their Chatmail servers unless they really know what they're doing, and if we take this step we probably also need to detect: powerdns, bind, maradns, dnsmaq, tinydns/djbdns, etc.

You can detect the process name bound to port 127.0.0.1:53 with this one-liner:

```
> sudo ss -lptn 'src 127.0.0.1:53' | awk 'NR>1 {print $6,$7}' | sed 's/users:(("//;s/".*//'
unbound
```

missytake:
> You can detect the process name bound to port 127.0.0.1:53 with this one-liner:
> 
> ```
> > sudo ss -lptn 'src 127.0.0.1:53' | awk 'NR>1 {print $6,$7}' | sed 's/users:(("//;s/".*//'
> unbound
> ```

Thanks! Now I learned how to write pyinfra facts :) we can now check which service is running on a port, and use this to make sure other ports aren't occupied either. But that's for a future PR.

hpk42 (approved):
looks good, nice "fact" hack :)

missytake:
> looks good, nice "fact" hack :)

tbh it's not really a hack, but how you're supposed to use pyinfra :D I'm thinking about upstreaming it to pyinfra.facts.server, it's quite useful I think

PR #538, opened 2025-03-29
[draft] Reject unencrypted incoming mail

hpk42:
This passes tests but it's not complete because merging would break existing cleartext-usages of newsletter-subscriptions, and more, importantly, of throwaway login-emails.

### DIFF
--- a/chatmaild/src/chatmaild/filtermail.py
+++ b/chatmaild/src/chatmaild/filtermail.py
@@ -209,7 +209,6 @@ class BeforeQueueHandler:
         mail_encrypted = check_encrypted(message)
 
         _, from_addr = parseaddr(message.get("from").strip())
-        envelope_from_domain = from_addr.split("@").pop()
 
         logging.info(f"mime-from: {from_addr} envelope-from: {envelope.mail_from!r}")
         if envelope.mail_from.lower() != from_addr.lower():
@@ -223,26 +222,16 @@ class BeforeQueueHandler:
         if envelope.mail_from in self.config.passthrough_senders:
             return
 
-        passthrough_recipients = self.config.passthrough_recipients
-
         if mail_encrypted or is_securejoin(message):
             return
 
+        passthrough_recipients = self.config.passthrough_recipients
+
         for recipient in envelope.rcpt_tos:
-            if envelope.mail_from == recipient:
-                # Always allow sending emails to self.
-                continue
             if recipient_matches_passthrough(recipient, passthrough_recipients):
                 continue
-            res = recipient.split("@")
-            if len(res) != 2:
-                return f"500 Invalid address <{recipient}>"
-            _recipient_addr, recipient_domain = res
-
-            is_outgoing = recipient_domain != envelope_from_domain
-            if is_outgoing:
-                print("Rejected unencrypted mail.", file=sys.stderr)
-                return f"500 Invalid unencrypted mail to <{recipient}>"
+            print("Rejected unencrypted mail.", file=sys.stderr)
+            return "500 Invalid unencrypted mail"
 
 
 class SendRateLimiter:
--- a/chatmaild/src/chatmaild/filtermail.py
+++ b/chatmaild/src/chatmaild/filtermail.py
@@ -157,9 +157,12 @@ def check_encrypted(message):
     return True
 
 
-async def asyncmain_beforequeue(config):
-    port = config.filtermail_smtp_port
-    Controller(BeforeQueueHandler(config), hostname="127.0.0.1", port=port).start()
+async def asyncmain_beforequeue(config, mode):
+    if mode == "outgoing":
+        port = config.filtermail_smtp_port
+    else:
+        port = config.filtermail_smtp_port_incoming
+    Controller(BeforeQueueHandler(config, mode), hostname="127.0.0.1", port=port).start()
 
 
 def recipient_matches_passthrough(recipient, passthrough_recipients):
@@ -172,8 +175,9 @@ def recipient_matches_passthrough(recipient, passthrough_recipients):
 
 
 class BeforeQueueHandler:
-    def __init__(self, config):
+    def __init__(self, config, mode):
         self.config = config
+        self.mode = mode
         self.send_rate_limiter = SendRateLimiter()
 
     async def handle_MAIL(self, server, session, envelope, address, mail_options):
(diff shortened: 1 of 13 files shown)


cmdeploy/src/cmdeploy/tests/plugin.py
```diff
@@ -62,7 +62,7 @@ def sshdomain(maildomain):
 def maildomain2():
     domain = os.environ.get("CHATMAIL_DOMAIN2")
     if not domain:
-        pytest.skip("set CHATMAIL_DOMAIN2 to a ssh-reachable chatmail instance")
+        pytest.skip("set CHATMAIL_DOMAIN2 to second chatmail server (FQDN form)")
```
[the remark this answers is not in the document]
hpk42: FQDN does not imply there is a trailing dot, does it?  It's not sufficient to use an ssh-name because the test attempts to create an e-mail address

chatmaild/src/chatmaild/user.py
```diff
@@ -13,6 +13,7 @@ def __init__(self, maildir, addr, password_path, uid, gid):
         self.maildir = maildir
         self.addr = addr
         self.password_path = password_path
+        self.enforce_E2EE_path = maildir.joinpath("enforceE2EE")
```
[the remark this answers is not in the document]
hpk42: i wanted to have a short name.  E2EE  in the context of chatmail is defined as OpenPGP. don't think there is ambiguity.  maybe `enforceE2EEincoming` -- don't want to introduce Camelcase.

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -230,6 +250,61 @@ def check_DATA(self, envelope):
         for recipient in envelope.rcpt_tos:
             if recipient_matches_passthrough(recipient, passthrough_recipients):
                 continue
+            p = self.config.mailboxes_dir.joinpath(recipient, "inclear")
```
adbenitez, line 253: what is this "inclear"??? sorry if I overlooked the explanation somewhere :thinking:

cmdeploy/src/cmdeploy/service/filtermail-incoming.service.f
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Incoming Chatmail Postfix before queue filter 
+
+[Service]
+ExecStart={execpath} {config_path} incoming
```
adbenitez, line 5: this could have been a single service, but I guess it is interesting to see which one causes more memory usage, OTOH what is likely to happen is that both will use more memory usage than having a single service

adbenitez:
it seems to be missing some explanation/docs about how to use the `enforceE2EE` file hack and there seems to also be an "inclear" file hack?

chatmaild/src/chatmaild/filtermail.py
hpk42: no, it's a very good point! this is left-over and needs to change, but i first need to figure out why no test complained.

hpk42: fixed in https://github.com/chatmail/server/pull/538/commits/e74e942ed2131fc5fcf1ab2f2f3e4f3e4d6ef843

hpk42:
> it seems to be missing some explanation/docs about how to use the `enforceE2EE` file to allow unencrypted etc. and there seems to also be an undocumented "inclear" file hack?

added a section in the README https://github.com/chatmail/server/pull/538/commits/4b0d7a0fa282f72341fe7466d244f96249ae710e  -- (and removed the wrong "inclear" left-overs in the previous commit)

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -214,24 +236,75 @@ def check_DATA(self, envelope):
...
+        if message.get("auto-submitted"):
+            _, from_addr = parseaddr(message.get("from").strip())
+            if from_addr.lower().startswith("mailer-daemon@"):
+                return
 
         for recipient in envelope.rcpt_tos:
-            if recipient_matches_passthrough(recipient, passthrough_recipients):
+            print(f"filtering recipient: {recipient}")
+            if self.config.get_user(recipient).is_incoming_cleartext_ok():
                 continue
```
missytake, line 304: Does envelope.rcpt_tos only contain recipients on the own chatmail server? Others wouldn't be in our vmail directory and we couldn't create a user object for them...

missytake (approved):
Looks reasonable, all in all :)

chatmaild/src/chatmaild/filtermail.py
hpk42: i think postfix's smtpd on port 25, before it passes the mail to the filtermail-incoming service, checks whether the recipient mailboxes exist: when i tried to send a cleartext message to a not-existing address, it was rejected and filtermail never saw the mail. but i'll modify this to be on the safe side.

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -214,24 +236,76 @@ def check_DATA(self, envelope):
...
+        # we want cleartext mailer-daemon messages to pass through
+        # chatmail core will typically not display them as normal messages
+        if message.get("auto-submitted"):
+            _, from_addr = parseaddr(message.get("from").strip())
+            if from_addr.lower().startswith("mailer-daemon@"):
+                return
 
         for recipient in envelope.rcpt_tos:
-            if recipient_matches_passthrough(recipient, passthrough_recipients):
+            print(f"filtering recipient: {recipient}")
```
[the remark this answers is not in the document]
hpk42: yes but i removed it.  it's too verbose.

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -214,24 +236,76 @@ def check_DATA(self, envelope):
...
             return
 
-        passthrough_recipients = self.config.passthrough_recipients
+        print("Incoming: Filtering unencrypted mail.", file=sys.stderr)
+
+        # we want cleartext mailer-daemon messages to pass through
+        # chatmail core will typically not display them as normal messages
+        if message.get("auto-submitted"):
+            _, from_addr = parseaddr(message.get("from").strip())
+            if from_addr.lower().startswith("mailer-daemon@"):
```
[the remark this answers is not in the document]
hpk42: checking now for it.

PR #539, opened 2025-04-01
use "relay" instead of "server" in most places, and "address" instead of "account" 

hpk42:


README.md
```diff
@@ -324,20 +325,20 @@ this is ensured by `filtermail` proxy.
...
 
-You can test it by resolving `MX` records of your server domain
-and then connecting to MX servers (e.g `mx.example.org`) with
+You can test it by resolving `MX` records of your relay domain
+and then connecting to MX relays (e.g `mx.example.org`) with
 `openssl s_client -connect mx.example.org:25 -verify_hostname mx.example.org -verify_return_error -starttls smtp`
 from the host that has open port 25 to verify that certificate is valid.
 
-When providing a TLS certificate to your server,
+When providing a TLS certificate to your chatmail server,
```
missytake, line 336:
```suggestion
When providing a TLS certificate to your chatmail relay, 
```

or "chatmail relay server" maybe? We could also specifically talk about nginx, postfix + dovecot here, but that would be overkill I think.

README.md
```diff
@@ -403,46 +404,47 @@ in this case, just run `ssh-keygen -R "mail.example.org"` as recommended.
...
+   If other MTAs try to deliver messages to your chatmail domain they may fail intermittently,
+   as DNS catches up with the new site settings 
+   but normally will retry delivering messages
    for at least a week, so messages will not be lost.
 
-5. Finally, you can execute `cmdeploy run --ssh-host 13.12.23.42` to turn on chatmail on the new server.
-    Your users will be able to use the chatmail server as soon as the DNS changes have propagated.
-    Voilà!
+5. Finally, you can execute `cmdeploy run --ssh-host 13.12.23.42` to turn on chatmail on the new relay.
+   Your users will be able to use the chatmail site as soon as the DNS changes have propagated.
```
missytake, line 415:
```suggestion
   Your users will be able to use the chatmail relay as soon as the DNS changes have propagated.
```

Many users will not access the "site" at this point; mostly "site" is fine for the migration guide, as "site" also has this geographical connotation, but we should only user it when it's admin-facing, not user-facing imho.

missytake (approved):
Some comments :) overall I'm not convinced of the term; but as it is not as common as "server", maybe we can fill it with our own meaning.

hpk42:
thanks missytake -- addressed your suggestions in a commit.  I think it's fine to use "chatmail relay servers" sometime.  Relays are also servers in today's reality, even if they don't function as a traditional e-mail server with accounts.

PR #544, opened 2025-04-02
Postfix master.cf: use 127.0.0.1 for consistency

feld:
Elsewhere in the master.cf we intentionally use 127.0.0.1 instead of localhost which is useful because it's possible that sometimes it could initially fail if IPv6 is enabled on the OS but these services are not bound to ::1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@
 - Send SNI when connecting to outside servers
   ([#524](https://github.com/chatmail/server/pull/524))
 
+- postfix master.cf: use 127.0.0.1 for consistency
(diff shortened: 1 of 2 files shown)


missytake:
I don't understand why the CI isn't running, you should have access to the github action secrets. I'll push a commit on top of it to check it

missytake:
In https://github.com/chatmail/relay/pull/551 the CI works - it seems you need to start the PR from a branch, not from a fork.

I'll just merge this as the CI passed in https://github.com/chatmail/relay/pull/551, and in the future creating a fork makes more sense for you? now that you have access :)

PR #548, opened 2025-04-07
Do not encourage non-random addresses and weak passwords

link2xt:


r10s (approved):
makes sense to me.

also lowers the incentive to spell the address to someone, which is not working at the end

adbenitez (approved):

hpk42 (approved):

missytake (approved):
So many green checkmarks.

PR #549, opened 2025-04-08
add a hint that deletion may be earlier

r10s:
there is another mention of times in privacy.md,
however, there the gist is about that things are deleted, it is fine if that happens earlier there (also it is not excluded).

targets discussion from https://github.com/chatmail/relay/pull/504

### DIFF
--- a/www/src/info.md
+++ b/www/src/info.md
@@ -15,9 +15,10 @@ for the usage in chats, especially DeltaChat.
 
 - You may send up to {{ config.max_user_send_per_minute }} messages per minute.
 
-- Messages are unconditionally removed {{ config.delete_mails_after }} days after arriving on the server.
-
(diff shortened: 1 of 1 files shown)


missytake (approved):

r10s:
seems https://nine.testrun.org/info.html is not updated automatically - also the change from #548 9 hours ago is not live yet.

is that expected? just asking, i am fine if that needs manual deploy on new, tagged releases or so

missytake:
Yeah, we don't deploy nine.testrun.org automatically per GitHub CI - for a website it would be fine I guess, but as of now that would mean deploying the whole chatmail relay service

r10s:
makes totally sense to not deploy automatically and break it randomly. thanks for explaining

PR #552, opened 2025-04-10
cleanly time out trying to connect to port 25 

hpk42:
treat failure to connect to port 25 as "skip" test result. It's often happening with VPNs or some providers.


missytake (approved):
Looks good! Also works with riseup VPN

PR #553, opened 2025-04-10
fix timezone handling such that client/server do not need to have same

hpk42:


missytake (approved):
Didn't test, but it is in any case less broken than before :)

PR #556, opened 2025-04-11
chatmaild: allow echobot to receive unencrypted messages by default

missytake:
follow-up for #535 and #538

### DIFF
--- a/chatmaild/src/chatmaild/user.py
+++ b/chatmaild/src/chatmaild/user.py
@@ -58,7 +58,8 @@ class User:
             if not self.addr.startswith("echo@"):
                 logging.error(f"could not write password for: {self.addr}")
                 raise
-        self.enforce_E2EE_path.touch()
+        if not self.addr.startswith("echo@"):
+            self.enforce_E2EE_path.touch()
 
     def set_last_login_timestamp(self, timestamp):
         """Track login time with daily granularity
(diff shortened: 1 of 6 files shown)


missytake:
This is mostly useful to test whether a specific server is responsive; for actual users it is less useful. But some operators know about it and use it every once in a while as part of a quick health check, me included.

chatmaild/src/chatmaild/user.py
```diff
@@ -58,7 +58,8 @@ def set_password(self, enc_password):
             if not self.addr.startswith("echo@"):
                 logging.error(f"could not write password for: {self.addr}")
                 raise
-        self.enforce_E2EE_path.touch()
+        if not self.addr.startswith("echo@"):
+            self.enforce_E2EE_path.touch()
```
hpk42, line 62: would rather do this in echo-bot setup (chatmaild/echo.py) not in this generaly code path

missytake: echo.py is run by the `echobot` user, which doesn't have write permission in this path... we could do stuff with `setfacl` I guess, as this is anyway installed to get echobot running, but that sounds overly complicated.

hpk42 (approved):
still not super-happy about how echo-bot is special-cased in general code paths because of permissions but ok for now.

PR #567, opened 2025-04-26
Only update sysctl settings if needed

adonm:
If running in a constrained environment (e.g. an incus / systemd container), setting sysctl limits is constrained, this tweak just checks existing settings and if large enough continues instead of applying


missytake (approved):
Sounds good :) hope it works for you

adonm:
Working well so far :) (ps love the project - great safe way for our kids to chat with us!)

https://chat.adonm.dev/ is the instance I spun up running out of an incus container

PR #568, opened 2025-04-28
cmdeploy: exit if a necessary port is occupied by an unexpected process

### LINKED ISSUE #562: ChatMail destroys existing nginx site configurations
Instead of overwriting /etc/nginx/nginx.conf, why not add configs in /etc/nginx/sites-available and symlink them as is common?

This is just bad and reckless.
How do I get my old nginx.conf back? There is no backup.

There is not even a warning about it. Not that I have seen.

missytake:
fix #562

### DIFF
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -611,10 +611,28 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
     # to use 127.0.0.1 as the resolver.
     from cmdeploy.cmdeploy import Out
 
-    process_on_53 = host.get_fact(Port, port=53)
-    if process_on_53 not in (None, "unbound"):
-        Out().red(f"Can't install unbound: port 53 is occupied by: {process_on_53}")
-        exit(1)
+    port_services = [
+        ("master", 25),
+        ("unbound", 53),
+        ("acmetool", 80),
+        ("imap-login", 143),
+        ("nginx", 443),
+        ("master", 465),
+        ("master", 587),
+        ("imap-login", 993),
+        ("iroh-relay", 3340),
+        ("nginx", 8443),
+        ("master", 10025),
+        ("master", 10026),
+        ("filtermail", 10080),
+        ("filtermail", 10081),
+    ]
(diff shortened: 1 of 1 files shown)


missytake:
This is not completely fail-proof as it doesn't check for existing configuration; but e.g. most nginx configs use port 80, so it would error out on that. For the other services it doesn't matter so much, I doubt someone would try to install chatmail relays over an existing postfix server.

missytake:
I also removed a statement which stops nginx so acmetool can be installed. I think we had it in for backwards compability, but by now all chatmail relays should have upgraded; even if not, there is a good error message now.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - If `dns-utils` needs to be installed before cmdeploy run, apt update to make sure it works
   ([#560](https://github.com/chatmail/relay/pull/560))
 
+- Don't deploy if one of the ports is occupied by an unexpected process
```
hpk42, line 8:
```suggestion
- Don't deploy if one of the ports used for chatmail relay services is occupied by an unexpected process
```

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -611,10 +611,28 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
...
+        ("nginx", 443),
+        ("master", 465),
+        ("master", 587),
+        ("imap-login", 993),
+        ("iroh-relay", 3340),
+        ("nginx", 8443),
+        ("master", 10025),
+        ("master", 10026),
+        ("filtermail", 10080),
+        ("filtermail", 10081),
```
hpk42, line 628: this is fragile. If we change master.cf and use different ports, this list here gets stale.  Maybe everything below 1000 is fine to check.

hpk42 (commented):
not sure about this PR, maybe alex or mark have an opinion?

cmdeploy/src/cmdeploy/__init__.py
missytake: They are config values usually, I am now pulling them from the config instead of hardcoding them.

PR #569, opened 2025-04-29
avoid crash on spurious empty file in the pending_notifications dir

adbenitez:
I wanted to create `enforceE2EEincoming` for every maildir folder in `/home/vmail/mail/DOMAIN/` I was not aware that there is a `pending_notifications` folder there, hard to realize when you have hundreds of folders there, it could happen to other people in general is better to be more resilient and just ignore such invalid files and delete them instead of crashing forever, dovecot process will fail badly then trying to connect to an nonexistent `/run/chatmail-metadata/metadata.socket`

### DIFF
--- a/chatmaild/src/chatmaild/notifier.py
+++ b/chatmaild/src/chatmaild/notifier.py
@@ -95,7 +95,12 @@ class Notifier:
                 logging.warning(f"removing spurious queue item: {queue_path!r}")
                 queue_path.unlink()
                 continue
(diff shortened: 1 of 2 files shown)


hpk42:
notification queue handling has a lot of tests. could you write another one to test for handling of spurious files there?

adbenitez:
> notification queue handling has a lot of tests. could you write another one to test for handling of spurious files there?

added a test

PR #576, opened 2025-06-02
Increase nginx connection limits

link2xt:


cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
```diff
@@ -2,11 +2,25 @@ load_module modules/ngx_stream_module.so;
...
 pid /run/nginx.pid;
 error_log syslog:server=unix:/dev/log,facility=local3;
 
 events {
-	worker_connections 768;
+        # Increase to avoid errors such as
+        #   768 worker_connections are not enough while connecting to upstream
+        # in the logs.
+        # <https://nginx.org/en/docs/ngx_core_module.html#worker_connections>
+	worker_connections 2048;
```
hpk42, line 23: i'd rather go for 8192 directly.  At the current rate we might otherwise hit the limit already tomorrow morning or so.

hpk42 (approved):

cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
link2xt: The change is per worker, do we really expect 2x users?

PR #577, opened 2025-06-02
increase number of dovecot login connections

hpk42:


cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
```diff
@@ -177,18 +177,21 @@ service auth-worker {
...
-    # High-security mode.
-    # Each process serves a single connection and exits afterwards.
-    # This is the default, but we set it explicitly to be sure.
+    # We are not using "high-security-mode" because even if dovecot
+    # login would be compromised, there are only E2EE messages to be
+    # found or sent, and an attacker doesn't have the key used with this address. 
     # See <https://doc.dovecot.org/admin_manual/login_processes/#high-security-mode> for details.
-    service_count = 1
+    
+    # Performance Mode. Each process serves up to 100 connections and exits afterwards.
```
link2xt, line 185: "High-Performance mode" is enabled with `service_count = 0`: https://doc.dovecot.org/2.3/admin_manual/login_processes/#high-performance-mode
The problem with this is not only security, but possible memory leaks as few processes are never restarted.

Increasing `service_count` to 100 makes the process not exit immediately after logging in a single user, but do it after 100 users. Not sure how it even works with `imap-login` when it has already served 100 connections but still has some open TLS proxied connections open, will it terminate them or take space in the `process_limit` while actually serving less than 100 connections?

cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
```diff
@@ -177,18 +177,21 @@ service auth-worker {
...
+    # Performance Mode. Each process serves up to 100 connections and exits afterwards. 
+    service_count = 100
 
     # Inrease the number of simultaneous connections.
     #
     # As of Dovecot 2.3.19.1 the default is 100 processes.
     # Combined with `service_count = 1` it means only 100 connections
     # can be handled simultaneously.
-    process_limit = 10000
+    # We allow up to 5000 * 100 = 500K connections
```
link2xt, line 193: See above, I'm not sure it actually guarantees 500K connections.
In "high-performance mode" it is `client_limit * process_limit`, it "high-security mode" it is `process_limit`, but with `service_count = 100` it's not clear from the documentation what happens.

Can't we just increase the `process_limit` instead?

cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
link2xt: In https://doc.dovecot.org/2.3/configuration_manual/service_configuration/#service-configuration the documentation says explicitly it's not a good idea to set `service_count` to something other than 0 or 1:
> `service_count` is commonly set to unlimited (0) for these services. Otherwise when the service_count is beginning to be reached, the total number of available connections will shrink. With very bad luck that could mean that all the processes are simply waiting for the existing connections to die away before the process can die and a new one can be created.

link2xt: I made a PR https://github.com/chatmail/relay/pull/578 reconfiguring to `service_count = 0`.

PR #578, opened 2025-06-02
Reconfigure imap-login to high-performance mode

link2xt:
High-security mode could be configured
to handle more connections by increasing process_limit, but has problems logging in many users at once after each Dovecot restart or config reload.

### DIFF
--- a/cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
+++ b/cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
@@ -177,20 +177,34 @@ service auth-worker {
 }
 
 service imap-login {
-    # High-security mode.
-    # Each process serves a single connection and exits afterwards.
-    # This is the default, but we set it explicitly to be sure.
-    # See <https://doc.dovecot.org/admin_manual/login_processes/#high-security-mode> for details.
-    service_count = 1
-
-    # Inrease the number of simultaneous connections.
+    # High-performance mode as described in
+    # <https://doc.dovecot.org/2.3/admin_manual/login_processes/#high-performance-mode>
     #
-    # As of Dovecot 2.3.19.1 the default is 100 processes.
-    # Combined with `service_count = 1` it means only 100 connections
-    # can be handled simultaneously.
-    process_limit = 10000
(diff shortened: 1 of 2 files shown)


adbenitez (approved):

cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
```diff
@@ -177,20 +177,34 @@ service auth-worker {
...
-    # This is the default, but we set it explicitly to be sure.
-    # See <https://doc.dovecot.org/admin_manual/login_processes/#high-security-mode> for details.
-    service_count = 1
-
-    # Inrease the number of simultaneous connections.
+    # High-performance mode as described in
+    # <https://doc.dovecot.org/2.3/admin_manual/login_processes/#high-performance-mode>
+    #
+    # So-called high-security mode described in
+    # <https://doc.dovecot.org/2.3/admin_manual/login_processes/#high-security-mode>
```
hpk42, line 184: doubled link

hpk42 (approved):
please fix comment, otherwise mergeable

cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
link2xt: It's not doubled, that is two different sections.

cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
```diff
@@ -177,20 +177,34 @@ service auth-worker {
...
-    # can be handled simultaneously.
-    process_limit = 10000
+    # Otherwise the whole process eventually dies
+    # with an error similar to
+    #   imap-login: Fatal: master: service(imap-login):
+    #   child 1422951 returned error 83
+    #   (Out of memory (service imap-login { vsz_limit=256 MB },
+    #    you may need to increase it)
+    # and takes down all its TLS connections at once.
+    vsz_limit = 1G
```
missytake, line 202: This sounds a bit dangerous, but if it happens we can still increase it, or look for other performance trade-offs.

PR #573, opened 2025-05-15
cmdeploy: upload chatmail/relay version to /etc 

missytake:
This makes it easier to see which changes have already been deployed.

PR #582, opened 2025-06-04
Checkout non-merge commit in CI

link2xt:


link2xt:
Without the fix the test fails on every PR since https://github.com/chatmail/relay/pull/573

PR #583, opened 2025-06-05
Expire push notification tokens after 90 days

### LINKED ISSUE #480: Expire notification tokens
Currently chatmail does not do anything special to expire stored notification tokens. Tokens are only removed if notifications.delta.chat server responds that the token is gone.

We should make the tokens expire, but after https://github.com/deltachat/deltachat-core-rust/pull/6488 is deployed for sufficient number of users.

link2xt:
Closes #480

### DIFF
--- a/chatmaild/src/chatmaild/metadata.py
+++ b/chatmaild/src/chatmaild/metadata.py
@@ -1,5 +1,7 @@
 import logging
 import sys
+import time
+from contextlib import contextmanager
 
 from .config import read_config
 from .dictproxy import DictProxy
@@ -7,8 +9,15 @@ from .filedict import FileDict
 from .notifier import Notifier
 
 
+def _is_valid_token_timestamp(timestamp, now):
+    # Token if invalid after 90 days
+    # or if the timestamp is in the future.
+    return timestamp > now - 3600 * 24 * 90 and timestamp < now + 60
+
+
 class Metadata:
-    # each SETMETADATA on this key appends to a list of unique device tokens
+    # each SETMETADATA on this key appends to dictionary
+    # mapping of unique device tokens
     # which only ever get removed if the upstream indicates the token is invalid
     DEVICETOKEN_KEY = "devicetoken"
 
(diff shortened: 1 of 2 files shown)


chatmaild/src/chatmaild/metadata.py
```diff
@@ -18,21 +21,42 @@ def __init__(self, vmail_dir):
...
+            now = int(time.time())
+            if isinstance(tokens, list):
+                data[self.DEVICETOKEN_KEY] = tokens = {t: now for t in tokens}
+
+            # Remove tokens that are older than a week
+            # or have invalid timestamp from the future.
+            expired_tokens = [
+                token
+                for token, timestamp in tokens.items()
+                if now > tokens[token] + 3600 * 24 * 7 or tokens[token] > now + 60
```
missytake:
```suggestion
                if tokens[token] < now - 3600 * 24 * 7 or tokens[token] > now + 60
```

I find your option a bit harder to read; let's always compare the token with the modification of `now`.

chatmaild/src/chatmaild/metadata.py
```diff
@@ -18,21 +21,42 @@ def __init__(self, vmail_dir):
...
+    @contextmanager
+    def _modify_tokens(self, addr):
         with self.get_metadata_dict(addr).modify() as data:
-            tokens = data.setdefault(self.DEVICETOKEN_KEY, [])
-            if token not in tokens:
-                tokens.append(token)
+            tokens = data.setdefault(self.DEVICETOKEN_KEY, {})
+            now = int(time.time())
+            if isinstance(tokens, list):
+                data[self.DEVICETOKEN_KEY] = tokens = {t: now for t in tokens}
```
missytake, line 36: woah this is nice. Didn't know you could initialize a dictionary from list comprehension in python. :exploding_head:

missytake (commented):
It's still a draft, right? maybe premature comments.

Only adding a new token expires an old token, right? So the client need send a new token to expire the old one? What does the week have to do with it then?

Or do we plan to expire tokens in `get_token_for_addr`?

link2xt:
> Only adding a new token expires an old token, right? So the client need send a new token to expire the old one? What does the week have to do with it then?

Client is storing the token on each connection since https://github.com/chatmail/core/pull/6488

The only case when the token will never expire is when the user has stopped using push notification tokens completely, e.g. switched to desktop or F-Droid version. This needs to be handled somehow.

r10s:
does this PR mean, if i do not start Delta Chat iOS for a week, and do not get messages in that week, i will not get any more PUSH notification?

link2xt:
Increased it to 90 days.

hpk42:
not sure about 90 days. haven't read the code yet but removing device token if you login but don't upload device token for too long ... should be doable in principle. we know the last login time.

link2xt:
> not sure about 90 days. haven't read the code yet but removing device token if you login but don't upload device token for too long ... should be doable in principle. we know the last login time.

So if you login from desktop without uploading device token for 45 days, what does it mean? That we should remove the phone token?

r10s (approved):
lgtm

link2xt:
Added changelog as well

Issue #574, opened 2025-05-17
Issue with Debian Package dovecot-imapd


pierinh00d:
Hi. Are there any issues with the Debian repositories of Chatmail? During the installation of the dovecot-imapd package, apt returns a GPG signature error:

Get:1 https://download.opensuse.org/repositories/home:/deltachat/Debian_12 ./ dovecot-core 1:2.3.21+dfsg1-3 [4515 kB]
Err:1 https://download.opensuse.org/repositories/home:/deltachat/Debian_12 ./ dovecot-core 1:2.3.21+dfsg1-3
  File has unexpected size (4516048 != 4515220). Mirror sync in progress? [IP: 134.76.12.6 443]
  Hashes of expected file:
   - SHA256:d631fe5ec5574d2d38e889b141cf8ffaab1e949fb2af306669584c3e37e37cc8
   - SHA1:efff845ff20e17d30041e125145330ad5397ddfc [weak]
   - MD5Sum:fe2f23367439220ad944bb2228d20144 [weak]
   - Filesize:4515220 [weak]
Get:2 https://download.opensuse.org/repositories/home:/deltachat/Debian_12 ./ dovecot-imapd 1:2.3.21+dfsg1-3 [1544 kB]
Err:2 https://download.opensuse.org/repositories/home:/deltachat/Debian_12 ./ dovecot-imapd 1:2.3.21+dfsg1-3
  File has unexpected size (1544256 != 1544428). Mirror sync in progress? [IP: 134.76.12.6 443]
  Hashes of expected file:
   - SHA256:834d014e8a91989f19b276cde9ae932cbda0c5461e89b506ab37ef31681e0c53
   - SHA1:4b05e9e6b866c775369360e3e9f30ddd31c976d7 [weak]
   - MD5Sum:3753644edb1682579fcd47cd42b1ee6c [weak]
   - Filesize:1544428 [weak]
E: Failed to fetch https://ftp.gwdg.de/pub/opensuse/repositories/home%3A/deltachat/Debian_12/amd64/dovecot-core_2.3.21%2Bdfsg1-3_amd64.deb File has unexpected size (4516048 != 4515220). Mirror sync in progress? [IP: 134.76.12.6 443]
   Hashes of expected file:
    - SHA256:d631fe5ec5574d2d38e889b141cf8ffaab1e949fb2af306669584c3e37e37cc8
    - SHA1:efff845ff20e17d30041e125145330ad5397ddfc [weak]
[...]

PR #590, opened 2025-06-28
dovecot: install from download.delta.chat instead of opensuse

missytake:
fix https://github.com/chatmail/relay/issues/574

### DIFF
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -346,7 +346,7 @@ def _install_dovecot_package(package: str, arch: str):
         src=url,
         dest=deb_filename,
         sha256sum=sha256,
-        cache_time=60 * 60 * 24 * 30,  # cache the .deb for a month
+        cache_time=9999999999999,  # never redownload the package
     )
 
     apt.deb(name=f"Install dovecot-{package}", src=deb_filename)


link2xt:
Seems there is no integrity check and `apt.deb` has no way to specify the checksum.
Also if some package is already installed, will this overwrite the package or do nothing?

missytake:
> Seems there is no integrity check and `apt.deb` has no way to specify the checksum. 

true; that's bad, but we have TLS encryption to download.delta.chat, that's better than nothing. And surely better than the current regular failing of build.opensuse.org. Creating a proper apt repository at download.delta.chat would be nice (also for deltachat-desktop) but is out of scope for now.

> Also if some package is already installed, will this overwrite the package or do nothing?

It will overwrite the package. (not sure if it will also do this if the installed version is *higher* than the .deb version, but... I can't think of a situation right now how a server would get there.

link2xt:
I think it needs at least a checksum hardcoded. Just downloading a file, checking the checksum and installing with `apt-get install` once the checksum passes.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -318,6 +318,39 @@ def _configure_postfix(config: Config, debug: bool = False) -> bool:
...
+        case ("imapd", "amd64"):
+            sha256 = "8d8dc6fc00bbb6cdb25d345844f41ce2f1c53f764b79a838eb2a03103eebfa86"
+        case ("imapd", "arm64"):
+            sha256 = "178fa877ddd5df9930e8308b518f4b07df10e759050725f8217a0c1fb3fd707f"
+        case ("lmtpd", "amd64"):
+            sha256 = "2f69ba5e35363de50962d42cccbfe4ed8495265044e244007d7ccddad77513ab"
+        case ("lmtpd", "arm64"):
+            sha256 = "89f52fb36524f5877a177dff4a713ba771fd3f91f22ed0af7238d495e143b38f"
+        case _:
+            sha256 = None
```
link2xt: This should throw an error instead, otherwise if there is a missing `case` above, we are not checking the checksum.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -318,6 +318,39 @@ def _configure_postfix(config: Config, debug: bool = False) -> bool:
...
+
+    files.download(
+        name=f"Download dovecot-{package}",
+        src=url,
+        dest=deb_filename,
+        sha256sum=sha256,
+        cache_time=60 * 60 * 24 * 30,  # cache the .deb for a month
```
link2xt: What is the use of redownloading the file every month?

missytake: It was the largest number I dared to enter; without `cache_time` I fear it redownloads it on every deploy. But yeah, we can make it a year I guess? ten years? 9999999999?

cmdeploy/src/cmdeploy/__init__.py
missytake: Instead of throwing an error, we can also install the normal dovecot version from the packages, which comes with the 1 second delay. That's much better; I can't really think of a case where this would happen anyway.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -318,6 +318,40 @@ def _configure_postfix(config: Config, debug: bool = False) -> bool:
...
+        case _:
+            apt.packages(packages=[f"dovecot-{package}"])
+            return
+
+    files.download(
+        name=f"Download dovecot-{package}",
+        src=url,
+        dest=deb_filename,
+        sha256sum=sha256,
+        cache_time=9999999999999,  # never redownload the package
```
link2xt, line 349: This now causes a failure `OverflowError: date value out of range`.

PR #591, opened 2025-06-28
CI: ignore PLC0415 in ruff (imports outside top level)

missytake:
this should make the CI green again.


missytake:
It fixes one CI error, https://github.com/chatmail/relay/pull/590 which is based on this PR fixes the other.

PR #592, opened 2025-06-29
CI: fail on lint issues

missytake:
The `cmdeploy fmt -v` step changes the local checkout of the repository, but returns exit code 0 - and then later, the undeployed changes test fails, with an unhelpful error which doesn't really point to the actual problem (the linting). 

Actually CI should already fail at the linting step.


missytake:
This failed (to be expected, so let's rebase on top of #591 to see whether the undeployed changes test passes now:

missytake:
An idea would also be to just remove the lint check from the deploy-on-staging-CI steps, as the ci.yaml workflows already check it. I guess we don't really want our whole deployment to fail *just* because of a code style disagreement.

PR #593, opened 2025-07-03
Adjust SSG pages & styles

DarkCat09:


DarkCat09:
Notes on the last commit:
* i was particularly surprised by the use of fractional pixels (e.g. `2.5px`) which doesn't make sense (unless a page is scaled in browser)
* relative sizes should be preferred over fixed (px), it's the "best practice"
* different margins on h1-h6 doesn't have any effect as they are merged with paragraphs' margins which always equal to 1em

DarkCat09:
As for https://github.com/chatmail/relay/pull/593/commits/2e8318647a515b94fb42231d9bb6f57ffcc0e071

Sorry, as I see it's the common design for all DeltaChat web pages, I'll return the button border and paddings, but please take a look at what i suggested, a button looks cleaner without a 1px-border (again, not `1.5px`, fractional pixels make no sense) and with wider horizontal padding.

www/src/main.css
```diff
@@ -66,21 +69,27 @@ code {
...
   color: red;
   font-weight: bold;
 }
 
+.link-centered {
+  display: block;
+  width: fit-content;
+  margin: 0 auto;
+  text-align: center;
+}
```
missytake, line 85: This looks better on mobile but weird on desktop. I think I'd drop this suggestion.

www/src/page-layout.html
```diff
@@ -1,24 +1,24 @@
...
-    <li><a href="index.html">home</a></li>
-    <li><a href="info.html">info</a></li>
-    <li><a href="privacy.html">privacy</a></li>
-    <li><a href="https://github.com/deltachat/chatmail">public code ↗</a></li>
-    <li id="domain"><a href="index.html">{{ config.mail_domain }}</a></li>
+  <li><a href="index.html">home</a></li>
+  <li><a href="info.html">info</a></li>
+  <li><a href="privacy.html">privacy</a></li>
+  <li><a href="https://github.com/deltachat/chatmail">public code ↗</a></li>
+  <li id="domain"><a href="index.html">{{ config.mail_domain }}</a></li>
```
missytake: Changing only the indentation improves nothing about the page, but can create merge conflicts for anyone who adjusted the menu (e.g. for translation purposes). Please revert this.

www/src/privacy.md
```diff
@@ -51,11 +49,11 @@ Responsible for the processing of your personal data is:
 E-mail: {{ config.privacy_mail }}
 
 We have appointed a data protection officer:
-
 ```
 {{ config.privacy_pdo }}
 ```
 
+
```
missytake: same with changing blank lines. The markdown rendering equalizes them anyway; the merge conflicts are not worth it.

www/src/privacy.md
```diff
@@ -249,8 +250,8 @@ on these or possible objections.
 A deletion can be made
 directly in the Delta Chat email messenger.
 
-If you have any questions or complaints, 
-please feel free to contact us by email:  
+If you have any questions or complaints,
+please feel free to contact us by email:
```
missytake: same here.

missytake (commented):
Thanks for looking into this. Truth be told, I'm rather sceptical with this PR. What you call a "mess" (please watch your tone while judging the work of others, it can demotivate and is therefore bad practice) is part of what any operator is supposed to change about their instance; some of them are not developers, don't know git, and might not have seen a merge conflict so far. 

Aesthetic preferences, especially if they are only about code style and not about visible changes on the website, don't justify that others have to handle a merge conflict.

This means that a change to a .md or .html file needs to have a good reason, like legal issues with the current privacy policy or displaying a new config value on the info page.

The css changes are what I think could be worthwhile; the most notable difference is that the font sizes are a bit larger. The units are a bit more in the direction of web development best practice, indeed. Those we could probably merge, if someone else thinks it won't cause problems.

missytake:
> As for [2e83186](https://github.com/chatmail/relay/commit/2e8318647a515b94fb42231d9bb6f57ffcc0e071)
> 
> Sorry, as I see it's the common design for all DeltaChat web pages, I'll return the button border and paddings, but please take a look at what i suggested, a button looks cleaner without a 1px-border (again, not `1.5px`, fractional pixels make no sense) and with wider horizontal padding.

maybe I'm holding it wrong, but I don't see much difference? Except this black line to the left which is a bit ugly:

image

DarkCat09:
> Thanks for looking into this. Truth be told, I'm rather sceptical with this PR. What you call a "mess" (please watch your tone while judging the work of others, it can demotivate and is therefore bad practice) is part of what any operator is supposed to change about their instance

Yep, sorry, at that moment I was angry after editing web pages for my instance with different indentation (https://github.com/chatmail/relay/pull/593/commits/79c4e85eeaeefdedb4a5aa2064554b60b4d2863f) and even somehow non-ascii quotes in html (https://github.com/chatmail/relay/pull/593/commits/ade85c6880392690581c65e435309b2897ee0e0f)

DarkCat09:
I glanced at opened PRs and seen no changes correspoding static pages, so I supposed that it's OK to adjust whitespace as there would be no merge conflicts

www/src/page-layout.html
DarkCat09: Anyway, please agree on one code style in the chatmail team so there wouldn't be cases where CSS is 2-space indented, HTML is 4-space indented, some uhmm shell script is tab-indented

DarkCat09:
Another codestyle thing that I've just reverted: single HTML tags. Sometimes they're written as `<tag/>`, sometimes as `<tag />`, sometimes `<tag>`. It doesn't change how a page look, yes, but it does change how much a web developer wants to work with this code

www/src/page-layout.html
missytake: typically we use 4-space indentation. The css was copy-pasted from somewhere else I think, and no idea about the shell script^^

DarkCat09: Didn't check scripts actually, it was just a possible example :)

Okay, got it

PR #600, opened 2025-07-08
doveauth: add invite_tokens to restrict address creation

### LINKED ISSUE #594: [Feature request] Invite-only server
There's a way to restrict any account registration by creating `/etc/chatmail-nocreate`.

I personally don't want to host a public server but instead share it with people who i know.
Maybe implement account creation admin command or invitation codes?
Switching chatmail-nocreate manually is not so convenient i guess.

missytake:
fix #594 

---

To restrict address creation for anyone who doesn't have the invite link/QR code:

1. Use the `invite_token` option to add one or more tokens of your choice to `chatmail.ini`: `invite_token = s3cr3t privil3g3`
    - (recommendation: choose 9 or more letters, or it will be easily bruteforced)
3. Run `scripts/cmdeploy run`
4. Distribute a `dcaccount` invite link/QR code (like the one on your web page) with one of your invite tokens added at the end, for example: `dcaccount:https://example.org/new?s3cr3t`

---

To try it out: [`dcaccount:https://c17.testrun.org/new?asdf`](https://c17.testrun.org/new?asdf)
(Right now broken client-side because of https://github.com/chatmail/core/issues/6988)

image

### DIFF
--- a/www/src/index.md
+++ b/www/src/index.md
@@ -11,6 +11,7 @@ for Delta Chat users.  For details how it avoids storing personal information
 please see our [privacy policy](privacy.html). 
 {% endif %}
 
+{% if not config.invite_token %}
 <a class="cta-button" href="DCACCOUNT:https://{{ config.mail_domain }}/new">Get a {{config.mail_domain}} chat profile</a>
 
 If you are viewing this page on a different device
@@ -23,6 +24,10 @@ you can also **scan this QR code** with Delta Chat:
 🐣 **Choose** your Avatar and Name
 
 💬 **Start** chatting with any Delta Chat contacts using [QR invite codes](https://delta.chat/en/help#howtoe2ee)
+{% else %}
+**To join this instance, you need an invite link or QR code -
+ask the admin for an invite.**
+{% endif %}
 
 {% if config.mail_domain != "nine.testrun.org" %}
 <div class="experimental">Note: this is only a temporary development chatmail service</div>


missytake:
Hm -.-

right now, when I run `curl -X POST https://c17.testrun.org/new?asdf`, it responds with a passphrase that contains `asdf`.

but when I copy the very same URL into Delta Chat's "create new profile > use other server > scan invitation code", it tries to login with a passphrase which doesn't contain `asdf`. It seems the `?asdf` parameter is stripped when Delta Chat sends a POST request?
When I login manually with a passphrase that contains `asdf` it works.

missytake:
So server-side everything works, just client-side it doesn't.

missytake:
> So server-side everything works, just client-side it doesn't.

With 2.x.x it works :) So we can merge this now.

jreoka:
Awesome, can't wait for the merge! Good job!

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -24,10 +24,16 @@ def encrypt_password(password: str):
...
-        return False
+        if not config.invite_token or config.invite_token not in cleartext_password:
+            logging.warning(
+                f"blocked account creation because {NOCREATE_FILE!r} exists."
+            )
+            return False
 
-    if len(cleartext_password) < config.password_min_length:
+    if (
+        len(cleartext_password.replace(config.invite_token, ""))
```
hpk42: this can randomly fail depending on the invite_token (might be short) 
I think there should be a fixed "{invite_token}:password" format, and then do a more specific check.

missytake: I'd leave out the `:`, as it could be part of the password; but yes, the invite_token needs to be at the beginning of the password now, that is what newemail.py generates anyway.

missytake:
> 1. Could you add docs to the README  to tell how to use this invite token?  I am not sure yet i understand how this is overall supposed to be used from admins and users.

Do we want it to be easy to find? Personally I'd leave it out of the README.md and point anyone who really asks for it to this PR description instead.

The chatmail ecosystem will be much more resilient *and* anonymous if people can generally sign up on relays of people they *don't* know. I get why some people need a fence, so I built a fence; but we want to have a public park, not a Kleingartenverein.

chatmaild/src/chatmaild/doveauth.py
hpk42: if someone makes eg a 4-char invite code there might still be random passwords that match it, i am afraid.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -696,7 +696,9 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
         running_service = host.get_fact(Port, port=port)
         if running_service:
             if running_service not in service:
-                Out().red(f"Deploy failed: port {port} is occupied by: {running_service}")
+                Out().red(
+                    f"Deploy failed: port {port} is occupied by: {running_service}"
+                )
```
hpk42: strange that this is part of the PR.

hpk42:
if invite codes are set, then normal QR codes without an invite token (like the one on the web page) don't work anymore, right? 
The web-page needs to not show the QR code then and instead a message that an invite is needed. 

Also, the invite code maybe needs have a minimum length of 9 or so because otherwise someone can brute force account creation (invite tokens with 4 characters for example are probably easy to crack in a few days, even without showing up in monitoring, it's just ~1M combinations, can be tried with one password per second, and in ten days you have the invite token). 
 
Given that the recommended UX flow is the QR code / link, it doesn't matter that it becomes harder to type manually by requiring long invite codes.  Besides, chatmail.ini so far contains no secrets and can be public in some repo even -- with the invite codes it's more sensitive. 

Lastly, i am still grappling with the overall idea and the implications of invite links. Not 100% convinced yet.

cmdeploy/src/cmdeploy/__init__.py
missytake: Alright, I put it into https://github.com/chatmail/relay/pull/633 instead.

missytake:
> if invite codes are set, then normal QR codes without an invite token (like the one on the web page) don't work anymore, right? The web-page needs to not show the QR code then and instead a message that an invite is needed.

Done :)

> Also, the invite code maybe needs have a minimum length of 9 or so because otherwise someone can brute force account creation (invite tokens with 4 characters for example are probably easy to crack in a few days, even without showing up in monitoring, it's just ~1M combinations, can be tried with one password per second, and in ten days you have the invite token).

True, I added a recommendation on token length to the README and the PR description. Enforcing it would add complexity for showing error messages; let's hope admins can read and choose risks.

> Given that the recommended UX flow is the QR code / link, it doesn't matter that it becomes harder to type manually by requiring long invite codes. Besides, chatmail.ini so far contains no secrets and can be public in some repo even -- with the invite codes it's more sensitive.

True, but I don't think it will become a problem. So far I haven't seen a chatmail.ini in a public repository - and if you have public sysadmin docs, you anyway need to think about how to avoid committing secrets. I'd say, if someone asks, we can allow setting invite tokens per environment variable.

> Lastly, i am still grappling with the overall idea and the implications of invite links. Not 100% convinced yet.

Same tbh. This PR is fun to implement though :D and I think the compromise of implementing it but hiding the documentation in a GitHub PR is good enough, as some people seem to need this feature. Then again - once it's in the world, it's hard to put back in the box :/

missytake:
This PR is stale, and makes even less sense with [multi-relay](https://fosdem.org/2026/schedule/event/3F9VTU-deltachat-chatmail-relays-multi-transport/). closing it for now; if we reconsider relay propagation strategies we can re-open it.

PR #599, opened 2025-07-08
dovecot: fix errors on re-deployment

missytake:
Somehow re-installing dovecot from .deb throws errors; in CI we didn't see them, as CI only runs once before resetting the server. If we one day upgrade dovecot, we need to think of something different; for now this should fix the problems on `main`.

PR #601, opened 2025-07-08
Avoid datetime overflow, cache deb for a year instead of 300k years

link2xt:


missytake:
also included in https://github.com/chatmail/relay/pull/599 already

link2xt:
Let's merge 599 instead

PR #606, opened 2025-08-06
Build without ICU

missytake:
testing https://github.com/chatmail/dovecot/pull/4


missytake:
Great, seems to work.

PR #608, opened 2025-08-07
feat: Check whether GCC is installed in initenv.sh

cliffmccarthy:
- The Python modules installed by initenv.sh require a compiler to build.
- Before proceeding with installation of Python dependencies, check whether the 'gcc' command is available by running it with the --version argument.  If it is not available, print a helpful message and exit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- Check whether GCC is installed in initenv.sh
+  ([#608](https://github.com/chatmail/relay/pull/608))
+
 - Expire push notification tokens after 90 days
   ([#583](https://github.com/chatmail/relay/pull/583))
 
--- a/scripts/initenv.sh
+++ b/scripts/initenv.sh
@@ -9,9 +9,9 @@ if command -v lsb_release 2>&1 >/dev/null; then
         echo "You need to install python3-dev for installing the other dependencies."
         exit 1
       fi
-      if ! dpkg -l | grep build-essential 2>&1 >/dev/null
+      if ! gcc --version 2>&1 >/dev/null
       then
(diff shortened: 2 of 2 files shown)


cliffmccarthy:
Because `build-essential` is just a bundle of several other packages, there might be a more specific package (or packages) we could require, but I didn't do all the experimentation that would be necessary to determine exactly which set of packages that would be.  `build-essential` is sufficient in the Debian environment I am working with, and seems to me like a reasonable thing to check for.

cliffmccarthy:
Okay, that makes this a CI-breaking change, so I guess I need to go the route of identifying the more specific set of packages that are needed.  It's probably `gcc`, maybe `make`.  I'll do some more experiments to try to identify exactly what we require.  I bet ubuntu-latest is providing whatever it is, without including the build-essential bundle.

cliffmccarthy:
It appears that the `gcc` package is suficient.  Rather than check for the package, I revised it to attempt to run the command.  Here is the kind of lengthy explanation I gave in the commit message:

For the current set of Python dependencies, without GCC, the build process fails when building the crypt-r package.  According to the error message, on my system the exact command it tries to run is 'x86_64-linux-gnu-gcc', but rather than depend on this variant specifically, the script checks for the generic 'gcc' command, so as to avoid coupling the check to an architecture or operating system.  Similar problems arise if we attempt to check for packages by name; the compiler binary is provided by 'gcc-11', but the symlinks that provide the unversioned commands (as used by the Python build) come from a package named 'gcc'.  Trying to be too precise in what we check for could lead to unnecessary failures in some environments, or become a maintenance challenge in the future.  For that reason, this change simply attempts to run 'gcc' and uses that as a probably-sufficient proxy for having what the Python package install will need.

PR #609, opened 2025-08-07
feat: Automate file ownership setting from host migration process

cliffmccarthy:
- Added a step to deploy_chatmail() that sets ownership of paths that are copied over as part of the upgrade process.
- Removed manual chown step from README.md.

### DIFF
--- a/cmdeploy/src/cmdeploy/__init__.py
+++ b/cmdeploy/src/cmdeploy/__init__.py
@@ -748,6 +748,20 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
     _remove_rspamd()
     opendkim_need_restart = _configure_opendkim(mail_domain, "opendkim")
 
+    #
+    # If this system is pre-populated with data from a previous instance,
+    # we might need to adjust ownership of files.
+    #
+    stateful_paths = {
+        "/etc/dkimkeys": "opendkim",
+        "/home/vmail/mail": "vmail",
+        "/run/echobot": "echobot",
+        "/var/lib/acme": "root",
+    }
+    for stateful_path, path_owner in stateful_paths.items():
+        files.directory(stateful_path)  # In case it doesn't exist yet.
+        server.shell("chown {}: -R {}".format(path_owner, stateful_path))
+
     systemd.service(
         name="Start and enable OpenDKIM",
         service="opendkim.service",
--- a/README.md
+++ b/README.md
@@ -395,15 +395,6 @@ in this case, just run `ssh-keygen -R "mail.example.org"` as recommended.
    Postfix and Dovecot are disabled for now; we will enable them later. 
    We first need to make the new site fully operational. 
 
-3. On the new site, run the following to ensure the ownership is correct in case UIDs/GIDs changed:
-
-   ```
-    chown root: -R /var/lib/acme
-    chown opendkim: -R /etc/dkimkeys
-    chown vmail: -R /home/vmail/mail
-    chown echobot: -R /run/echobot
-   ```
-
 4. Now, update DNS entries. 
 
    If other MTAs try to deliver messages to your chatmail domain they may fail intermittently,


cliffmccarthy:
A possible reason for not proceeding with this change (at least in its current form) is that the `chown` operation is non-idempotent and runs every time the deployment process is run.  This could possibly be time-consuming on a large server with a lot of state data, especially in `/home/vmail/mail`.  I find it helpful to have these steps automated for moving server state to a new host, but I realize this might not be universally applicable, so this should be considered in the context of other chatmail servers that may use this code, especially the main one.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -748,6 +748,20 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
...
+    #
+    stateful_paths = {
+        "/etc/dkimkeys": "opendkim",
+        "/home/vmail/mail": "vmail",
+        "/run/echobot": "echobot",
+        "/var/lib/acme": "root",
+    }
+    for stateful_path, path_owner in stateful_paths.items():
+        files.directory(stateful_path)  # In case it doesn't exist yet.
+        server.shell("chown {}: -R {}".format(path_owner, stateful_path))
```
missytake, line 763: Why not use the `user` parameter of https://docs.pyinfra.com/en/2.x/operations/files.html#files-directory to chown the directory? It's idempotent and saves us an operation :)

missytake (changes requested):

missytake:
And a changelog entry would be nice :)

cmdeploy/src/cmdeploy/__init__.py
cliffmccarthy: I didn't use that because I thought it would only update the directory itself, not the whole tree.  But now I see there is also a "recursive" argument, so I'll test that and see if that will handle everything we need.

cliffmccarthy: It appears the `files.directory()` operation doesn't guarantee the resulting state of the tree ownership.  When run with `recursive=True`, it still only looks at the ownership of the top-level directory to decide whether to take any action.  So if the ownership of the top level is correct, files lower in the tree with incorrect ownership will not get updated.  If the ownership of the top level needs to be changed, then just like `chown -R`, it touches every file in the tree, even ones that had correct ownership to begin with.  (I verified this by looking at inode change times.)

It also appears there is no simple equivalent of the implicit group behaviour of the colon in the chown commands from the README.  We would have to code that ourself or just provide the desired group as additional data in the code.

Maybe the scenario of having mixed ownership in these directories isn't something we see in practice, and the change-or-not functionality of `files.directory()` would do the right thing, but it certainly isn't identical to the chowns currently specified.  I assume the main scenario we're trying to handle is something where usernames and/or UIDs don't match between the systems, and that would probably be a whole-tree mismatch, and might get handled correctly.  But given that "recursive" isn't always recursive, I wouldn't be confident relying on pyinfra to produce the desired outcome.

The change I have proposed here feels like shell scripting in a different form, and that is somewhat regrettable, but I don't see a better way to do this than `chown -R`.  If this change doesn't make sense for the general code base, I'm fine with leaving it as a manual step in the README.  Or maybe we need some sort of separate automation of migration routines, though that seems a deviation from the "declared desired state" nature of the current pyinfra tooling.

cliffmccarthy:
> And a changelog entry would be nice :)

Added, though this will probably become a merge conflict with https://github.com/chatmail/relay/pull/608, so after one is merged, I'll rebase the other (unless we decide not to proceed with this one (or the other one)).

cliffmccarthy:
I should clarify what I mean about `chown` above -- the command *is* in fact idempotent with regard to what it operates on, namely file ownership.  When I say it is not idempotent, I'm referring to the fact that it bumps inode change times as a side effect, which is usually a minor detail but occasionally may be important.  It seems like there should be a word for an operation that takes no action when the target is already in the desired state, but I can't think of one at the moment.

missytake:
> I should clarify what I mean about `chown` above -- the command _is_ in fact idempotent with regard to what it operates on, namely file ownership. When I say it is not idempotent, I'm referring to the fact that it bumps inode change times as a side effect, which is usually a minor detail but occasionally may be important. It seems like there should be a word for an operation that takes no action when the target is already in the desired state, but I can't think of one at the moment.

The word you're looking for is "idempotent" :) but if `chown` also bumps the inode change time, it does not only operate on file ownership, so I'd say it is not idempotent.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -748,6 +748,20 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
     _remove_rspamd()
     opendkim_need_restart = _configure_opendkim(mail_domain, "opendkim")
 
+    #
+    # If this system is pre-populated with data from a previous instance,
+    # we might need to adjust ownership of files.
+    #
+    stateful_paths = {
+        "/etc/dkimkeys": "opendkim",
+        "/home/vmail/mail": "vmail",
```
missytake, line 757: I just realized - `/home/vmail/mail` can be *very* large. On large instances, chowning it can take several minutes (e.g. almost 48 minutes on nine.testrun.org); not something we should call on every `cmdeploy run`.

I suggest that we only call the chown steps when `--disable-mail` was passed? Then they will not be executed during *every* deploy.

cmdeploy/src/cmdeploy/__init__.py
missytake: You are correct, let's use your `server.shell` call for now. I filed a bug in pyinfra: https://github.com/pyinfra-dev/pyinfra/issues/1435

cmdeploy/src/cmdeploy/__init__.py
cliffmccarthy: I suspected this could be high overhead on a server like that, so I agree that we should make this something that runs outside the normal operational flow.  It seems like there are probably other use cases for `--disable-mail` where you still wouldn't want to incur that 48-minute overhead, so I'm starting to think we need a different "verb" for fixing up the state of a migrated server, to be called as a separate step.  Or we just keep those steps manual in the docs.

Thanks for submitting the issue in pyinfra!

cliffmccarthy:
Another possibility for how to deal with this occurred to me.  What if we establish standard numeric UIDs for the users that own these directories?  Then, any move between hosts that follow the standard would not need the chown.  I assume there is the potential for these UIDs to "drift" depending on what exact set of packages is installed on the host, and that's what we're trying to address with the chown.  Do we have a requirement to accommodate systems where these usernames are pre-existing with arbitrary UIDs?  If not, forcing specific UIDs could be a solution.  (And then we would need to re-order the migration steps to do the tar after the initial `cmdeploy run` so that the users exist.)

In fact, is doing the tar after creating the users (with arbitrary UIDs) sufficient to ensure proper ownership?  I haven't tested this, but I *think* tar will attempt to use the username embedded in the archive, so as long as the same username exists on both sides, it should come across correctly.  Is there a reason we have to populate the stateful paths before the first `cmdeploy run`, or is re-ordering the operations an option?

(There may still be other non-`tar` methods of transporting the data between systems that are sensitive to numeric UID, but with the `tar`-based instructions in the README, I think establishing the necessary users first might be sufficient.  (And should we still document the ownership requirements, because of that?))

missytake:
This PR is pretty stale and we have too many open PRs, closing it for now. If new info arises we can re-open it :)

Issue #604, opened 2025-07-10
`--ssh-host` does not actually change the ssh host


gracen-writes-code:
- Server OS (Operating System) - preferably Debian 12: Alpine Linux 3.22
- On which OS you run cmdeploy: Alpine Linux 3.22
- chatmail/relay version: `46d2334e9c6909ae5192eac62f3bd30d12d3342e`

## Expected behavior

cmdeploy uses the host provided by `--ssh-host` when connecting to the server for installation.

## Actual behavior

cmdeploy still attempts to connect to the public-facing hostname passed to `cmdeploy init` instead of the provided SSH host.

### Steps to reproduce the problem:

1. `cmdeploy init <some hostname>`
2. `cmdeploy run --ssh-host <different hostname>`

### Logs

I unfortunately lost the logs (I wasn't planning on filing an issue when I initially noticed this), but I can't imagine they'd be necessary.

## Notes

I noticed this issue while trying to set up chatmail in a VM, where the VM would have no publicly available SSH access. I needed to set it up in a VM because I already had postfix running on my server. I'd imagine this would be a common use-case for server owners who already have existing mail infrastructure, but want to host a chatmail relay without having to use a different machine for it.

missytake:
Thanks for submitting this - I will look into the problem, it would indeed be nice to run `cmdeploy` on the target machine itself.

For now just two notes on your environment:
- using Alpine will not work - the `cmdeploy` setup script depends on `apt`, we recommend (and test only with) Debian 12 systems.
- deploying a chatmail relay to a host which is already used for other services will likely not work either, and we don't recommend it. Installing it touches many config files which other services might depend upon, for example the nginx config. If you use any of the ports 25, 53, 80, 143, 443, 465, 587, 993, 3340, or 8443 already, expect cmdeploy to fail; but worse, right now it doesn't fail with a useful warning, but will overwrite config files and might crash (see https://github.com/chatmail/relay/pull/568 for this).

So exactly your last comment "this would be a common use-case for server owners who already have existing mail infrastructure, but want to host a chatmail relay without having to use a different machine for it", well, this will likely not work. using a separate machine is *way less trouble* than trying to reconcile the different philosophies of running mail servers with each other. It doesn't have to be a strong machine; just use a raspi or the cheapest VPS you can find.

gracen-writes-code:
[...]

PR #614, opened 2025-08-09
Add installation using Docker Compose and many other improvements for implementation in Docker.

Keonik1:
Hi, when I decided to set up a chatmail server for personal use,  I was frustrated to discover that there was no Docker installation, so I created a preliminary implementation.

### Brief description of changes
- Add installation via docker compose (MVP 1)
- Add markdown tabs blocks
- Fix [Issue 604](https://github.com/chatmail/relay/issues/604)
- Add `--skip-dns-check` argument to `cmdeploy run` command
- Add `--force` argument to `cmdeploy init` command
- Add startup for `fcgiwrap.service`
- Add extended check when installing `unbound.service`
- Add configuration parameters (`is_development_instance`, `use_foreign_cert_manager`, `acme_email`, `change_kernel_settings`, `fs_inotify_max_user_instances_and_watchers`)

### Full description of changes
The main thing that was done was to add support for running in Docker, but in addition to that, the problems that made this method of running impossible were also solved. 

I think the advantages of running in Docker are obvious:
- faster deployment from scratch
- greater repeatability of runs
- easier migration
- better isolation
- more comfortable management

This is only the first implementation of installation using Docker, but it works and seems to be stable (at least I haven't found any anomalies), but in the future it will need to be brought to a more adequate state (with the abandonment of using systemd inside and the division of everything into several containers or 1 container, but without systemd management, but that's just a thought for the future).

The modifications were made with a focus on **full compatibility** with the current installation and are additional optional parameters.

In addition, some QoL features and instance documentation unification were added so that administrators would not have to create separate pages for different language groups of users.

Some errors that I encountered during deployment have also been fixed (and to be honest, I don't understand why some of them didn't occur for others -_-), namely:
- [Issue 604](https://github.com/chatmail/relay/issues/604) with a non-working `--ssh_host` argument
- Manual launch of the `fcgiwrap.service` service, which does not download or start when rolled out using the current method
- Use of a third-party certificate manager
- Error when installing `unbound.service`, when it is installed but does not show that it occupies port 53

All changes are described in detail in the `changelog.md`.

---
I hope my improvements meet your requirements and you accept the PR, because I really believe that this is a very important part of the project's development — installation using Docker is easier, which means that many more people will be able to install it and thus popularize it.


hpk42:
great work and offer of contribution, @Keonik1 -- the silence since friday is more accidental but several people are excited about this PR and want to try it out and feedback.  Not doing that myself right now but wanted to let you know :)

cliffmccarthy:
I am very interested in deploying with containers as well.  In particular, I would like to decouple from systemd and the host operating system, so that the software content of the components can be managed independently.  Doing that will require an understanding of what all the components are, and toward that end, I created https://github.com/chatmail/relay/pull/615 to document what I know so far.

I like the re-use here of the pyinfra-based script to populate the container; I think sharing the process as much as possible between containers and host-based deploys is a good idea.  I think with some revisions the process could use pyinfra's `@local` functionality, which would be beneficial with or without containers.  Thanks for getting things started on the path towards container-based deploys!

docs/DOCKER_INSTALLATION_EN.md
```diff
@@ -0,0 +1,310 @@
...
+
+Before starting traefik, configuration files must be prepared; otherwise, it will not start correctly.
+
+First, run these commands in the console, replacing their values with the correct ones:
+
+```shell
+export YOUR_EMAIL=your_email@gmail.com
+mkdir -p "./data/traefik"
+cd "./data/traefik"
+```
```
missytake: Can't we simply commit the traefik config to this repository? The only variable I see is the Let's Encrypt email address, and we can just leave it out, no? With a well configured ACME client which requests new certs on time, you don't need email notifications by letsencrypt.

docs/DOCKER_INSTALLATION_EN.md
```diff
@@ -0,0 +1,310 @@
...
+
+    current_hash=$(calculate_hash)
+    previous_hash=$current_hash
+
+    while true; do
+        current_hash=$(calculate_hash)
+        if [[ "$current_hash" != "$previous_hash" ]]; then
+        # TODO: add an option to restart at a specific time interval 
+            echo "[INFO] Certificate's folder hash was changed, restarting nginx, dovecot and postfix services."
+            systemctl restart nginx.service
```
missytake: reloading nginx should be sufficient here, no?

CHANGELOG.md
```diff
@@ -2,6 +2,35 @@
...
+- Add markdown tabs blocks for rendering multilingual pages. Add russian language support to `index.md`, `privacy.md`, and `info.md`.
+  ([#614](https://github.com/chatmail/relay/pull/614))
+
+- Fix [Issue 604](https://github.com/chatmail/relay/issues/604), now the `--ssh_host` argument of the `cmdeploy run` command works correctly and does not depend on `config.mail_domain`.
+  ([#614](https://github.com/chatmail/relay/pull/614))
+
+- Add `--skip-dns-check` argument to `cmdeploy run` command, which disables DNS record checking before installation.
+  ([#614](https://github.com/chatmail/relay/pull/614))
+
+- Add `--force` argument to `cmdeploy init` command, which recreates the `config.ini` file.
```
missytake:
```suggestion
- Add `--force` argument to `cmdeploy init` command, which recreates the `chatmail.ini` file.
```
;)

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -49,6 +49,9 @@
 # Deployment Details
 #
 
+# if set to "True" on main page will be showed dev banner
```
missytake:
```suggestion
# set to "False" to remove the "development instance" banner on the main page.
```
Just an idea, makes a bit clearer what the default option is :)

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -676,8 +677,10 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
     from cmdeploy.cmdeploy import Out
 
     process_on_53 = host.get_fact(Port, port=53)
+    if host.get_fact(SystemdStatus, services="unbound").get("unbound.service"):
+        process_on_53 = "unbound"
     if process_on_53 not in (None, "unbound"):
-        Out().red(f"Can't install unbound: port 53 is occupied by: {process_on_53}")
+        Out().red(f"Can't install unbound: port 53 is occupied by: {process_on_53}")
```
missytake:
```suggestion
        Out().red(f"Can't install unbound: port 53 is occupied by: {process_on_53}")
```

Call me pedantic, but I haate trailing spaces, especially if they clutter `git diff`^^

docker/files/setup_chatmail_docker.sh
```diff
@@ -0,0 +1,74 @@
...
+chown opendkim:opendkim /etc/dkimkeys/opendkim.txt
+
+# TODO: Move to debug_commands after git clone is moved to dockerfile. 
+git config --global --add safe.directory /opt/chatmail
+./scripts/initenv.sh
+
+./scripts/cmdeploy init --config "${INI_FILE}" $INI_CMD_ARGS $MAIL_DOMAIN
+bash /update_ini.sh
+
+./scripts/cmdeploy run --ssh-host localhost --skip-dns-check
```
missytake:
```suggestion
./scripts/cmdeploy run --ssh-host @local --skip-dns-check
```

See https://docs.pyinfra.com/en/2.x/getting-started.html#ad-hoc-commands-with-pyinfra, with `@local` we don't even need SSH afaict.

docker/chatmail_server.dockerfile
```diff
@@ -0,0 +1,102 @@
...
+    apt-get install -y locales && \
+    sed -i -e "s/# $LANG.*/$LANG UTF-8/" /etc/locale.gen && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=$LANG \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update && \
+    apt-get install -y \
+        openssh-client \
+        openssh-server \
```
missytake: If we tell pyinfra in `setup_chatmail_docker.sh` to deploy to `@local`, we don't need to install these I think :)

docker/chatmail_server.dockerfile
```diff
@@ -0,0 +1,102 @@
...
+      curl -fsSL "$url" -o "$file"; \
+      echo "$sha256  $file" | sha256sum -c -; \
+      apt-get install -y "$file"; \
+      rm -f "$file"; \
+    done \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN systemctl enable \
+        ssh \
+        fcgiwrap
```
missytake: Isn't fcgiwrap enabled by cmdeploy later? I think we can drop it here :)

docker/chatmail_server.dockerfile
```diff
@@ -0,0 +1,102 @@
...
+      case "$pkg" in \
+        core) sha256="43f593332e22ac7701c62d58b575d2ca409e0f64857a2803be886c22860f5587" ;; \
+        imapd) sha256="8d8dc6fc00bbb6cdb25d345844f41ce2f1c53f764b79a838eb2a03103eebfa86" ;; \
+        lmtpd) sha256="2f69ba5e35363de50962d42cccbfe4ed8495265044e244007d7ccddad77513ab" ;; \
+      esac; \
+      url="https://download.delta.chat/dovecot/dovecot-${pkg}_2.3.21%2Bdfsg1-3_amd64.deb"; \
+      file="/tmp/$(basename "$url")"; \
+      curl -fsSL "$url" -o "$file"; \
+      echo "$sha256  $file" | sha256sum -c -; \
+      apt-get install -y "$file"; \
```
missytake: isn't it `dpkg -i ` instead of `apt-get install -y`?

docker/chatmail_server.dockerfile
```diff
@@ -0,0 +1,102 @@
...
+    ssh-keygen -P "" -t rsa -b 2048 -f /root/.ssh/id_rsa && \
+    mkdir -p /root/.ssh && \
+    cat /root/.ssh/id_rsa.pub >> /root/.ssh/authorized_keys && \
+    SSH_USER_CONFIG="/root/.ssh/config" && \
+    echo "Host localhost" >  "$SSH_USER_CONFIG" && \
+    echo "    HostName localhost" >>  "$SSH_USER_CONFIG" && \
+    echo "    User root" >>  "$SSH_USER_CONFIG" && \
+    echo "    StrictHostKeyChecking no" >>  "$SSH_USER_CONFIG" && \
+    echo "    UserKnownHostsFile /dev/null" >>  "$SSH_USER_CONFIG"
+    ## TODO: deny access for all insteed root form 127.0.0.1 https://unix.stackexchange.com/a/406264
```
missytake: This could also be left out if pyinfra deploys to `@local` instead per SSH :)

docker/docker-compose-default.yaml
```diff
@@ -0,0 +1,58 @@
+services:
+  chatmail:
+    build: 
+      context: ./docker
+      dockerfile: chatmail_server.dockerfile
```
missytake: can we call this `chatmail_relay.dockerfile?`

docker/files/entrypoint.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ "${USE_FOREIGN_CERT_MANAGER,,}" == "true" ]; then
+    if [ ! -f "$PATH_TO_SSL_CONTAINER/fullchain" ]; then
+        echo "Error: file '$PATH_TO_SSL_CONTAINER/fullchain' does not exist. Exiting..." > /dev/stderr
+        exit 1
+    fi
+    if [ ! -f "$PATH_TO_SSL_CONTAINER/privkey" ]; then
+        echo "Error: file '$PATH_TO_SSL_CONTAINER/privkey' does not exist. Exiting..." > /dev/stderr
+        exit 1
+    fi
+fi
```
missytake: Does it make sense to give more concrete advice on how to resolve this, or is this hard because of the different traefik vs acmetool possibilities?

docker/files/setup_chatmail_docker.sh
```diff
@@ -0,0 +1,74 @@
...
+    debug_commands
+fi
+
+if [ "$FORCE_REINIT_INI_FILE" == "true" ]; then 
+    INI_CMD_ARGS=--force
+fi
+
+/usr/sbin/opendkim-genkey -D /etc/dkimkeys -d $MAIL_DOMAIN -s opendkim
+chown opendkim:opendkim /etc/dkimkeys/opendkim.private
+chown opendkim:opendkim /etc/dkimkeys/opendkim.txt
```
missytake, line 61: Doesn't this happen in `cmdeploy run` anyway?

docker/docker-compose-default.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - /run/lock
+    logging:
+      driver: json-file
+      options:
+        max-size: "10m"
+        max-file: "3"
+    environment:
+      MAIL_DOMAIN: <your_domain>
+      CHANGE_KERNEL_SETTINGS: "False"
+      ACME_EMAIL: <your_email>
```
missytake:
```suggestion
      MAIL_DOMAIN: $MAIL_DOMAIN
      CHANGE_KERNEL_SETTINGS: "False"
      ACME_EMAIL: $ACME_EMAIL
```

`MAIL_DOMAIN` and `ACME_EMAIL` need to be passed to the container correctly.

docker/chatmail_server.dockerfile
```diff
@@ -0,0 +1,102 @@
...
+        rsync \
+        unbound \
+        unbound-anchor \
+        dnsutils \
+        postfix \
+        acl \
+        nginx \
+        libnginx-mod-stream \
+        fcgiwrap \
+        cron \
```
missytake: Why not install these dependencies during the `cmdeploy run` step? This would save us from having double code paths doing the same.

It would also help with the acmetool problem - right now, nginx starts with the default config, binding to port 80 to display a default page. Later, when `cmdeploy run` overwrites the nginx config, the `sites-enabled` directory isn't included anymore; but when the container starts up, nginx is already installed with the default config, and nothing else can bind to port 80. That's unfortunate.

A quickfix would be running `unlink /etc/nginx/sites-enabled/default` after this command.

docker/docker-compose-default.yaml
```diff
@@ -0,0 +1,58 @@
...
+      MAX_MESSAGE_SIZE: "50M"
+      # DEBUG_COMMANDS_ENABLED: "true"
+      # FORCE_REINIT_INI_FILE: "true"
+      # USE_FOREIGN_CERT_MANAGER: "True"
+      # ENABLE_CERTS_MONITORING: "true"
+      # CERTS_MONITORING_TIMEOUT: 10
+      # IS_DEVELOPMENT_INSTANCE: "True"
+    ports:
+      - "25:25"
+      - "587:587"
```
missytake:
```suggestion
    ports:
      - "25:25"
      - "80:80"
      - "587:587"
```

We need port 80 for the acmetool challenge

docs/DOCKER_INSTALLATION_EN.md
```diff
@@ -0,0 +1,310 @@
...
+
+7. Start docker compose and wait for the installation to finish:
+
+```shell
+docker compose up -d # start service
+docker compose logs -f chatmail # view container logs, press CTRL+C to exit
+```
+
+8. After installation is complete, you can open `https://<your_domain_name>` in your browser.
+
```
missytake, line 98: Here we need steps to set the other DNS records. Without DKIM, the relay will not federate correctly. Unfortunately, running `cmdeploy dns` doesn't work well if you deploy to localhost, because it right now assumes that chatmail is running remotely and that it can login with SSH somewhere. So let's just leave this out for now, or recommend some DNS records manually?

www/src/page-layout.html
```diff
@@ -9,7 +9,7 @@
     <title>{{ config.mail_domain }} {{ pagename }}</title>
     <link rel="stylesheet" href="./main.css">
     <link rel="icon" href="/logo.svg">
-    <link rel=”mask-icon” href=”/logo.svg” color=”#000000">
+    <link rel="mask-icon" href="/logo.svg" color="#000000">
```
missytake: I understand why you want to change this, but please revert this. This might cause merge conflicts for other operators when they want to upgrade their instance; many of them don't know git well enough to deal with this. We should not cause a merge conflict so easily.

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -331,8 +361,9 @@ def main(args=None):
         return parser.parse_args(["-h"])
 
     def get_sshexec():
-        print(f"[ssh] login to {args.config.mail_domain}")
-        return SSHExec(args.config.mail_domain, verbose=args.verbose)
+        host = args.ssh_host if hasattr(args, "ssh_host") and args.ssh_host else args.config.mail_domain
+        print(f"[ssh] login to {host}")
+        return SSHExec(host, verbose=args.verbose)
```
missytake: This is a good step towards getting `cmdeploy dns` to run as well... but maybe we can completely avoid logging in to localhost? That would be much simpler. Let's see, for now at least deploying works, after all :)

docker/docker-compose-default.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - "143:143"
+      - "993:993"
```
missytake:
```suggestion
      - "143:143"
      - "465:465"
      - "993:993"
```

And there is no reason not to publish port 465 :) the tests connect to it for example.

docker/files/setup_chatmail_docker.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# TODO: Move to debug_commands after git clone is moved to dockerfile. 
+git config --global --add safe.directory /opt/chatmail
+./scripts/initenv.sh
```
missytake, line 68:
```suggestion
rm -rf venv/
./scripts/initenv.sh
```

`initenv.sh` sometimes fails with `./scripts/initenv.sh: 18: venv/bin/pip: not found`, seems the virtual environment gets corrupted sometimes? I don't have a good explanation, but this makes it less flaky.

...sure, it makes the container take longer to start up, but let's first make it work, and then make it fast?

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -32,17 +32,28 @@ def init_cmd_options(parser):
...
+    inipath = args.inipath
     if args.inipath.exists():
-        print(f"Path exists, not modifying: {args.inipath}")
-        return 1
-    else:
-        write_initial_config(args.inipath, mail_domain, overrides={})
-        out.green(f"created config file for {mail_domain} in {args.inipath}")
+        if not args.recreate_ini:
+            out.green(f"[WARNING] Path exists, not modifying: {inipath}")
+            return 0
```
missytake: This makes `cmdeploy/src/cmdeploy/tests/test_cmdeploy.py::TestCmdline::test_init_not_overwrite` fail - let's adjust that test so it also tests the overwrite variable?

missytake (commented):
Soo many changes :) thanks for all the effort and thoughts that went into this.

Sorry for taking some days to reply, but I wanted to look at it in full detail, and have some fix suggestions already. I got acmetool to work, for example.

The tests are mostly fine with this, but unfortunately many tests try to login via SSH - maybe we can try to run `docker exec` if a chatmail container is running on the same machine?
A github action which deploys the docker container automatically would also be nice :) We can use a hetzner VPS for that, like for the cmdeploy CI. It would be good to have CI notice when something fails with a new change, to avoid regressions.

In general, not bad how many variables and configuration options you added, e.g. that `CHANGE_KERNEL_SETTINGS` is set to False by default in `setup_chatmail_docker`. In most situations the defaults will probably never be overridden, but it's good that it's explicit what's the behavior.

The markdown tabs are also a nice touch, but we might need more discussion to merge them into main. The question is, which languages do you prefer, which do you offer? Only offering english is not much better, sure. It requires some thought, I think, or could go to an example page so admins can choose to have the tabs if they want to be multi-lingual. Right now, many admins just translate the www directory to their mother tongue.

All in all, thanks for all of this! I suggest to address the comments, and merge it to a `docker` branch for now, so we can make it more stable in subsequent PRs. But in any case there will be some interested testers already :)))

Keonik1:
@missytake, Hello!
Thank you for your feedback. I quickly scanned through it and will try to answer all questions and fix everything that needs to be fixed this week. Unfortunately, I only have free time in the evenings, so I hope the fixes won't take too long.

I've also thought about it and will try to send fixes in the format of `1 fix = 1 commit` to make it easier to check.

Keonik1:
@missytake 
> The markdown tabs are also a nice touch, but we might need more discussion to merge them into main. The question is, which languages do you prefer, which do you offer?

i have 2 ideas:
1. First, introduce the languages that we know and can translate well. It is even possible to take a transfer from known current instances.
2. Add a number of the most popular languages and translate them through a neural network or standard translator, and then wait for a native speaker to come and fix the errors :D

It's also probably worth adding the ability to determine which languages will be displayed in the installed instance, so that admins can simply enable or disable if necessary. I'll think about the implementation of this at the end, when I fix everything else.

docs/DOCKER_INSTALLATION_EN.md
Keonik1: i think no? As reference i used [`acmetool.hook`](https://github.com/chatmail/relay/blob/main/cmdeploy/src/cmdeploy/acmetool/acmetool.hook)
```shell
#!/bin/sh
set -e
EVENT_NAME="$1"
[ "$EVENT_NAME" = "live-updated" ] || exit 42
systemctl restart nginx.service
systemctl reload dovecot.service
systemctl reload postfix.service
```

And as i can see that dovecot use fullchain and privkey in the [config file](https://github.com/chatmail/relay/blob/main/cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2#L223-L224), I don't know if he knows how to dynamically apply certificates.

```
ssl_cert = </var/lib/acme/live/{{ config.mail_domain }}/fullchain
ssl_key = </var/lib/acme/live/{{ config.mail_domain }}/privkey
```

missytake: Yeah, seems what acmetool.hook does is overkill as well: https://unix.stackexchange.com/a/247460

typically, reloading is enough to apply renewed TLS certificates :) there are just some services which can't be reloaded, but postfix, dovecot, and nginx can be :)

docker/chatmail_server.dockerfile
missytake: well, apt-get install does work too, apparently :D

docker/files/entrypoint.sh
Keonik1: tl;dr
This "crutch" prevents starting all services before we don't confirm that we have all required files, because on first launch receiving certificates can take a little bit of time

@missytake when I tested this, I got the problem that fullchain and privkey do not mount to the container when it's started (on first launch), because receiving certificates is not an instant process (unbelievable :D). To prevent this, I wrote this check, but it didn't work because at that moment I was directly mounting fullchain and privkey to the container, not a directory. After some debugging, I moved from direct file mounting to mounting a directory but decided that checking certs on startup with a detailed description of what went wrong is better when the user got an error when all was already started.

docker/files/setup_chatmail_docker.sh
Keonik1: No, it will not happen :(
I don't know why. I checked the source code from `cmdeploy`, which states that the keys will be created, and their owner will be changed, but in my case they always belong to root... and this was the only solution that worked for me.

docker/docker-compose-default.yaml
Keonik1: In this case, I didn’t set the variables because it required creating a .env file.
It's not a big problem to change this to use variables — I will do it, but a bit later, because I also need to update the instructions. :D

docker/chatmail_server.dockerfile
Keonik1: The main reason for this decision is to prevent downloading packages every time the container starts.
Yes, we install them inside cmdeploy, but if the packages are already installed, the process will finish much faster.

As for nginx - thanks! I’ll take a look at that.
Although in my case, the problem was that acmetool couldn't obtain certificates at all, even when nginx wasn’t running.
But since I have no prior experience with acmetool, I can't rule out the possibility that the issue was actually caused by the pre-installed nginx.

cmdeploy/src/cmdeploy/cmdeploy.py
Keonik1: I don't know how to do this correctly :(
Can you attach an suggesstion?

Keonik1: Neural network wrote this but I can't check how this works, may be you can?
```python
def test_init_creates_file(self, capsys):
    assert main(["init", "chat.example.org"]) == 0
    out, err = capsys.readouterr()
    assert "created config file for" in out.lower()


def test_init_not_overwrite(self, capsys):
    assert main(["init", "chat.example.org"]) == 0
    capsys.readouterr()

    assert main(["init", "chat.example.org"]) == 0
    out, err = capsys.readouterr()
    assert "[warning] path exists" in out.lower()


def test_init_force_recreates_file(self, capsys):
    assert main(["init", "chat.example.org"]) == 0
    capsys.readouterr()

    assert main(["init", "chat.example.org", "--recreate-ini"]) == 0
    out, err = capsys.readouterr()
    assert "[warning] force argument was provided" in out.lower()
    assert "created config file for" in out.lower()
```

docker/chatmail_server.dockerfile
Keonik1: Wow, the nginx configuration suggestion really helped, thank you. But instead of removing the nginx installation from the dockerfile, I added `unlink /etc/nginx/sites-enabled/default` to `entrypoint.sh`.

Keonik1: Regarding moving package installation from Dockerfile to cmdeploy, I still think it's not worth doing because of performance issues due to having to transfer packages every time it starts up.

If you agree, I think we can close this thread.

Keonik1: Actually, I’ve been thinking about what we can remove from the container to reduce its size, but I don’t have enough expertise in this area. For example, I’m not sure how essential gcc and python-dev are for us. We could easily remove git to free up around 50MB, but then it’s unclear where the product version would come from.

Ideally, we should be putting a pre-built binary into the container, without any build tools. Similar to how it's done with Dovecot. Alternatively, we could use a multi-stage Docker build, but for that, I need to understand what belongs strictly to the "dev" packages and what is required for "production".

That said, I think we can start addressing these questions after we push the first implementation to the repository. Then we can gradually optimize the container, eventually ensuring that it strictly corresponds to a specific version of Chatmail with guaranteed reproducibility.

docker/chatmail_server.dockerfile
Keonik1: I don't think it needs to be redone, at least not right now, so maybe we can close the thread?

docker/chatmail_server.dockerfile
missytake: > Actually, I’ve been thinking about what we can remove from the container to reduce its size, but I don’t have enough expertise in this area. For example, I’m not sure how essential gcc and python-dev are for us. We could easily remove git to free up around 50MB, but then it’s unclear where the product version would come from.

Without gcc and python-dev, scripts/initenv.sh will not run successfully. They are necessary to install cmdeploy via pip.

missytake: > Wow, the nginx configuration suggestion really helped, thank you. But instead of removing the nginx installation from the dockerfile, I added `unlink /etc/nginx/sites-enabled/default` to `entrypoint.sh`.

I think you still need to reload (maybe even restart) nginx in entrypoint.sh, no? otherwise it will still occupy port 80 I think.

missytake: > Regarding moving package installation from Dockerfile to cmdeploy, I still think it's not worth doing because of performance issues due to having to transfer packages every time it starts up.

I think for the beginning, reducing maintenance overhead is more important than startup performance. It's better if we don't have to think about two places when we add/remove a package, for example.

docker/docker-compose-traefik.yaml
```diff
@@ -77,17 +95,20 @@ services:
         max-size: "10m"
         max-file: "3"
     command:
-      - --configFile=/config.yaml
+      - "--configFile=/config.yaml"
+      - "--certificatesresolvers.letsEncrypt.acme.email=${ACME_EMAIL:-my.email@gmail.com}"
```
missytake:
```suggestion
      - "--certificatesresolvers.letsEncrypt.acme.email=${ACME_EMAIL}"
```

Adding a non-working default variable doesn't make sense I think. Does traefik work without an ACME email? That would make more sense I think, to leave out optional entries for now. Especially as the setup refreshes certs on its own anyway.

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -99,7 +99,11 @@ def run_cmd(args, out):
     deploy_path = importlib.resources.files(__package__).joinpath("deploy.py").resolve()
     pyinf = "pyinfra --dry" if args.dry_run else "pyinfra"
     ssh_host = args.config.mail_domain if not args.ssh_host else args.ssh_host
+
     cmd = f"{pyinf} --ssh-user root {ssh_host} {deploy_path} -y"
+    if sshexec == "localhost":
+        cmd = f"{pyinf} @local {deploy_path} -y"
```
missytake, line 107: This is a good quick fix, maybe we can add something like this to the `cmdeploy dns` calls as well, so they don't try to request DNS records from the chatmail relay itself, but from the local (docker) host.

docs/DOCKER_INSTALLATION_EN.md
```diff
@@ -79,113 +81,12 @@ Mandatory variables for deployment via Docker:
 
 - `CHANGE_KERNEL_SETTINGS` – Change kernel settings (`fs.inotify.max_user_instances` and `fs.inotify.max_user_watches`) on startup. Changing kernel settings inside the container is not possible! (default: `False`)
 
-5. Configure environment variables in the `.env` file. These variables are used in the `docker-compose.yaml` file to pass repeated values.
-
 6. Build the Docker image:
 
 ```shell
 docker compose build chatmail
```
missytake:
```suggestion
docker compose build
```

Can be even simpler I think.

docker/docker-compose-traefik.yaml
```diff
@@ -67,6 +69,22 @@ services:
...
+    logging:
+      driver: json-file
+      options:
+        max-size: "10m"
+        max-file: "3"
+    working_dir: /app
+    entrypoint: sh -c '
+      touch acme.json &&
+      sudo chown 0:0 ./acme.json &&
+      sudo chmod 600 ./acme.json'
```
missytake:
```suggestion
      chown 0:0 ./acme.json &&
      chmod 600 ./acme.json'
```

sudo is not available in the container, it's the root user anyway.

docker/files/entrypoint.sh
```diff
@@ -1,13 +1,17 @@
 #!/bin/bash
 set -eo pipefail
 
+unlink /etc/nginx/sites-enabled/default
```
missytake:
```suggestion
unlink /etc/nginx/sites-enabled/default || true
```

Otherwise it fails if the link doesn't exist (anymore).

docker/files/setup_chatmail_docker.sh
missytake: Ah, it seems that the `_su_user` argument doesn't work inside docker containers. When I tried to run it manually, I got:

```
# su -l opendkim -c "/usr/sbin/opendkim-genkey -D /etc/dkimkeys -d docker.testrun.org -s test"
This account is currently not available.
```

But in that case it is totally fine to run it in `setup_chatmail_docker.sh`.

missytake: Maybe add a comment that this is because `_su_user` doesn't work inside docker containers.

missytake:
We're getting there! With the last comments addressed, I suggest to merge it into main, but maybe don't mention it in README.md for now (or only as experimental option). But @hpk42 and @link2xt should also add their opinion on this.

To summarize what's needed in the long run:

- A refactoring of the SSHExec logic, so it also supports running commands locally with `os.subprocess` or so, and with `docker exec chatmail $command`. @hpk42 did the last one and has some opinions about it I think ;)
  - We need this to get `cmdeploy dns` to work, which is a critical part of the setup. Federation will not work if users can't generate an opendkim record. A quick fix like the one in `cmdeploy run` might be good enough for now, I think I'll work on it today or tomorrow :)
  - Getting `cmdeploy test` to work would be necessary as well, so we can check in the CI whether a change breaks one of the setup options. This might be slightly more involved than getting `cmdeploy dns` to work.
- We need to resolve the question what is the priority: code maintainability or container needs like startup speed and traefik support?
  - We should decide whether the traefik setup is going to be part of the supported options, or a suggestion on how to add chatmail to an existing docker host. It adds maintenance load (e.g. checking the traefik setup in the CI).

cmdeploy/src/cmdeploy/cmdeploy.py
missytake: This should be fixed with https://github.com/Keonik1/relay/pull/2 :)

docker/docker-compose-traefik.yaml
Keonik1: Okay, I thought Traefik required an email address, but I checked and it doesn't...

docker/chatmail_server.dockerfile
Keonik1: I need to think about the best way to do this. Because some of the packages still need to be put in the container.

docker/example.env
```diff
@@ -0,0 +1,5 @@
+MAIL_DOMAIN="chat.example.com"
+ACME_EMAIL="my.email@gmail.com"
+
+CERTS_ROOT_DIR_HOST="./traefik/data/letsencrypt/certs"
```
missytake: This variable is a bit confusing if you use the non-traefik-setup... and without traefik, it is actually not used, right?

docker/docker-compose-traefik.yaml
```diff
@@ -0,0 +1,136 @@
...
+      - "25:25"
+      - "587:587"
+      - "143:143"
+      - "465:465"
+      - "993:993"
+    volumes:
+      ## system
+      - /sys/fs/cgroup:/sys/fs/cgroup:rw # required for systemd
+      - ./:/opt/chatmail
+      - ${CERTS_ROOT_DIR_HOST}:${CERTS_ROOT_DIR_CONTAINER}:ro
```
missytake:
```suggestion
      - ./traefik/data/letsencrypt/certs:${CERTS_ROOT_DIR_CONTAINER}:ro
```

Then we could also just hardcode the value here.

missytake (commented):
I think we can hardcode this variable, as it is only used in one place anyway :)

chatmaild/src/chatmaild/config.py
```diff
@@ -43,6 +46,16 @@ def __init__(self, inipath, params):
...
         self.disable_ipv6 = params.get("disable_ipv6", "false").lower() == "true"
+        self.use_foreign_cert_manager = (
+            params.get("use_foreign_cert_manager", "false").lower() == "true"
+        )
+        self.acme_email = params["acme_email"]
+        self.change_kernel_settings = (
+            params.get("change_kernel_settings", "true").lower() == "true"
+        )
+        self.fs_inotify_max_user_instances_and_watchers = int(
+            params["fs_inotify_max_user_instances_and_watchers"]
```
link2xt: This looks like it will fail if the setting is not in the config file.

docker/example.env
Keonik1: It can be used not only for installation with Traefik, but also for any third-party certificate manager. Traefik is simply more convenient in this regard and is a good example.

I think you can write a comment directly in the example.env file stating that if you are using the standard installation, you can leave this field untouched, as it will not have any effect.

Or we can remove it altogether if we want a more linear installation.

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -25,24 +25,37 @@
...
+    inipath = args.inipath
     if args.inipath.exists():
-        print(f"Path exists, not modifying: {args.inipath}")
-        return 1
-    else:
-        write_initial_config(args.inipath, mail_domain, overrides={})
-        out.green(f"created config file for {mail_domain} in {args.inipath}")
+        if not args.recreate_ini:
+            out.green(f"[WARNING] Path exists, not modifying: {inipath}")
+            return 1 ### need research - can we set return code as zero?
```
link2xt: I think we can because nobody uses it, but why? Running `init` when it does nothing is likely not what you want.

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -63,16 +76,23 @@ def run_cmd_options(parser):
         dest="ssh_host",
         help="specify an SSH host to deploy to; uses mail_domain from chatmail.ini by default",
     )
+    parser.add_argument(
+        "--skip-dns-check",
+        dest="dns_check_disabled",
+        action="store_true",
+        help="disable checks nslookup for dns",
```
link2xt:
```suggestion
        help="disable nslookup checks for DNS",
```

docker/files/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -eo pipefail
+
+unlink /etc/nginx/sites-enabled/default || true
+
+if [ "${USE_FOREIGN_CERT_MANAGER,,}" == "true" ]; then
+    if [ ! -f "$PATH_TO_SSL/fullchain" ]; then
+        echo "Error: file '$PATH_TO_SSL/fullchain' does not exist. Exiting..." > /dev/stderr
+        sleep 2
```
link2xt, line 9: What is the reason for `sleep` here? Is it a workaround for something?

docker/files/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -eo pipefail
+
+unlink /etc/nginx/sites-enabled/default || true
+
+if [ "${USE_FOREIGN_CERT_MANAGER,,}" == "true" ]; then
```
link2xt:
```suggestion
if [ "${USE_FOREIGN_CERT_MANAGER,,}" = true ]; then
```

`=` is standard, `==` is a bash extension for pattern matching that is not needed here.

docker/files/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+        exit 1
+    fi
+    if [ ! -f "$PATH_TO_SSL/privkey" ]; then
+        echo "Error: file '$PATH_TO_SSL/privkey' does not exist. Exiting..." > /dev/stderr
+        sleep 2
+        exit 1
+    fi
+fi
+
+SETUP_CHATMAIL_SERVICE_PATH="${SETUP_CHATMAIL_SERVICE_PATH:-/lib/systemd/system/setup_chatmail.service}"
```
link2xt, line 19:
```suggestion
: "${SETUP_CHATMAIL_SERVICE_PATH:=/lib/systemd/system/setup_chatmail.service}"
```

Same result, just shorter.

docker/files/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+
+env_vars=$(printenv | cut -d= -f1 | xargs)
+sed -i "s|<envs_list>|$env_vars|g" $SETUP_CHATMAIL_SERVICE_PATH
```
link2xt, line 25: Do we actually need to pass all the environment variables? This is not easy to understand without reading the `setup_chatmail.service` template. At least needs a comment to explain what happens here.

docker/files/entrypoint.sh
Keonik1: @link2xt because it will restart ~5 containers per second, but we need 2-10 seconds to obtain certificates

docker/files/entrypoint.sh
Keonik1: @link2xt 
systemd starts services without passing environment variables to them. A running service can only use the environment variables specified in the `some.service` file.

cmdeploy/src/cmdeploy/cmdeploy.py
Keonik1: Well, this is probably an architectural issue. I think that calling init again shouldn't cause an error.

@hpk42 @missytake
What do you think about this?

missytake: I agree with @link2xt that if re-creating the chatmail.ini file is not wished for, an existing chatmail.ini is an error; your command will not have the effect you want it to have (creating a default configuration file for a given domain).

Scripts which want to re-create the ini like `setup_chatmail_docker.sh` can simply pass `--force` or `|| true` here, depending on the desired behavior.

docker/files/entrypoint.sh
missytake: True, but also harder to read for people without an extensive bash background^^ :shrug:

Keonik1: I agree with missytake. The current entry is more understandable to a wider audience.

Keonik1:
@missytake I've added a feature to allow only selected languages to be displayed. Could you please take a look at the implementation? From my tests, it works fine, and this approach adds flexibility in who wants what languages in their instance.

I hope to test and upload the DNS related changes this weekend.

cmdeploy/src/cmdeploy/www.py
```diff
@@ -21,13 +28,51 @@ def snapshot_dir_stats(somedir):
...
+        lang_file = lang_folder / f"{base_name}.md"
+        lang_name = LANGUAGE_NAMES.get(lang_code, lang_code)
+
+        if lang_file.exists():
+            content = lang_file.read_text().strip()
+        else:
+            print(f"[WARNING]: Missing file {lang_file}. Inserting fallback message.")
+            content = "Content for this language is not available, please contact your server administrator."
+
+        markdown_blocks.append(f"/// tab | {lang_name}\n{content}\n///")
```
missytake: If there is only one language, can we hide the language chooser at the top?

cmdeploy/src/cmdeploy/www.py
```diff
@@ -64,39 +109,43 @@ def int_to_english(number):
...
-            render_vars["username_min_length"] = int_to_english(
-                config.username_min_length
-            )
-            render_vars["username_max_length"] = int_to_english(
-                config.username_max_length
-            )
-            render_vars["password_min_length"] = int_to_english(
-                config.password_min_length
-            )
+        if path.suffix == ".md" and '.' not in path.stem:
```
missytake: what do we check the stem for?

docker/docker-compose-default.yaml
```diff
@@ -19,7 +19,7 @@ services:
       options:
         max-size: "10m"
         max-file: "3"
-    environment:
+    environment: #all possible variables you can check inside README and /chatmaild/src/chatmaild/ini/chatmail.ini.f
```
missytake:
```suggestion
    environment:  # all possible variables you can check inside README and /chatmaild/src/chatmaild/ini/chatmail.ini.f
```

call me pedantic ;)

docker/docker-compose-default.yaml
```diff
@@ -31,6 +31,7 @@ services:
       # ENABLE_CERTS_MONITORING: "true"
       # CERTS_MONITORING_TIMEOUT: 10
       # IS_DEVELOPMENT_INSTANCE: "True"
+      # LANGUAGES: "EN" # all possible options you can check inside /chatmaild/src/chatmaild/ini/chatmail.ini.f
```
missytake, line 34: Setting this in .env didn't work for me

Keonik1: It wasn't supposed to work. Variables defined in the `.env` file can be used in the `docker-compose.yaml` file itself, but they are not passed to the container, only if you explicitly specify them using substitution:
```yaml
services:
  chatmail:
    environment:
      LANGUAGES: ${LANGUAGES}
```
Or special keyword:
```yaml
services:
  chatmail:
    env_file:
      - ./.env
```
I also clarified this point with dockerDocsAi on the official docker documentation site, and he confirmed it.

cmdeploy/src/cmdeploy/www.py
Keonik1: My bad. It was left after experimenting with language-specific file structure, it's not relevant now, I just forgot to delete this piece :D

docker/docker-compose-default.yaml
missytake: Hm, does anything speak against passing the whole env file to the container? I kinda expected it to work, and configuring things in an .env file is kinda best practice for docker-compose setups.

Then again, most projects just offer one static docker-compose.yaml which you don't really edit - we *could* leave out the `cp docker-compose-default.yaml docker-compose.yaml` step, *offer* to do `cp docker-compose-traefik.yaml docker-compose.yaml`, and just make everything important configurable in `.env`. This would make it even more a "default" kind of thing, and a streamlined setup.

Keonik1: Configuring settings in the `.env` file is a best practice with docker compose, but this is not the case here.

Most projects work in a way where you download only `docker-compose.yaml` and create a `.env` file next to it, and you do NOT clone the whole repository.

We can't do that, at least at this stage, because we have to clone the whole repository and mount the files inside the container.

With this approach, if we have one docker-compose that is not in the `.gitignore` file and the end user changes something in it, that's it, there will be immediate conflicts when trying to do a `git pull`.

Plus installing via docker-compose via traefik or not is very different. Here you can't do like in kubernetes (helm) and regulate the whole configuration with one `values.yaml`.

All I can suggest at this stage is to create a separate file, for example `chatmail.env`, store in it settings similar to those in `chatmail.ini` in the form of commented variables and mount the file with the `env_file` keyword. However, this file will also be in the `.gitignore`, so as not to cause problems when updating the repository. (This file will be in addition to the existing `.env` file)

We can make an issue for the future, but at the current stage I don't think this is something that needs to be resolved until we remove systemd from the container and as a consequence we can prepare a docker image with chatmail already installed inside.

Gurkengewuerz:
are there any updates on this? i would love to self host chatmailer but i only have docker hosts. this would be great!

one addition: a github release action for github images would be awesome too!

Keonik1:
@Gurkengewuerz Yes, work is in progress. Slower than I would like :( but still in progress.

Keonik1:
@Gurkengewuerz 
It seems to work, but possibly in a limited state (without the option to send messages through different servers, but within one server, everything works).

You can clone the repository [from here](https://github.com/Keonik1/relay/tree/main) and try installing it in Docker right now.

missytake:
To make this PR easier to review, I suggest breaking it up into smaller chunks. I started with https://github.com/chatmail/relay/pull/656, and will try to rebase a copy of this branch on top of it, to see whether it makes it easier to review.

Other features which should be pulled out of this branch into their own PRs:

- markdown language switcher: https://github.com/chatmail/relay/pull/658
- run SSH commands locally: https://github.com/chatmail/relay/pull/659
- Add `--skip-dns-check` argument to cmdeploy run command: https://github.com/chatmail/relay/pull/661
- Add startup for `fcgiwrap.service`: https://github.com/chatmail/relay/pull/657
- Local SSH execution for `docker-dns` and `docker-tests` (see also #629) : https://github.com/chatmail/relay/pull/660
- acme_email config value: https://github.com/chatmail/relay/pull/663
- traefik support: https://github.com/chatmail/relay/pull/662
- Add extended check when installing `unbound.service`? this is unfortunately a bit hard to reproduce

missytake:
> To make this PR easier to review, I suggest breaking it up into smaller chunks. I started with #656, and will try to rebase a copy of this branch on top of it, to see whether it makes it easier to review.

Okay, I'm done with the splitting up, and there are now 8 smaller PRs which we can review :)))

I suggest to close this in favor of https://github.com/chatmail/relay/pull/664, @Keonik1 what do you think?

Keonik1:
> > To make this PR easier to review, I suggest breaking it up into smaller chunks. I started with #656, and will try to rebase a copy of this branch on top of it, to see whether it makes it easier to review.
> 
> Okay, I'm done with the splitting up, and there are now 8 smaller PRs which we can review :)))
> 
> I suggest to close this in favor of #664, @Keonik1 what do you think?

Yes, I think we can close this

paulo-roger:
Hi! What's the status update about this? I am really interesting in oferring a chatmail relay, but I must use docker. I was gonna create a DOCKERFILE myself but an official method is quite better.

one question: could it be used alpine for the container?

missytake:
> Hi! What's the status update about this? I am really interesting in oferring a chatmail relay, but I must use docker. I was gonna create a DOCKERFILE myself but an official method is quite better.

This PR now lives at https://github.com/chatmail/relay/pull/664 :) it's gonna get rebased + reviewed soon, but it should work as it is. Maybe don't use it for a production deployment yet.

> one question: could it be used alpine for the container?

No, right now cmdeploy requires apt as package manager, and systemd.

paulo-roger:
Thank you very much!! @missytake

PR #615, opened 2025-08-11
docs: Add ARCHITECTURE.md with diagram of components

cliffmccarthy:
- For starters, this file is just a diagram of components of a chatmail server.  In the future, this document can grow into a more complete description of the architecture of the server, the deployment process, and the design intent behind what is and isn't in the code base.
- The name ARCHITECTURE.md is inspired by this article, which also has good suggestions about what to put in the file: https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,4 +1,5 @@
-This diagram shows components of the chatmail server:
+This diagram shows components of the chatmail server; this is a draft
+overview as of mid-August 2025:
 
 ```mermaid
 graph LR;
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -3,36 +3,48 @@ overview as of mid-August 2025:
 
 ```mermaid
 graph LR;
-    cmdeploy --> sshd;
-    cron --> expunge;
-    cron --> acmetool;
-    cron --> chatmail-metrics;
-    chatmail-metrics --> /var/www/html;
-    acmetool --> certs;
-    acmetool --> acmetool-redirector;
-    acmetool-redirector --> certs;
-    nginx --> /var/www/html;
-    nginx --> certs;
-    nginx --> newemail.py;
-    nginx --> |465|postfix;
-    nginx --> autoconfig.xml;
-    nginx --> |993|dovecot;
-    autoconfig.xml --> postfix;
-    autoconfig.xml --> dovecot;
-    postfix --> certs;
-    postfix --> /home/vmail/mail;
-    postfix --> |10080,10081|filtermail;
-    postfix --> echobot;
-    postfix --> |doveauth.socket|doveauth;
-    dovecot --> certs;
-    dovecot --> |doveauth.socket|doveauth;
-    dovecot --> /home/vmail/mail;
-    dovecot --> |metadata.socket|chatmail-metadata;
-    doveauth --> /home/vmail/mail;
-    expunge --> /home/vmail/mail;
-    chatmail-metadata --> iroh-relay;
+    cmdeploy --- sshd;
+    letsencrypt --- |80|acmetool-redirector;
+    acmetool-redirector --- |443|nginx-right(["`nginx
+    (external)`"]);
+    nginx-external --- |465|postfix;
+    nginx-external(["`nginx
+    (external)`"]) --- |8443|nginx-internal["`nginx
+    (internal)`"];
+    nginx-internal --- website["`Website
+    /var/www/html`"];
+    nginx-internal --- newemail.py;
+    nginx-internal --- autoconfig.xml;
+    certs-nginx[("`TLS certs
+    /var/lib/acme`")] --> nginx-internal;
+    cron --- chatmail-metrics;
+    cron --- acmetool;
+    cron --- expunge;
+    chatmail-metrics --- website;
+    acmetool --> certs[("`TLS certs
+    /var/lib/acme`")];
+    nginx-external --- |993|dovecot;
+    autoconfig.xml --- postfix;
+    autoconfig.xml --- dovecot;
+    postfix --- echobot;
+    postfix --- |10080,10081|filtermail;
+    postfix --- users["`User data
+    home/vmail/mail`"];
+    postfix --- |doveauth.socket|doveauth;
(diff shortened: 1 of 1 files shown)


cliffmccarthy:
The current revision is a rough draft, and will probably be revised further as I study the code more.

ARCHITECTURE.md
```diff
@@ -0,0 +1,37 @@
+This diagram shows components of the chatmail server:
```
hpk42: Please qualify this sentence that's it's a draft overview as of mid August 2025.

cliffmccarthy: Thanks for the review!  I updated the text as you suggested.  I also added labels to some of the arrows to indicate the ports and sockets that the components communicate by.

ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+```mermaid
+graph LR;
+    cmdeploy --> sshd;
+    cron --> expunge;
+    cron --> acmetool;
+    cron --> chatmail-metrics;
+    chatmail-metrics --> /var/www/html;
+    acmetool --> certs;
+    acmetool --> acmetool-redirector;
+    acmetool-redirector --> certs;
```
missytake:
```suggestion
    acmetool-redirector --> nginx;
```
I would argue that acmetool-redirector doesn't actually have much to do with certs, and mostly just redirects traffic to nginx?

ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+    nginx --> /var/www/html;
```
missytake:
```suggestion
    nginx --> |8443|/var/www/html;
```
We could include that port 8443 is needed to show the webroot directory :)

ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+    nginx --> certs;
+    nginx --> newemail.py;
```
missytake:
```suggestion
    nginx --> |8443|newemail.py;
```
same for newemail.py :)

ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+    cmdeploy --> sshd;
+    cron --> expunge;
+    cron --> acmetool;
+    cron --> chatmail-metrics;
+    chatmail-metrics --> /var/www/html;
+    acmetool --> certs;
+    acmetool --> acmetool-redirector;
+    acmetool-redirector --> certs;
+    nginx --> /var/www/html;
+    nginx --> certs;
```
missytake:
```suggestion
    certs --> nginx;
```
And one could argue that nginx is dependent on the certs, not the other way round.

ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+    nginx --> newemail.py;
+    nginx --> |465|postfix;
+    nginx --> autoconfig.xml;
+    nginx --> |993|dovecot;
+    autoconfig.xml --> postfix;
+    autoconfig.xml --> dovecot;
+    postfix --> certs;
```
missytake:
```suggestion
    certs --> postfix;
```
same for postfix...

ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+    postfix --> /home/vmail/mail;
+    postfix --> |10080,10081|filtermail;
+    postfix --> echobot;
+    postfix --> |doveauth.socket|doveauth;
+    dovecot --> certs;
```
missytake:
```suggestion
    certs --> dovecot;
```
...and dovecot ;)

ARCHITECTURE.md
```diff
@@ -0,0 +1,38 @@
...
+overview as of mid-August 2025:
+
+```mermaid
+graph LR;
+    cmdeploy --> sshd;
+    cron --> expunge;
+    cron --> acmetool;
+    cron --> chatmail-metrics;
+    chatmail-metrics --> /var/www/html;
+    acmetool --> certs;
```
missytake:
```suggestion
    acmetool --> /var/lib/acme;
```

For `/home/vmail` and `/var/www/html` we also talk about the path, because they contain state... we could do that for the certs as well :) that would make it clearer what to backup/manage. maybe `TLS certs /var/lib/acme`, `Users /home/vmail`, and `Website /var/www/html`? That would make it more obvious what the paths contain.

missytake (approved):
This is a great start, the only missing things I can think of are opendkim and the lastlogin service, but we can also merge it as it is and add them later :)

I added some suggestions, which would make more sense for me - but as this diagram is not so I can understand it, but for newcomers, I'd leave it to you which you want to adapt :)

missytake:
Let me know what changes you want to include, then I'll merge this :)

cliffmccarthy:
Thanks for all the review feedback!  I will submit a new revision in the next day or two.

cliffmccarthy:
I tried to incorporate all your suggestions in the latest revision.  This isn't perfect, but it's starting to push the limits of what can be reasonably expressed in a mermaid diagram.  I'm open to any suggestions about better colours for the duplicated nodes.  I do want to incorporate opendkim and lastlogin, but that will probably jumble the layout again, leading to another round of graph-hacking.  I'll take that up in another revision (possibly in this PR if it's still open when I get to it).

Here are the details from the latest commit message:

- Implemented changes suggested in review by missytake:
    - Removed relation between acmetool-redirector and certs.
    - Added internal nginx listening on port 8443.
    - Changed direction of arrows between certs and the services that use them.  This makes the arrow show the direction of information flow, rather than a "depends on" relation.
    - For filesystem paths, added a descriptive name to the node.
- Replaced most arrows with plain lines, to simply show that a relationship exists between the two nodes.  This also reduces visual clutter, since the graph is pretty dense with information already.
- Split nginx and certs into two nodes, to reduce entanglement in the graph.  These "linked" nodes are given a different shape and filled with a different colour, to highlight the fact that they are the same node.
- Revised text about the meaning of edges in the graph.

missytake:
Thanks, let's merge this for now :) I also have some ideas, which go more into the direction of "depending on" than "information flow" actually. Might open a PR about it :)

PR #617, opened 2025-08-19
nginx: allow custom nginx config under sites-enabled

### LINKED ISSUE #562: ChatMail destroys existing nginx site configurations

missytake:
fix #562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- Allow custom nginx config files
+  ([#617](https://github.com/chatmail/relay/pull/617))
+
 - Check whether GCC is installed in initenv.sh
   ([#608](https://github.com/chatmail/relay/pull/608))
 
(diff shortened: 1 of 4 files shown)


link2xt:
I don't like the idea of expanding the scope with the ability to host arbitrary static websites on the same machine, because then you may want to host non-static websites, it may interfere with certificate renewal if something fails etc. I also don't see where this is requested in the linked issue #562.

If someone wants to host other websites next to chatmail, then chatmail needs to be virtualized and a proxy in front of it can dispatch traffic to chatmail or to another website.

missytake:
True, maybe #618 is the better path forward here. People who want to host non-static website can be referred to this PR for manual tinkering, but supporting this use case is probably a bad idea.

PR #618, opened 2025-08-19
www: introduce www_folder config item

### LINKED ISSUE #529: (Option to) avoid building/deploying webpages
Hello,

as a chatmail instance, we would like to ship our own www version of the homepage, so it would be cool if the build/deploy process for default pages could be (optionally) skipped at cmdeploy run.

Thanks :)

missytake:
- [x] Some kind of test would be nice, as the upload logic is kinda intricate. Maybe put it into a shorter function which can be tested.

fix #529

~~based on https://github.com/chatmail/relay/pull/617, mostly because they share a README section. But better review/merge that one before.~~


missytake:
I rebased this PR on main, because #617 is controversial - so now we can review and merge this independently :)

PR #621, opened 2025-08-21
feat: setup TURN server

### LINKED ISSUE #619: add STUN/TURN server for calls
the  STUN/TURN server needs to be picked up by https://github.com/deltachat/calls-webapp then. maybe also other parts need adaption

@adbenitez @link2xt @missytake @all please add more details to this issue or create new ones, i have no good idea what is needed

link2xt:
Closes #619 
Based on #650 

This disables STUN server in iroh-relay and deploys another `coturn` as a STUN/TURN server.

I have tested that online tests using realtime channels still pass against a server that has this deployed. Also tested that "pong" app works between two profiles using the same relay but connected using different Wi-Fi networks, so at least realtime channels are not broken by this.

Corresponding core PR: https://github.com/chatmail/core/pull/7218

### DIFF
--- a/chatmaild/src/chatmaild/metadata.py
+++ b/chatmaild/src/chatmaild/metadata.py
@@ -7,6 +7,7 @@ from .config import read_config
 from .dictproxy import DictProxy
 from .filedict import FileDict
 from .notifier import Notifier
+from .turnserver import turn_credentials
 
 
 def _is_valid_token_timestamp(timestamp, now):
@@ -75,11 +76,12 @@ class Metadata:
 
 
 class MetadataDictProxy(DictProxy):
-    def __init__(self, notifier, metadata, iroh_relay=None):
+    def __init__(self, notifier, metadata, iroh_relay=None, turn_hostname=None):
         super().__init__()
         self.notifier = notifier
         self.metadata = metadata
         self.iroh_relay = iroh_relay
+        self.turn_hostname = turn_hostname
 
     def handle_lookup(self, parts):
         # Lpriv/43f5f508a7ea0366dff30200c15250e3/devicetoken\tlkj123poi@c2.testrun.org
@@ -98,6 +100,11 @@ class MetadataDictProxy(DictProxy):
             ):
                 # Handle `GETMETADATA "" /shared/vendor/deltachat/irohrelay`
                 return f"O{self.iroh_relay}\n"
+            elif keyname == "vendor/vendor.dovecot/pvt/server/vendor/deltachat/turn":
+                res = turn_credentials()
+                port = 3478
+                return f"O{self.turn_hostname}:{port}:{res}\n"
+
         logging.warning(f"lookup ignored: {parts!r}")
         return "N\n"
 
@@ -121,6 +128,7 @@ def main():
 
     config = read_config(config_path)
     iroh_relay = config.iroh_relay
+    mail_domain = config.mail_domain
 
     vmail_dir = config.mailboxes_dir
     if not vmail_dir.exists():
@@ -134,7 +142,10 @@ def main():
     notifier.start_notification_threads(metadata.remove_token_from_addr)
 
     dictproxy = MetadataDictProxy(
-        notifier=notifier, metadata=metadata, iroh_relay=iroh_relay
+        notifier=notifier,
+        metadata=metadata,
+        iroh_relay=iroh_relay,
+        turn_hostname=mail_domain,
     )
 
     dictproxy.serve_forever_from_socket(socket)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- Setup TURN server
+  ([#621](https://github.com/chatmail/relay/pull/621))
+
 - Update iroh-relay to 0.35.0
   ([#650](https://github.com/chatmail/relay/pull/650))
 
--- a/chatmaild/pyproject.toml
+++ b/chatmaild/pyproject.toml
@@ -29,6 +29,7 @@ echobot = "chatmaild.echo:main"
 chatmail-metrics = "chatmaild.metrics:main"
 delete_inactive_users = "chatmaild.delete_inactive_users:main"
 lastlogin = "chatmaild.lastlogin:main"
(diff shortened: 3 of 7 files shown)


link2xt:
Version of coturn from Debian 13 (4.6.1-2+b2) works in both Chromium and Firefox.
Version of coturn from Debian 12 (4.6.1-1) does not work in Firefox when I test with https://icetest.info/
Might be this issue: https://github.com/coturn/coturn/issues/1438

Maybe we should try https://github.com/mycrl/turn-rs instead and have a static binary that is the same everywhere. Or we can install newer .deb of coturn directly.

fluoriteByte:
> Or we can install newer .deb of coturn directly.

as someone who has experimented with mixing packages that expect different versions of debian, this would a problem down the line, especially since coturn depends on 16 different dependencies, any mismatch and apt will eventually just destroy the system lol

link2xt:
Now there is a metadata command to get temporary TURN credentials like this:
```
? GETMETADATA "" (/private/turn)
* METADATA "" (/private/turn {39}
1758650868:8Dqkyyu11MVESBqjbIylmB06rv8=)
? OK Getmetadata completed (0.001 + 0.000 + 0.001 secs).
```

1758650868 is the login and expiration timestamp at the same time, 8Dqkyyu11MVESBqjbIylmB06rv8= is the password.

link2xt:
I tested turn-rs, but it is not easy to configure automatically as you need to put the IP address into the configuration file: https://github.com/mycrl/turn-rs/issues/144

link2xt:
I have added on CI server a file `/etc/apt/apt.conf.d/99defaultrelease` with `APT::Default-Release "bookworm";`.
Then added trixie to sources list:
```
deb http://deb.debian.org/debian trixie main contrib non-free-firmware
deb http://deb.debian.org/debian trixie-updates main contrib non-free-firmware
deb http://security.debian.org/debian-security trixie-security main contrib non-free-firmware
```

Trying to install version from `trixie` failed, `apt install coturn/trixie` reported:
```
The following packages have unmet dependencies:
 libc-bin : Depends: libc6 (< 2.37) but 2.41-12 is to be installed
 libc6 : Breaks: base-files (< 13.3~) but 12.4+deb12u12 is to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
```

So installing a version from `stable` on `oldstable` does not work.

link2xt:
Rust turn server works with this in `/etc/systemd/system/turn-server.service`:
```[Unit]
Description=A pure rust-implemented turn server.
After=network.target

[Service]
Type=simple
Restart=always
ExecStart=/usr/local/bin/turn-server --config=/etc/turn-server/config.toml

[Install]
WantedBy=multi-user.target
```

And this config:
```toml
[turn]
realm = "ci-chatmail.testrun.org"

[[turn.interfaces]]
transport = "udp"
bind = "49.12.217.82:3478"
external = "49.12.217.82:3478"

[[turn.interfaces]]
transport = "tcp"
bind = "49.12.217.82:3478"
external = "49.12.217.82:3478"

[[turn.interfaces]]
transport = "udp"
bind = "[2a01:4f8:c013:13c3::1]:3478"
external = "[2a01:4f8:c013:13c3::1]:3478"

[[turn.interfaces]]
transport = "tcp"
bind = "[2a01:4f8:c013:13c3::1]:3478"
external = "[2a01:4f8:c013:13c3::1]:3478"

[log]
level = "info"

[auth]
static_auth_secret = "north"

[auth.static_credentials]
ohV8aec1 = "zo3theiY"
```

We have to somehow put IP addresses into the config because of https://github.com/mycrl/turn-rs/issues/144
Binding to `[::]` should be enough to bind on IPv6 and IPv4 at the same time, but then we cannot specify external IPv4 address and external IPv6 address.

Both static login/password and so-called REST authentication works.

REST has nothing to do with REST, it's called this for historical reasons, actually it is a password generation from a timestamp scheme you can see implemented in Python in this PR as well with a secret hardcoded to "north" as in coturn example.

link2xt:
Maybe Rust TURN server does not actually work as a TURN server, @WofWca says attempts to connect to allocated endpoints return ICMP Port Unreachable and I also don't see the ports allocated on the server: https://github.com/mycrl/turn-rs/issues/136#issuecomment-3340407041

I am looking at SDP offer generated by Delta Chat, this is what it sends in relay-only mode:
```
a=candidate:2143093290 1 udp 33562368 49.12.217.82 59968 typ relay raddr 0.0.0.0 rport 0 generation 0 network-cost 999
a=candidate:3079513688 1 udp 33565440 2a01:4f8:c013:13c3::1 60928 typ relay raddr :: rport 0 generation 0 network-cost 999
```

TURN server is running on port 3478 so port 59968 is not allocated and connecting to it fails.

With coturn (turn.jami.net) relay candidate looks like this:
```
a=candidate:1538841534 1 udp 33562623 51.254.39.157 14224 typ relay raddr ::ffff:x.x.x.x rport 45565 generation 0 network-cost 999
```

x.x.x.x is apparently my ISP IP address.

It would probably be fine to not allocate an address for real on the TURN server and distinguish the clients by virtual `rport`, but the candidate should have port 3478 and not port 59968. Besides, turn-rs returns 0 as raddr/rport.

I opened an issue at https://github.com/mycrl/turn-rs/issues/145

chatmaild/src/chatmaild/metadata.py
```diff
@@ -98,6 +100,11 @@ def handle_lookup(self, parts):
             ):
                 # Handle `GETMETADATA "" /shared/vendor/deltachat/irohrelay`
                 return f"O{self.iroh_relay}\n"
+            elif keyname == "vendor/vendor.dovecot/pvt/server/vendor/deltachat/turn":
+                res = turn_credentials()
```
WofWca, line 104: Since the credentials are temporary, is it guaranteed that DC clients will re-get this metadata within 5 days?

https://github.com/chatmail/chatmail-turn/blob/7f7cc0c0f24108dee6acc62e00694512bd3fe399/src/main.rs#L35

link2xt: Core refreshes credentials if they are about to expire in 12 hours.

Issue #623, opened 2025-08-29
Error during update, port already in use


pierinh00d:
Hi. I wanted to update my chatmail installation. I ran the usual:

`$ git pull --rebase --autostash`

and then 

`$ ./scripts/cmdeploy run`

from my relay source directory, but:

```
$ ./scripts/cmdeploy run
[ssh] login to chatmail.culturanerd.it
Collecting initial DNS settings..............
[$ pyinfra --ssh-user root chatmail.culturanerd.it /home/******/relay/cmdeploy/src/cmdeploy/deploy.py -y]
--> Loading config...
--> Loading inventory...
--> Connecting to hosts...
    [chatmail.culturanerd.it] Connected

--> Preparing operation files...
    Loading: /home/******/relay/cmdeploy/src/cmdeploy/deploy.py
Deploy failed: port 25 is occupied by: smtpd

--> Disconnecting from hosts...
Deploy failed

```

It is obvious that the ports are busy on the destination server. But why is this happening? Why is this check being performed? Is this something new?

```
$ git rev-parse HEAD
73831c74d91eed010266fd0a630fa9ba2fe11ee7
```

Thank you a  lot

missytake:
> It is obvious that the ports are busy on the destination server. But why is this happening? Why is this check being performed? Is this something new?

It's something new - the goal is to check whether the ports chatmail relay needs are available for us.

When I implemented this, the postfix process on port 25 was called `master`, not it's called `smtpd` apparently. There is a fix now, you can try to `git pull` again.

PR #624, opened 2025-08-29
cmdeploy: on Ubuntu, postfix calls its port 25 process 'smtpd'

missytake:
It seems on some systems the process is not called `master`, but `smtpd`. Maybe it was introduced by https://github.com/chatmail/relay/pull/590 ?

Let's see.

fix https://github.com/chatmail/relay/issues/623


cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -676,7 +676,7 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
     from cmdeploy.cmdeploy import Out
 
     port_services = [
-        ("master", 25),
+        (["master", "smtpd"], 25),
```
ralphtheninja, line 679: This line should be repeated on all other "master" lines below.

link2xt:
I don't think it is related to Ubuntu. The port is normally listened on by `master`, but on actively used system receiving message the port is used both by `master` and forked `smtpd` because `master` does not close the listening socket when forking to `smtpd`. `smtpd` probably does not accept new connections, but it still has a listening port 25 socket in its process.

ralphtheninja:
> I don't think it is related to Ubuntu. The port is normally listened on by `master`, but on actively used system receiving message the port is used both by `master` and forked `smtpd` because `master` does not close the listening socket when forking to `smtpd`. `smtpd` probably does not accept new connections, but it still has a listening port 25 socket in its process.

We know it's not related to Ubuntu since the person reporting the issue was using Debian 12. Just a misunderstanding in the title.

PR #628, opened 2025-09-11
WIP: cmdeploy command to deploy a reverse proxy

missytake:


hpk42:
this PR is pretty stale, and against an older main.
I agree it'd be nice to automate setup of proxies but they are quite different from a chatmail relay.  I'd suggest to use pyinfra separately from the "cmdeploy run" machinery, and not treat proxies as a special kind of chatmail relay. (maybe i misundestood the PR, and it's already doing this?)

PR #632, opened 2025-09-11
dovecot: keep mailbox index only in memory

missytake:
According to research by @hpk42 and @feld, this could save a lot of storage. login seems to take 0.34 seconds longer if the index is not cached on disk; not a big deal.

Deleting all the index files is not part of this PR, can be done manually once for servers which want to reduce storage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## untagged
 
-- dovecot: keep mailbox index only in memory
+- dovecot: keep mailbox index only in memory to avoid unnecessary disc usage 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## untagged
 
+- dovecot: keep mailbox index only in memory
```
hpk42:
```suggestion
- dovecot: keep mailbox index only in memory to avoid unnecessary disc usage 
```

link2xt:
Looks like this at least broke Seen flag synchronization for multi-device: #640

PR #636, opened 2025-09-14
nginx: be more specific with the server name

missytake:
Request by dorfzwockel, who runs other services on the same server


hpk42 (approved):
i'd still add a changelog entry because it might be relevant to operators.

PR #637, opened 2025-09-14
Replace expiry "find" commands with a new chatmaild.expire python module + a reporting one

hpk42:
Before this PR, in https://github.com/chatmail/relay/blob/1.7.0/cmdeploy/src/cmdeploy/dovecot/expunge.cron.j2 we execute a number of "find" commands, which at least on nine.testrun.org take overall 45+ minutes to complete, and access the full directory structure 9 times (kernel caching helps a little).  The new `chatmaild.expiry` traverses all mailboxes only once and has tests.  

On the server side, you can run "python -m chatmaild.expire /usr/local/lib/chatmail.ini" to see what it would remove.  Only if you specify "--remove" will it actually remove it.  

The PR adds also "python -m chatmaild.fsreport" which gives various statistics of file usages, last login times etc.  This is not run in any cron job and purely for curiosity purposes (and to help determine what we might do if storage becomes full again).

### DIFF
--- a/chatmaild/src/chatmaild/fsreport.py
+++ b/chatmaild/src/chatmaild/fsreport.py
@@ -81,7 +81,10 @@ class Report:
         print("## Mailbox storage use analysis")
         print(f"Mailbox data total size: {HSize(self.size_extra + all_messages)}")
         print(f"Messages total size    : {HSize(all_messages)}")
-        percent = self.size_extra / (self.size_extra + all_messages) * 100
+        try:
+            percent = self.size_extra / (self.size_extra + all_messages) * 100
+        except ZeroDivisionError:
+            percent = 100
         print(f"Extra files : {HSize(self.size_extra)} ({percent:.2f}%)")
 
         print()
@@ -98,7 +101,7 @@ class Report:
         user_logins = self.num_all_logins - self.num_ci_logins
 
         def p(num):
-            return f"({num / user_logins * 100:2.2f}%)"
+            return f"({num / user_logins * 100:2.2f}%)" if user_logins else "100%"
 
         print()
         print(f"## Login stats, from date reference {datetime.fromtimestamp(self.now)}")
--- a/chatmaild/src/chatmaild/tests/test_expire.py
+++ b/chatmaild/src/chatmaild/tests/test_expire.py
@@ -79,6 +79,11 @@ def test_stats_mailbox(mbox1):
     assert mbox3.last_login is None
 
 
+def test_report_no_mailboxes(example_config):
+    args = (str(example_config._inipath),)
+    report_main(args)
+
+
 def test_report(mbox1, example_config):
     args = (str(example_config._inipath),)
     report_main(args)


chatmaild/pyproject.toml
```diff
@@ -27,7 +27,8 @@ chatmail-metadata = "chatmaild.metadata:main"
 filtermail = "chatmaild.filtermail:main"
 echobot = "chatmaild.echo:main"
 chatmail-metrics = "chatmaild.metrics:main"
-delete_inactive_users = "chatmaild.delete_inactive_users:main"
+expire = "chatmaild.expire:main"
```
adbenitez: meta: would be nicer if all chatmail-added commands started with `chatmail-` like the `chatmail-metrics` command, this makes it easier to discover by typing chatmail+TAB and get autocompletion of available chatmail commands while avoiding conflicts with other tools with same name ex "expire" seems quite generic

hpk42: i agree about some common prefix, maybe "cm-" is enough. not sure this PR here should change it though.

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+        print(f"non-ci:  {K(user_logins)}")
+        print(f"ci:      {K(self.num_ci_logins)}")
+        for days, active in self.login_buckets.items():
+            print(f"last {days:3} days: {K(active)} {p(active)}")
+
+
+def main(args=None):
+    """Report about filesystem storage usage of all mailboxes and messages"""
+    parser = ArgumentParser(description=main.__doc__)
+    # parser.add_argument(
```
link2xt: Commented out code.

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+class Report:
+    def __init__(self, now, min_login_age, mdir):
+        self.size_extra = 0
+        self.size_messages = 0
+        self.now = now
+        self.min_login_age = min_login_age
+        self.mdir = mdir
+
+        self.num_ci_logins = self.num_all_logins = 0
+        self.login_buckets = dict((x, 0) for x in (1, 10, 30, 40, 80, 100, 150))
```
link2xt:
```suggestion
        self.login_buckets = {(x, 0) for x in (1, 10, 30, 40, 80, 100, 150)}
```

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    parser.add_argument(
+        "--maxnum",
+        default=None,
+        action="store",
+        help="maximum number of mailbxoes to iterate on",
+    )
+
+    args = parser.parse_args([str(x) for x in args] if args else args)
```
link2xt: Why arguments need to be converted to `str`? `args` is even `None` outside the tests, and seems to be a list of strings already inside the tests.

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
+"""
+command line tool to analyze mailbox message storage
+
+example invocation:
+
+    python -m chatmaild.fsreport /home/vmail/mail/nine.testrun.org
```
link2xt: nine.testrun.org is not an example domain.

chatmaild/src/chatmaild/expire.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def main(args):
+    """Expire mailboxes and messages according to chatmail config"""
+    parser = ArgumentParser(description=main.__doc__)
+    parser.add_argument(
+        "chatmail_ini", action="store", help="path pointing to chatmail.ini file"
+    )
+    parser.add_argument(
+        "mailboxes_dir", action="store", help="path to directory of mailboxes"
```
link2xt: chatmail.ini is a required argument already, mailboxes directory can be determined from there, no need for users to specify it manually.

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+        "--mdir",
+        action="store",
+        help="only consider 'cur' or 'new' or 'tmp' messages for summary",
+    )
+
+    parser.add_argument(
+        "--maxnum",
+        default=None,
+        action="store",
+        help="maximum number of mailbxoes to iterate on",
```
link2xt:
```suggestion
        help="maximum number of mailboxes to iterate on",
```

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+
+from chatmaild.expire import iter_mailboxes
+
+DAYSECONDS = 24 * 60 * 60
+MONTHSECONDS = DAYSECONDS * 30
+
+
+def D(timestamp, now=datetime.utcnow().timestamp()):
+    diff_seconds = int(now) - int(timestamp)
+    # assert diff_seconds >= 0, (int(timestamp), int(now))
```
link2xt: Commented out code.

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+from argparse import ArgumentParser
+from datetime import datetime
+
+from chatmaild.expire import iter_mailboxes
+
+DAYSECONDS = 24 * 60 * 60
+MONTHSECONDS = DAYSECONDS * 30
+
+
+def D(timestamp, now=datetime.utcnow().timestamp()):
```
link2xt: Definitely not a good function name, especially without any documentation or comments.

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+    diff_seconds = int(now) - int(timestamp)
+    # assert diff_seconds >= 0, (int(timestamp), int(now))
+    return f"{int(diff_seconds / DAYSECONDS):2.0f}d"
+
+
+def K(size):
+    if size < 1000:
```
link2xt: `M` always returns megabytes, but `K` somehow sometimes returns bytes? Should this logic go into `H`?

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+        return f"{size:6.0f}"
+    elif size < 10000:
+        return f"{size / 1000:3.2f}K"
+    return f"{int(size / 1000):5.0f}K"
```
link2xt: Why do you format an integer using an `f` (float) formatter? I guess `int` is not needed and the same effect can be achieved with a formatter, but no idea what the expected result is.

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def M(size):
+    return f"{int(size / 1000000):5.0f}M"
+
+
+def H(size):
```
link2xt: I expected this to be used everywhere to format numbers, but now it seems that `K` and `M` are used somewhat arbitrarily directly, likely based on the numbers expected. But then for small servers where the number of users is below 100 it does not make sense to format them with `K` (except that, well, `K` currently has some logic that should have been in `H`, so it sort of works).

chatmaild/src/chatmaild/fsreport.py
```diff
@@ -0,0 +1,172 @@
...
+    def __init__(self, now, min_login_age, mdir):
+        self.size_extra = 0
+        self.size_messages = 0
+        self.now = now
+        self.min_login_age = min_login_age
+        self.mdir = mdir
+
+        self.num_ci_logins = self.num_all_logins = 0
+        self.login_buckets = dict((x, 0) for x in (1, 10, 30, 40, 80, 100, 150))
+        self.message_buckets = dict((x, 0) for x in (0, 160000, 500000, 2000000))
```
link2xt: Where do these numbers come from? Also for `login_buckets`?

chatmaild/src/chatmaild/expire.py
```diff
@@ -0,0 +1,171 @@
...
+    config = read_config(args.chatmail_ini)
+    now = datetime.utcnow().timestamp()
+    if args.days:
+        now = now - 86400 * int(args.days)
+
+    maxnum = int(args.maxnum) if args.maxnum else None
+    exp = Expiry(config, dry=not args.remove, now=now, verbose=args.verbose)
+    for mailbox in iter_mailboxes(os.path.abspath(args.mailboxes_dir), maxnum=maxnum):
+        exp.process_mailbox_stat(mailbox)
+    print(exp.get_summary())
```
link2xt, line 173: If this is meant to with `cron`, it should not print anything. `cron` assumes that any job that printed something failed and sends the message to the user (vmail I guess).

I think the best way to handle it is to turn `expire` into a long-running service so its logs can be inspected with journalctl. This only needs a long-running loop which sleeps for a day in-between.

This way we will also be able to collect metrics by parsing its logs with `mtail` and e.g. display how many accounts or messages expired during the day.

chatmaild/src/chatmaild/expire.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def remove_file(self, path):
+        if self.verbose:
+            print_info(f"removing {path}")
+        if not self.dry:
+            try:
+                os.unlink(path)
+            except FileNotFoundError:
+                print_info(f"file not found/vanished {path}")
+        self.del_files += 1
```
missytake, line 91: It would be good to have tests that this only unlinks files for one user, in case others have a valid reason to keep the file...

Right now we use hardlinks for deduplicating messages which are in more than one mailbox. the mtime will be the same for all of them, so for active mailboxes they should expire at the same time. But if one mboxdir is expiring and removed, this shouldn't delete the messages in all the other mailboxes.

chatmaild/src/chatmaild/expire.py
```diff
@@ -0,0 +1,171 @@
...
+            else:
+                continue
+            changed = True
+        if changed:
+            self.remove_file("maildirsize")
+
+    def get_summary(self):
+        return (
+            f"Removed {self.del_mboxes} out of {self.all_mboxes} mailboxes "
+            f"and {self.del_files} out of {self.all_files} files "
```
missytake: should we be transparent about `self.del_files` not including the messages from deleted mailboxes?

chatmaild/src/chatmaild/expire.py
missytake: It could be a systemd timer unit that runs regularly. No need to handle the loop in the `expire` module. And then it still ends up in one journal.

It would be kinda neat to turn all cron jobs into systemd timers tbh.

chatmaild/src/chatmaild/expire.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+    parser.add_argument(
+        "--days", action="store", help="assume date to be days older than now"
+    )
+
+    parser.add_argument(
+        "--maxnum",
+        default=None,
+        action="store",
+        help="maximum number of mailbxoes to iterate on",
```
missytake:
```suggestion
        help="maximum number of mailboxes to iterate on",
```
small typo

chatmaild/src/chatmaild/expire.py
```diff
@@ -0,0 +1,171 @@
...
+            self.remove_mailbox(mbox.basedir)
+            return
+
+        # all to-be-removed files are relative to the mailbox basedir
+        os.chdir(mbox.basedir)
+        self.all_files += len(mbox.messages)
+        for message in mbox.messages:
+            if message.mtime < cutoff_mails:
+                self.remove_file(message.relpath)
+            elif message.size > 200000 and message.mtime < cutoff_large_mails:
```
missytake, line 112: can we make 200000 a CLI argument for manual tinkering? Or where does this number come from?

missytake (approved):
Great PR, looking forward to deploying this :)

Personally I see many good comments to improve this PR, but no blockers.

chatmaild/pyproject.toml
hpk42: added a prefix 'chatmail-' for now.

chatmaild/src/chatmaild/fsreport.py
hpk42: i sometimes have test code that passes Path objects but removed it here now.

chatmaild/src/chatmaild/fsreport.py
hpk42: message_buckets: 160K is the "download-on-demand" minimum setting.  500K and 2MB are made up to get a sense of distribution of larger than 160K messages.

hpk42: login-buckets are pretty made up ... but "1" and "30" are good for daily/monthly active users. 
if you have any reasoning for different numbers, happy about a commit.

chatmaild/src/chatmaild/expire.py
hpk42: the previous "find" commands also used 200K, so this is rather a port here. 
All limits/numbers could be make configurable but i'd prefer this in a new PR.

chatmaild/src/chatmaild/expire.py
hpk42: not sure i understand correctly.  FWIW I don't think we don't need to test that hardlinks work.

chatmaild/src/chatmaild/expire.py
hpk42: think it'd be ok but i added "in existing mailboxes" to the second line.

chatmaild/src/chatmaild/expire.py
hpk42: i've unified both "chatmail-fsreport" and "chatmail-expire" to take a chatmail.ini file. 
https://github.com/chatmail/relay/commit/aeec17c058e3cb8ae6b11b31aa136e95c833872c

chatmaild/src/chatmaild/fsreport.py
hpk42: unified all size output to use "H" now.

chatmaild/src/chatmaild/fsreport.py
hpk42: removed the int, simplified the K function.

chatmaild/src/chatmaild/fsreport.py
hpk42: it's now largely H that is used, and K is simplified.

chatmaild/src/chatmaild/fsreport.py
hpk42: fixed

chatmaild/src/chatmaild/fsreport.py
adbenitez: yes, I think 1-2 letters names are better avoided, they are also hard to grep etc.
also the lack of typing hints doesn't help

adbenitez (approved):

chatmaild/src/chatmaild/expire.py
hpk42: ok, i moved it to be a systemd timer see here: 
https://github.com/chatmail/relay/pull/637/commits/0e4899ecf9d4f5fd7b992e1103b7de273ac307a9

Now it would be useful if chatmail-expire provides more ouput as it his neatly sent to journalctl now. 
but i think this can come at a later PR.

link2xt:
Some comments behind this "hidden conversations" are unresolved:
image
It's literally typos and commented out code, so likely was just not expanded.

link2xt (approved):
There are still some unresolved comments, otherwise LGTM.

Also ARCHITECTURE.md still has a line saying "cron --- expunge;", can be removed as expunge does not run with cron anymore.

chatmaild/src/chatmaild/fsreport.py
hpk42: removed D, changed K/M to be contained in a single `HSize` formatting function

chatmaild/src/chatmaild/fsreport.py
hpk42: function is removed now

link2xt:
I have force-pushed to main, assuming this was not the intended way to merge this:
image

PR #646, opened 2025-09-24
Increase maxproc for reinjecting ports from 10 to 100

link2xt:
Otherwise under high load filtermail
starts printing "Connection refused" errors to the log.


hpk42 (approved):
makes sense

PR #648, opened 2025-09-28
Use max username length in newemail.py, not min

cliffmccarthy:
- username_min_length and username_max_length are both set to a default value of 9 in the chatmail.ini.f template.  When they have the same value, it doesn't matter which one we use in newemail.py (which handles the /new URL).  However, if they are configured to different values by the admin, then the current implementation using username_min_length chooses from a smaller set of possible usernames.
- Revised create_newemail_dict() in newemail.py to use username_max_length as the length of the random username it offers via the /new URL.  This randomizes within a much larger set of possible usernames.

### DIFF
--- a/chatmaild/src/chatmaild/newemail.py
+++ b/chatmaild/src/chatmaild/newemail.py
@@ -15,7 +15,7 @@ ALPHANUMERIC_PUNCT = string.ascii_letters + string.digits + string.punctuation
 
 
 def create_newemail_dict(config: Config):
-    user = "".join(random.choices(ALPHANUMERIC, k=config.username_min_length))
+    user = "".join(random.choices(ALPHANUMERIC, k=config.username_max_length))
     password = "".join(
         secrets.choice(ALPHANUMERIC_PUNCT)
         for _ in range(config.password_min_length + 3)


missytake:
I think the shorter username is preferred in many cases. Then again, people don't remember & type them anyway :thinking:

cliffmccarthy:
I suppose that raises the question of what the purpose of setting a range between min and max would be.  As I understand it, the `/new` URL is used by Delta Chat to support a smooth automatic registration process, but that a user could manually enter any username and password as long as they meet the requirements in the server's configuration.  So given that, if a user wants a name that is easy to type in, they could manually create one, right?

link2xt:
> user could manually enter any username and password as long as they meet the requirements in the server's configuration. So given that, if a user wants a name that is easy to type in, they could manually create one, right?

This is not a supported usecase. It depends on using plaintext authentication and if we ever support SCRAM (widely used in XMPP) or OPAQUE (same advantages as SCRAM but the password is not revealed even during registration) this would not be possible. If we say that logging in with arbitrary login and password is supported, then we can never get away from plaintext auth.

adbenitez:
I fully agree, better remove the two settings and just have one, a range is not necessary, and people shouldn't login manually trying to create memorable addresses anyways, since the addresses are not important but the key is the identify

cliffmccarthy:
Thanks for the guidance!  When I get a little time, I should be able to do the work to consolidate the min and max into a single value.

Edit:  I mis-read the post above -- the proposal is to remove the configurability entirely.  That should be feasible as well, and avoids having to handle the question of how users with an existing chatmail.ini move to a different keyword.  I'll work on the necessary changes soon.

Edit 2:  I guess both possibilities are suggested above -- a single value, or no value.  I shouldn't try to understand requirements before I'm awake.  :-)  How should we proceed?

link2xt:
@cliffmccarthy I opened a related issue about username and password generation: https://github.com/chatmail/relay/issues/697
This is a larger change and not discussed yet, but I think we want to get rid of HTTPS requests during account creation one way or another and this flexibility of having chatmail operator configure username lengths is in the way.

Issue #563, opened 2025-04-25
Error message when sending PGP Message with Armor Headers


mueckinger:
When sending a message containing Armor Headers (see example below), I'll get the error message:
`523 Encryption Needed: Invalid Unencrypted Mail`
Armor Headers fully comply with RFC4880 and RFC3156.
I was told in the chat that this is caused by the following checks in filtermail.py and that Armor Headers are not allowed:
```python
def check_armored_payload(payload: str):
    prefix = "-----BEGIN PGP MESSAGE-----\r\n\r\n"
    if not payload.startswith(prefix):
        return False
```

Improvements:

1. Parts that deviate from [mentioned standards](https://github.com/chatmail/core/blob/main/standards.md), like not allowing Armor Headers, should be documented.

2. The error message is wrong and misleading as the message itself **is encrypted and fully valid**

Sample message with Armor Headers:
```
Return-Path: <REDACTED>
Mime-Version: 1.0
Date: Sat, 19 Apr 2025 10:12:17 +0200
From: REDACTED
To: REDACTED
Subject: [...]
MIME-Version: 1.0
Autocrypt: addr=REDACTED; prefer-encrypt=mutual
Content-Type: multipart/encrypted; boundary=foo;
 protocol="application/pgp-encrypted"; charset=UTF-8
Envelope-To: <REDACTED>

--foo
Content-Type: application/pgp-encrypted

Version: 1

--foo
Content-Type: application/octet-stream

-----BEGIN PGP MESSAGE-----
Version: GopenPGP 2.7.4
Comment: [https://gopenpgp.org](https://github.com/chatmail/relay/issues#)

wV4DdCVjRfOT3TQSAQdAhQMNpbhYRucWjnY1I71LhWGGs8WpfmT13v2m5otldWEw
3f18e0kpohjzY3Rar6HmT9NBSwWGYDIqWVxwezxH/PPS2zpVvKoqpm0XAPBlf3aQ
0j0BAaqSfs00X4T3/Z6UilkWth36+cD1cejT12eRi9x4dL6HbfwlhCxkljuqn61G
8I9ZzpNjIYbQ6YvQUbym
=3Dvins
-----END PGP MESSAGE-----

--foo--
```

mueckinger:
Ok, I switched from the helper "github.com/ProtonMail/gopenpgp/v2/helper", which can create the armored message with a single command, to "github.com/ProtonMail/gopenpgp/v3/crypto". Four more lines of code, but without the armor headers.

[...]

PR #655, opened 2025-10-07
filtermail: allow Version comment in incoming PGP messages 

### LINKED ISSUE #616: encrypted messages from Proton Mail are rejected
while sending from Chatmail to Proton Mail is fine, messages the other way round are rejected with the error `Invalid Unencrypted Mail (in reply to end of DATA command)`

i suspect, this  comes from the OpenPGP header `Version: ProtonMail`, similar to initially also not understood `Version: GopenPGP 2.7.4` at #563

[Full rejection mail](https://gist.github.com/r10s/8c4efdab1fbdb06e4ef4a4413978b6ce)

for testing:
- **create** a free Proton Mail account
- **validate** that account, otherwise, you cannot receive encrypted messages
- **export** public key as `.asc` in Proton Mail at "Settings / All Settings / Encryption and Keys"
- **import** that key in Delta Chat using [.asc to vCard](https://webxdc.org/apps/#r10s-asc-to-vcard)
- **send** a message from Delta Chat to Proton -> that works
- **trust** the key in Proton Mail - otherwise, without any UI feedback [sic!], you'll reply unencrypted 
- **reply** to the message

when using chatmail, the last point results in an error.

when using classic mail, the last point works as well

missytake:
fix #616

### DIFF
--- a/chatmaild/src/chatmaild/filtermail.py
+++ b/chatmaild/src/chatmaild/filtermail.py
@@ -107,7 +107,6 @@ def check_armored_payload(payload: str, outgoing: bool):
     if payload.startswith(version_comment):
         version_line = payload.splitlines()[0]
         payload = payload.removeprefix(version_line)
-        print("Removed:" + version_line)
         if outgoing:
             return False
 


link2xt:
Test is missing.

If we do this anyway, then I'd rather not hardcode Protonmail and allow other `Version` headers as well.

And this should not be allowed for outgoing mails, only for incoming ones.

missytake:
> And this should not be allowed for outgoing mails, only for incoming ones.

Ah wait, that's still missing.

missytake:
Now outgoing mails with `Version:` comment are prohibited and tests pass as well :)

Let's hope nobody spams nine users unencrypted with spam in the `Version:` comment

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -96,6 +102,18 @@ def check_armored_payload(payload: str):
         return False
     payload = payload.removesuffix(suffix)
 
+    # Remove Protonmail version comment.
```
link2xt: This is not about protonmail anymore

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -96,6 +102,18 @@ def check_armored_payload(payload: str):
...
+    proton_version_comment = "Version:"
```
link2xt: Space is required, so you can as well check for "Version: ": https://www.rfc-editor.org/rfc/rfc9580.html#name-armor-headers

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -96,6 +102,18 @@ def check_armored_payload(payload: str):
         return False
     payload = payload.removesuffix(suffix)
 
+    # Disallow comments in outgoing messages
+    version_comment = "Version: "
+    if payload.startswith(version_comment):
+        version_line = payload.splitlines()[0]
+        payload = payload.removeprefix(version_line)
+        print("Removed:" + version_line)
```
link2xt: Debug printing?

hpk42:
> Test is missing.
> 
> If we do this anyway, then I'd rather not hardcode Protonmail and allow other `Version` headers as well.
> 
> And this should not be allowed for outgoing mails, only for incoming ones.

IMO checking specifically for Protonmail is sufficient for now. If we get another case, we can consider generalizing, but i wouldn't do this ahead.  Basically we want to be as strict as we can about the checks, to minimize attack/spam surface.

missytake:
> > Test is missing.
> > If we do this anyway, then I'd rather not hardcode Protonmail and allow other `Version` headers as well.
> > And this should not be allowed for outgoing mails, only for incoming ones.
> 
> IMO checking specifically for Protonmail is sufficient for now. If we get another case, we can consider generalizing, but i wouldn't do this ahead. Basically we want to be as strict as we can about the checks, to minimize attack/spam surface.

There was already the case with GopenPGP: https://github.com/chatmail/relay/issues/563

Issue #622, opened 2025-08-26
capsys.readouterr does not capture output from the Out class


Keonik1:
Problem: https://github.com/Keonik1/relay/pull/2#discussion_r2299996176

Brief description:

capsys.readouterr() does not read messages that are output via the Out class and the Out.green, Out.red, Out.yellow functions themselves.

The likely reason is that Out redirects output directly to stdout, consoles, rather than to Python's stdout, which is redefined by pytest.

Maybe this will be usefull: https://stackoverflow.com/a/52844694

missytake:
in https://github.com/chatmail/relay/pull/656, I tried with `capfd` instead of `capsys`, that didn't help either.

PR #656, opened 2025-10-07
cmdeploy: add --force to cmdeploy init for recreating chatmail.ini

missytake:
Needed for https://github.com/chatmail/relay/pull/614 - `setup_chatmail_docker.sh` recreates chatmail.ini based on environment variables on every run.

### DIFF
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -25,9 +25,6 @@ from .sshexec import SSHExec
 # cmdeploy sub commands and options
 #
 
-def is_pytest():
-    return "PYTEST_CURRENT_TEST" in os.environ
-
 
 def init_cmd_options(parser):
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -345,6 +367,12 @@ def get_sshexec():
     args.get_sshexec = get_sshexec
 
     out = Out()
+    if is_pytest(): ## issue: https://github.com/chatmail/relay/issues/622
```
hpk42: not a good idea to make production-used code test whether it's in a test environment because that can easily introduce bugs that only occur in production.

cmdeploy/src/cmdeploy/tests/test_cmdeploy.py
```diff
@@ -25,11 +25,17 @@ def test_parser(self, capsys):
 
     def test_init_not_overwrite(self, capsys):
```
hpk42, line 26: i believe you could use the "capfd" fixture which captures on the file-descriptor level instead of on the python level of 'sys.std*' that "capsys" provides. 

This should also resolve #622 without needing to do "is_pytest()" and similar shenenigans.

hpk42 (changes requested):
see comments.

cmdeploy/src/cmdeploy/tests/test_cmdeploy.py
missytake: capfd doesn't work either... I reverted it to print().

PR #658, opened 2025-10-07
www: add markdown language switcher

missytake:
From #614 - but it really is its own feature, so it should be in a separate PR.


missytake:
Closing this for now, we have too many open PRs and it's not clear we want this. In general, it is much easier to design a completely custom web page by now: https://chatmail.at/doc/relay/getting_started.html#custom-web-pages

r10s:
+1 for closing. this seems are premature optimisation.

in practise, the relays we got aware over the last months, are pretty locally targeted - and are already translated then, to the locals they want to support.

and if a relay operator really wants to support different languages, they can do that by the tools available - if in doubt, add a "see english below", some js or whatnot. this seems nothing we need to think about currently.

Issue #629, opened 2025-09-11
Rework SSHExec


missytake:
SSHExec is an object in `cmdeploy.py`, which we can use to get the result of bash commands on the relay. Since it was last reworked in https://github.com/chatmail/relay/pull/396, our needs have evolved. Obstacles of the current implementation:

- it always invariably connects to `mail_domain`, the relay host
- it is initialized in cmdeploy.py's `main()` function, so the functions which come afterwards can hardly modify it
- it doesn't offer to simply pass a bash command, you need to pass a function to it which calls a `remote.rshell` method
- it offers two different callback functions for console output, but no option to suppress it completely
- the various functions which execute the bash command on the target system are spread out over many different files: `cmdeploy.py`, `dns.py`, `sshexec.py`, `remote/rdns.py`, and `remote/rshell.py`

Situations in which we need it to work differently:

- in https://github.com/chatmail/relay/pull/628, we want to run commands on a reverse-proxy
- in https://github.com/chatmail/relay/pull/614, we want to run commands locally
- in https://github.com/Keonik1/relay/pull/3, we want to run DNS queries locally instead of on the relay
- in https://github.com/chatmail/relay/compare/docker-dns...docker-tests, we want to run commands in a docker container
- in https://github.com/chatmail/relay/pull/620 we need to suppress the "[ssh] login to..." output

The current implementation has a certain elegance to it, to make it easier to test; I wonder whether it can be enhanced to fit all these needs. At the same time it would be nice to make it simpler and more maintainable (it took me a few weeks to understand the current obstacles well enough to write down this issue).

PR #659, opened 2025-10-08
cmdeploy: allow to run SSH commands locally

### LINKED ISSUE #604: `--ssh-host` does not actually change the ssh host

missytake:
fix #604
related to #629
pulled out of https://github.com/Keonik1/relay/pull/3


cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
```diff
@@ -31,7 +31,7 @@ def test_logged(self, sshexec, maildomain, capsys):
         )
         out, err = capsys.readouterr()
         assert err.startswith("Collecting")
-        assert err.endswith("....\n")
+        #assert err.endswith("....\n")
```
missytake, line 36: I disabled parts of this test; cmdeploy dns works as expected, but `capsys.readouterr()` fails to capture what `log_progress()` writes to `sys.stderr`. It looks related to https://stackoverflow.com/a/29567317

It's weird because capturing what `print_stderr()` posts to `sys.stderr` works. But we can't put `log_progress()` in `sshexec.py`, because then we can't import it in `remote/rdns.py`, as it is outside of the remote module, and `from cmdeploy.sshexec import log_progress` somehow doesn't work.

It's messy, and before I waste another hour with this I'd rather disable this in the test; it's just checking whether the output is pretty, after all, nothing critical. But if anyone has an idea what is going wrong, feel free to contribute.

What could also help is dropping `remote` as a separate python module. That would anyway help with code structure, right now it is spread out over many files. But by now, I understand only half of what's going on there, so someone else would need to do this.

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -74,7 +74,7 @@ def run_cmd_options(parser):
     parser.add_argument(
         "--ssh-host",
         dest="ssh_host",
-        help="specify an SSH host to deploy to; uses mail_domain from chatmail.ini by default",
+        help="Deploy to 'localhost' or to a specific SSH host",
```
hpk42: why drop the mail_domain from chatmail.ini remark?

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -139,11 +143,17 @@ def dns_cmd_options(parser):
         default=None,
         help="write out a zonefile",
     )
+    parser.add_argument(
+        "--ssh-host",
+        dest="ssh_host",
+        help="Run the DNS queries on 'localhost' or on a specific SSH host",
+    )
```
hpk42: i'd make the addition of this option into a function, and call it from both places where "--ssh-host" is supported (run and dns, maybe others?)

cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
hpk42: k, could you add a small comment like "# XXX could not figure out how capturing can be made to work properly"

cmdeploy/src/cmdeploy/cmdeploy.py
missytake: Because the argument is optional anyway - the default value doesn't need to be specified imho.

```suggestion
        help="Deploy to 'localhost' or to a specific SSH host instead of chatmail.ini's mail_domain",
```

Maybe that would be an option, but when we also add "or @docker" to the description, it becomes too long.

PR #660, opened 2025-10-08
cmdeploy: prepare for being able to run commands in docker containers

missytake:
Based on #659, this is also pulled out of https://github.com/Keonik1/relay/pull/3. It enables us to run some DNS checks in a docker container; it could also be part of #659, but I put it in a separate PR, as it isn't useful before #614.

### DIFF
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -99,7 +99,7 @@ def run_cmd(args, out):
     pyinf = "pyinfra --dry" if args.dry_run else "pyinfra"
 
     cmd = f"{pyinf} --ssh-user root {ssh_host} {deploy_path} -y"
-    if ssh_host in ["localhost", "docker"]:
+    if ssh_host in ["localhost", "@docker"]:
         cmd = f"{pyinf} @local {deploy_path} -y"
 
     if version.parse(pyinfra.__version__) < version.parse("3"):
@@ -366,7 +366,7 @@ def get_parser():
 def get_sshexec(ssh_host: str, verbose=True):
     if ssh_host in ["localhost", "@local"]:
         return LocalExec(verbose, docker=False)
-    elif ssh_host == "docker":
+    elif ssh_host == "@docker":
         return LocalExec(verbose, docker=True)
     if verbose:
         print(f"[ssh] login to {ssh_host}")


hpk42:
I suggest to rename "docker" to ":docker" throughout the PR to disambiguate from regular ssh hosts.  
Also it's a bit weird to merge this PR before the actual docker PR, and there are no tests or CI yet.

missytake:
What about `@docker`? pyinfra uses `@local` for running on localhost, we support that as well, that would make it a bit more consistent.

I wonder if any of those options could mess with SSH configs...

hpk42:
yes, @docker is better if we have @local already.  i think "ssh @docker" is invalid, so i guess it's ok wrt ssh config. anything with an @ prefix is not an ssh target.

missytake:
> there are no tests or CI yet.

The only thing I see that makes sense to test for this PR is whether `pre_command` is passed correctly to sub-functions. But I fail mocking the `shell()` function to see the output of a command instead of executing it. Do you have an idea what I'm doing wrong here? Replacing `cmdeploy.remote.rshell.shell` with `test_cmdeploy.test_dns_when_ssh_docker.<locals>.shell` seems to work in all methods *except* the one that counts.

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -331,19 +341,23 @@ def get_parser():
...
 
 
+def get_sshexec(ssh_host: str, verbose=True):
+    if ssh_host in ["localhost", "@local"]:
+        return "localhost"
+    elif ssh_host == "@docker":
+        return "@docker"
+    if verbose:
+        print(f"[ssh] login to {ssh_host}")
+    return SSHExec(ssh_host, verbose=verbose)
```
hpk42, line 373: this looks weird to return strings or an object instance.  Usually functions should return a common type. 
Instead i would return a LocalExec instance or DockerExec instance, next to the SSHExec instance. 
roughly like 

```
class LocalExec:
     def __init__(self, verbose=False):
          self.verbose = verbose 
     def logged(self, func, **kwargs):
           if self.verbose:
               print(f"calling locally: {func.__name__}(**{kwargs})")
           return func(**kwargs)

```
and DockerExec similar.  So get_ssh_exec would return something where you can call "logged()" on. 

For DockerExec, it can check whether `func == remote.rdns.perform_initial_checks` and add `pre_command` to the kwargs.

PR #662, opened 2025-10-08
docker: add traefik support

missytake:
I pulled out traefik support from #614 so we can think about it individually, and merge normal docker support first.


hpk42:
why traefik and not any other alternative to acme? 
It may be better to make using acmetool optional, and only document how to integrate non-acmetool things, providing some hooks or so.

Keonik1:
@hpk42 
In my experience, Traefik has one of the best supports for working with Docker containers. All you need to do is specify the labels, and Traefik will do the rest.

In other alternatives (except for Kubernetes Ingress), this is much less convenient and native.

lig:
@Keonik1 Also, supporting Traefik opens up k3s as a supported platform which looks like a great choice (and a popular one) for hosting a small project and self-hosting in general.

ccclxxiii:
> @Keonik1 Also, supporting Traefik opens up k3s as a supported platform which looks like a great choice (and a popular one) for hosting a small project and self-hosting in general.

we run several k3s clusters and the default is traefik.  it also has the benefit of load-balancing tcp which may be useful depending on how container setup works in-cluster.  let's find out.

PR #668, opened 2025-10-13
Don't print echobot link when disabling mail

cliffmccarthy:
- On a fresh install, if cmdeploy is run the first time with the --disable-mail option, the echobot invite-link.txt file will not exist yet.
- Only print the echobot invite link if --disable-mail was not specified.  This fixes the fresh-install error case, and also makes sense when disabling mail in general, because the echo bot will not be available at that time.


cliffmccarthy:
My deployment workflow is to first run cmdeploy with the `--disable-mail` argument, and then subsequently run it again without that option to enable the services.  This might be unusual, but it's how I came across the error when it tries to print the file that hasn't been created by echobot yet.

PR #669, opened 2025-10-14
test: Handle Git errors in test_deployed_state()

cliffmccarthy:
- This is a counterpart to pull request #607.  Revised test_deployed_state() to perform the same error-handling on Git commands that cmdeploy does.  If 'git rev-parse' returns an error, the value "unknown" is used.  If 'git diff' returns an error, the null string is used.
- This fixes failures in environments where Git is not installed or where the .git subdirectory is not present (as long as the server was deployed in the same way).

### DIFF
--- a/cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
+++ b/cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
@@ -209,8 +209,14 @@ def test_expunged(remote, chatmail_config):
 
 
 def test_deployed_state(remote):
-    git_hash = subprocess.check_output(["git", "rev-parse", "HEAD"]).decode()
-    git_diff = subprocess.check_output(["git", "diff"]).decode()
+    try:
+        git_hash = subprocess.check_output(["git", "rev-parse", "HEAD"]).decode()
+    except Exception:
+        git_hash = "unknown\n"
+    try:
+        git_diff = subprocess.check_output(["git", "diff"]).decode()
+    except Exception:
+        git_diff = ""
     git_status = [git_hash.strip()]
     for line in git_diff.splitlines():
         git_status.append(line.strip().lower())


cliffmccarthy:
I'm not sure what the value of this test is, if it's just a replica of the code it is testing.  But this stops it from failing in environments where it can't access the repo Git data for whatever reason.

cliffmccarthy:
> maybe we should rather `pytest.skip()` it if git isn't installed?

That seems like a reasonable way to handle it.  That makes it more visible that it's not checking anything.

cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
```diff
@@ -209,8 +209,14 @@ def test_expunged(remote, chatmail_config):
 
 
 def test_deployed_state(remote):
-    git_hash = subprocess.check_output(["git", "rev-parse", "HEAD"]).decode()
-    git_diff = subprocess.check_output(["git", "diff"]).decode()
+    try:
+        git_hash = subprocess.check_output(["git", "rev-parse", "HEAD"]).decode()
+    except Exception:
```
missytake:
```suggestion
    except FileNotFoundError:
```

There might be other exceptions - we want to explicitly catch that git isn't installed, right?

missytake (approved):
What was the exception for ".git not present"?

cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
cliffmccarthy: I revised the code to handle the specific exceptions for a missing 'git' command, and a missing '.git' subdirectory.  Do you think the corresponding code in the main deploy_chatmail() function should also be updated to handle specific exceptions?

missytake: Aaaaah now I see that you did copy the exception handling from `cmdeploy/__init__.py` - that *did* make sense :upside_down_face: 

Actually then let's use the `unknown` content for /etc/chatmail-version, and also check for it in the test. We plan to use this to check whether chatmail is installed, so we certainly want to upload _something_, and uploading `unknown` if git isn't working is not bad.

Your very first commit was fine actually. Let's revert the other two? :see_no_evil:

cliffmccarthy: No problem!  I force-pushed the branch back to the original commit.

missytake: Great, thanks :)

PR #673, opened 2025-10-17
postfix: accept whole mail before passing it to filtermail

missytake:
This seems to fix issues with slow delivery on nine.testrun.org :) 

The problem was that filtermail's aiosmtpd dependency had problems with postfix passing parts of the incoming SMTP commands onwards to it; this caused a lot of asyncio back-and-forth. Now filtermail only gets a complete mail it can check in one go. Moving filtermail from async-based to thread-based processing might also help, maybe in rust, but is a lot more effort.

### DIFF
--- a/cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
+++ b/cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
@@ -195,9 +195,8 @@ def test_exceed_rate_limit(cmsetup, gencreds, maildata, chatmail_config):
         except smtplib.SMTPException as e:
             if i < chatmail_config.max_user_send_per_minute:
                 pytest.fail(f"rate limit was exceeded too early with msg {i}")
-            outcome = e.recipients[user2.addr]
-            assert outcome[0] == 450
-            assert b"4.7.1: Too much mail from" in outcome[1]
+            assert e.smtp_code == 450
+            assert b"4.7.1: Too much mail from" in e.smtp_error
             return
     pytest.fail("Rate limit was not exceeded")
 


missytake:
Huh, `cmdeploy/src/cmdeploy/tests/online/test_1_basic.py::test_exceed_rate_limit` is failing.

link2xt:
As far as I understand `smtpd_proxy_options=speed_adjust` did not actually help. I also don't see why it should help and why "aiosmtpd" should not be able to handle practically unlimited number of connections even if most of them get stuck waiting for actual data. Buffering the whole mail in postfix smtpd before passing it to filtermail will reduce the number of connections filtermail needs to handle at the same time, but if it is not a problem I'd rather not tweak the settings.

link2xt:
This change is also a likely cause of this error from the logs that appeared after deploying it:
```
postfix/smtps/smtpd[2219812]: proxy-reject: END-OF-MESSAGE: 451 4.3.0 Error: queue file write error
```

Without this change messages should not even be written into temporary queue. Writing messages to disk before passing them to filtermail is another step that may fail and does not speed up sending.

link2xt:
I opened an issue to track the problem this change is trying to solve: https://github.com/chatmail/relay/issues/675

ARCHITECTURE.md
```diff
@@ -48,3 +48,19 @@ graph LR;
 The edges in this graph should not be taken too literally; they
 reflect some sort of communication path or dependency relationship
 between components of the chatmail server.
+
+## Message between users on the same relay
```
link2xt, line 52: This documentation change is better moved in a separate PR, then it can also be merged quickly.

missytake:
> This change is also a likely cause of this error from the logs that appeared after deploying it:
> 
> ```
> postfix/smtps/smtpd[2219812]: proxy-reject: END-OF-MESSAGE: 451 4.3.0 Error: queue file write error
> ```
> 
> Without this change messages should not even be written into temporary queue. Writing messages to disk before passing them to filtermail is another step that may fail and does not speed up sending.

No, that was caused by restarting filtermail service.

ARCHITECTURE.md
missytake: I moved it to #678 :)

missytake:
Let's close this for now :) we can look at it again if #676 isn't sufficient for solving filtermail's performance issues.

cmdeploy/src/cmdeploy/tests/online/test_1_basic.py
```diff
@@ -195,9 +195,8 @@ def test_exceed_rate_limit(cmsetup, gencreds, maildata, chatmail_config):
         except smtplib.SMTPException as e:
             if i < chatmail_config.max_user_send_per_minute:
                 pytest.fail(f"rate limit was exceeded too early with msg {i}")
-            outcome = e.recipients[user2.addr]
-            assert outcome[0] == 450
-            assert b"4.7.1: Too much mail from" in outcome[1]
+            assert e.smtp_code == 450
```
link2xt, line 198: Test fixes can also be moved to a separate PR?

Issue #675, opened 2025-10-18
Outgoing filtermail is slow


link2xt:
It has been reported that filtermail is slowing down sending of messages.

According to the logs there was a 10 second delay between processing MAIL FROM and RCPT TO and between processing RCPT TO and DATA:
```
Oct 18 15:45:24 localhost postfix/smtps/smtpd[2147068]: NOQUEUE: client=unknown[XXX], sasl_method=PLAIN, sasl_username=alice@example.org
Oct 18 15:45:59 localhost filtermail[1881070]: INFO:mail.log:('127.0.0.1', 39566) >> b'MAIL FROM:<alice@example.org> BODY=8BITMIME SMTPUTF8'
Oct 18 15:45:59 localhost filtermail[1881070]: INFO:root:handle_MAIL from alice@example.org
Oct 18 15:45:59 localhost filtermail[1881070]: INFO:mail.log:('127.0.0.1', 39566) sender: alice@example.org
Oct 18 15:46:12 localhost filtermail[1881070]: INFO:mail.log:('127.0.0.1', 39566) >> b'RCPT TO:<alice@example.org>'
Oct 18 15:46:12 localhost filtermail[1881070]: INFO:mail.log:('127.0.0.1', 39566) recip: alice@example.org
Oct 18 15:46:24 localhost postfix/smtps/smtpd[944844]: NOQUEUE: client=unknown[2a01:4f8:c014:43c5::1], sasl_method=PLAIN, sasl_username=alice@example.org
Oct 18 15:46:24 localhost filtermail[1881070]: INFO:root:Processing DATA message from alice@example.org
Oct 18 15:46:24 localhost filtermail[1881070]: INFO:root:Check_encrypted from alice@example.org finished
Oct 18 15:46:24 localhost postfix/cleanup[2280407]: 96E6F17C2C9E: replace: header Subject: [...] from unknown[127.0.0.1]; from=<alice@example.org> to=<alice@example.org> proto=ESMTP helo=<[127.0.0.1]>: Subject: [...]
Oct 18 15:46:24 localhost postfix/qmgr[2376868]: 96E6F17C2C9E: from=<alice@example.org>, size=2984, nrcpt=2 (queue active)
[...]

PR #676, opened 2025-10-18
filtermail: run CPU-intensive handle_DATA in a thread pool executor

link2xt:
See
<https://docs.python.org/3/library/asyncio-eventloop.html#executing-code-in-thread-or-process-pools> for the documentation.

This should avoid processing of large messages from hogging asyncio thread and delaying async operations like accepting new connections.

This is addressing #675

### DIFF
--- a/chatmaild/src/chatmaild/filtermail.py
+++ b/chatmaild/src/chatmaild/filtermail.py
@@ -241,6 +241,10 @@ class OutgoingBeforeQueueHandler:
         return "250 OK"
 
     async def handle_DATA(self, server, session, envelope):
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(None, self.sync_handle_DATA, envelope)
+
+    def sync_handle_DATA(self, envelope):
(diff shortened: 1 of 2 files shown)


adbenitez (approved):

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -295,7 +295,10 @@ def __init__(self, config):
 
     async def handle_DATA(self, server, session, envelope):
         log_info("handle_DATA before-queue")
-        error = self.check_DATA(envelope)
+
+        # Run CPU-intensive checks in a thread pool.
```
adbenitez: mmmh but you added it actually to the incoming messages handler, this is missing in the outgoing messages handler class where in theory is the actual bottleneck

link2xt: Updated, should be in both.

link2xt:
`sendmail` should also be moved to the thread pool. Basically the whole handle_DATA is synchronous.

PR #677, opened 2025-10-19
filtermail: better performance for removing Version

missytake:
This reduces 100,000 calls of `src/chatmaild/tests/test_filtermail.py::test_check_armored_payload` from 11.00 to 10.52 seconds.

Related to #675

### DIFF
--- a/chatmaild/src/chatmaild/filtermail.py
+++ b/chatmaild/src/chatmaild/filtermail.py
@@ -104,7 +104,7 @@ def check_armored_payload(payload: str, outgoing: bool):
     # Disallow comments in outgoing messages
     version_comment = "Version: "
     if payload.startswith(version_comment):
-        splitindex = payload.find("\r\n") + 4
+        splitindex = payload.find("\r\n") + 2
         payload = payload[splitindex:]
         if outgoing:
             return False


chatmaild/src/chatmaild/filtermail.py
```diff
@@ -104,8 +105,8 @@ def check_armored_payload(payload: str, outgoing: bool):
     # Disallow comments in outgoing messages
     version_comment = "Version: "
     if payload.startswith(version_comment):
-        version_line = payload.splitlines()[0]
-        payload = payload.removeprefix(version_line)
+        splitindex = payload.find("\r\n") + 4
```
link2xt: Why +4?

link2xt: Why even do `find` for outgoing messages if the result is not used?

missytake: `find()` returns the first occurence of a line break as int. It is used one line later to remove the version comment line from the string.

link2xt: And then two lines later there is `return False`

missytake: Sorry, changed it to two - we also want to remove the `\r\n` from the payload, that's why I'm moving the splitindex two to the right here. First it was 4 because I thought `\r\n` were 4 characters instead of 2.

chatmaild/src/chatmaild/filtermail.py
```diff
@@ -104,8 +104,8 @@ def check_armored_payload(payload: str, outgoing: bool):
     # Disallow comments in outgoing messages
     version_comment = "Version: "
     if payload.startswith(version_comment):
-        version_line = payload.splitlines()[0]
-        payload = payload.removeprefix(version_line)
+        splitindex = payload.find("\r\n") + 2
+        payload = payload[splitindex:]
```
adbenitez, line 108: this is more simple and faster/effective:


```suggestion
        payload = payload.split("\r\n", maxsplit=1)[1]
```

chatmaild/src/chatmaild/filtermail.py
adbenitez: I created an alternative PR: https://github.com/chatmail/relay/pull/679

chatmaild/src/chatmaild/filtermail.py
link2xt: Unlike the current code this will fail with IndexError if no `\r\n` is found.

link2xt: Can use `payload.partition("\r\n")[2]`, this should always work.

adbenitez: there must be \r\n afterwards `Version: ` tho? (and if not rejection with error is better than allowing suspicious message with fake `Version:` comment)

missytake:
Closed in favor of #679

PR #680, opened 2025-10-19
add a system test for running the filtermail module 

hpk42:
only adds a test that runs filtermail as a subprocess (similar to how it runs in a chatmail-relay setup) and checks that incoming and outgoing filters basically work for passing through encrypted, and rejecting unencrypted mail. 
the test captures any setup errors (e.g. in main()) and the aiosmtpd loop of receiving messages via smtpd.

### DIFF
--- a/chatmaild/pyproject.toml
+++ b/chatmaild/pyproject.toml
@@ -13,7 +13,6 @@ dependencies = [
   "filelock",
   "requests",
   "crypt-r >= 3.13.1 ; python_version >= '3.11'",
-  "pytest-localserver",
 ]
 
 [tool.setuptools]
@@ -72,5 +71,6 @@ commands =
 [testenv]
 deps = pytest 
        pdbpp 
+       pytest-localserver
 commands = pytest -v -rsXx {posargs}
 """


hpk42:
the new test always passes on my machine, but on CI sometimes fails with a connection-refused, see https://github.com/chatmail/relay/actions/runs/18635645301/job/53127126079?pr=680
ASFAIK the smtp-localserver (where filtermail fails to connect to) should run, but maybe it's some localhost/::1 ipv4/ipv6 issue?

chatmaild/src/chatmaild/tests/test_filtermail_blackbox.py
```diff
@@ -0,0 +1,75 @@
...
+            print(out.decode("ascii"))
+            print(err.decode("ascii"), file=sys.stderr)
+
+        request.addfinalizer(fin)
+        return p
+
+    return popen
+
+
+def find_free_port():
```
link2xt: By the time this finishes, the port is already freed. As it is not reliable anyway, probably easier to just hardcode the port to 10025 and require that it is not used on the test machine.

chatmaild/src/chatmaild/tests/test_filtermail_blackbox.py
```diff
@@ -0,0 +1,78 @@
...
+import sys
+
+import pytest
+
+
+@pytest.fixture
+def smtpserver():
+    from pytest_localserver import smtp
+
+    server = smtp.Server("localhost")
```
link2xt: If this `localhost` happens to be IPv6-localhost, filtermail will fail to connect. This should be explicitly `127.0.0.1`, otherwise it is probably whatever you get from /etc/hosts

chatmaild/src/chatmaild/tests/test_filtermail_blackbox.py
hpk42: using a fixed port now.

chatmaild/src/chatmaild/tests/test_filtermail_blackbox.py
hpk42: this seems to make it work. thanks.

chatmaild/pyproject.toml
```diff
@@ -13,6 +13,7 @@ dependencies = [
   "filelock",
   "requests",
   "crypt-r >= 3.13.1 ; python_version >= '3.11'",
+  "pytest-localserver",
```
link2xt: This should go into `testenv` section of tox.ini, next to `pytest`, at the end of this file. We don't need pytest-localserver installed on the server, especially without pytest itself.

link2xt (approved):

PR #679, opened 2025-10-19
improve check_armored_payload()

adbenitez:
improve the code introduced in https://github.com/chatmail/relay/pull/655
partially solves https://github.com/chatmail/relay/issues/675

alternative to https://github.com/chatmail/relay/pull/677
 this is a comparison of this approach vs the other PR:
image

PR #683, opened 2025-10-20
filtermail: further optimize check_armored_payload()

missytake:
With python3.11 this seems to be the fastest option: https://github.com/chatmail/relay/pull/679#issuecomment-3423431811

All the incoming PGP-encrypted mails with Version headers will be processed so much faster now, a very important and necessary optimization during which we learned a lot about python string operations.


hpk42:
for the record, i am critical of this micro-optimization:

- it measures not real messages but just a short string which is very misleading -- for larger messages partition will likely be much worse than find even with python3.11 
- one call takes 0.2 micro-seconds, or 0.0002 milliseconds -- and we have maximum of 100 messages per second (more like 10-20).

PR #685, opened 2025-10-20
Require TLS 1.2 for outgoing SMTP connections

link2xt:


link2xt:
Cloudflare has some data on the distribution of versions for SMTP traffic (that passes DKIM, because this is what we are interested in):
https://radar.cloudflare.com/explorer?dataSet=email.security&groupBy=tls_version&filters=dkim%253DPass

There are some weird spikes of TLS 1.1 during the week, likely mass mailing and not normal messages, so average numbers are not really meaningful:
image
TLS 1.2 we definitely cannot cut if it is around 10% of the email traffic.

link2xt:
`journalctl -t postfix/smtp` has "server certificate verification failed" and "TLS is required, but was not offered by host" errors all the time. Don't know how TLS version errors will look like, but they will likely show up there.

PR #687, opened 2025-10-22
guard expire/fsreport file iteration against vanishing, improve reporting

hpk42:
also activates actual deletion (after quite some dry test runs on nine)

### DIFF
--- a/cmdeploy/src/cmdeploy/service/chatmail-expire.service.f
+++ b/cmdeploy/src/cmdeploy/service/chatmail-expire.service.f
@@ -5,5 +5,5 @@ After=network.target
 [Service]
 Type=oneshot
 User=vmail
-ExecStart=/usr/local/lib/chatmaild/venv/bin/chatmail-expire /usr/local/lib/chatmaild/chatmail.ini -v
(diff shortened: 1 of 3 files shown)


cmdeploy/src/cmdeploy/service/chatmail-expire.service.f
```diff
@@ -5,5 +5,5 @@
 [Service]
 Type=oneshot
 User=vmail
-ExecStart=/usr/local/lib/chatmaild/venv/bin/chatmail-expire /usr/local/lib/chatmaild/chatmail.ini -v
+ExecStart=/usr/local/lib/chatmaild/venv/bin/chatmail-expire /usr/local/lib/chatmaild/chatmail.ini -v --remove
```
link2xt, line 8: What is this change? The service was in dry-run mode before?

link2xt (approved):

cmdeploy/src/cmdeploy/service/chatmail-expire.service.f
link2xt: I see it is in commit message, actually looks like the most important change.

PR #688, opened 2025-10-22
acmetool: Increase RSA key size from 2048 to 4096

link2xt:


link2xt:
We can also switch to ECDSA instead, but I did not find the information on which curves Let's Encrypt actually supports currently:
https://github.com/hlandau/acmetool/blob/d3428cff2614cbd31d91484e7c7bd188db53bcf1/_doc/SCHEMA.md?plain=1#L247-L251

For letsencrypt.org my browser shows ECC with "256" as the key size. I think it is true that practically everyone who uses ECC keys uses nistp256 because other key sizes are significantly slower without much advantages. So whether other sizes are supported does not matter that much, if we switch to ECDSA then we can use the default nistp256.

PR #689, opened 2025-10-22
acmetool: use ECDSA keys instead of RSA

link2xt:
Replacement for https://github.com/chatmail/relay/pull/688

256-bit ECDSA keys are smaller than RSA-2048 and are roughly equivalent to RSA-3072. letsencrypt.org, google.com, github.com all use the same keys. See also https://letsencrypt.org/2020/09/17/new-root-and-intermediates.html#why-we-issued-an-ecdsa-root-and-intermediates for comparison.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- acmetool: use ECDSA keys instead of RSA
+  ([#689](https://github.com/chatmail/relay/pull/689))
+
 - Require TLS 1.2 for outgoing SMTP connections
   ([#685](https://github.com/chatmail/relay/pull/685))
 
--- a/cmdeploy/src/cmdeploy/acmetool/target.yaml.j2
+++ b/cmdeploy/src/cmdeploy/acmetool/target.yaml.j2
@@ -1,7 +1,8 @@
 request:
   provider: https://acme-v02.api.letsencrypt.org/directory
   key:
-    type: rsa
+    type: ecdsa
+    ecdsa-curve: nistp256
   challenge:
     webroot-paths:
     - /var/www/html/.well-known/acme-challenge


link2xt:
Keys are generated on every request, there is no need to do anything to update, it will be updated once previous certificate expires.

missytake:
What do you think about removing `/var/lib/acme` on staging2, deploying this branch manually, and check the cert to test this change?

link2xt:
I manually deployed it on https://ci-chatmail.testrun.org/, removed `/var/lib/acme/keys` and `/var/lib/acme/certs`, rebooted it and now it has a new ECC cert.

link2xt:
Removing the whole /var/lib/acme was a bad idea, did it on c2 and now it is ratelimited because CAA record was not updated with the new account.

missytake:
> Removing the whole /var/lib/acme was a bad idea, did it on c2 and now it is ratelimited because CAA record was not updated with the new account.

You can find a backup on:

```
root@ns:~# ls -la /tmp/acme
insgesamt 36
drwxr-xr-x  9 1001 1001 4096 16. Feb 2025  .
drwxrwxrwt 17 root root 4096 24. Okt 08:47 ..
drwx------  3 1001 1001 4096 16. Feb 2025  accounts
drwxr-xr-x 23 1001 1001 4096 16. Okt 14:23 certs
drwxr-xr-x  2 1001 1001 4096 16. Feb 2025  conf
drwxr-xr-x  2 1001 1001 4096 16. Feb 2025  desired
drwx------ 63 1001 1001 4096 16. Okt 14:23 keys
drwxr-xr-x  2 1001 1001 4096 16. Okt 14:23 live
drwx------  2 1001 1001 4096 16. Okt 14:23 tmp
```

missytake:
I restored the backup for now - I guess we need to disable CAA before testing again.

link2xt:
What backup did you restore? I'm talking about c2 server.

missytake:
> Removing the whole /var/lib/acme was a bad idea, did it on c2 and now it is ratelimited because CAA record was not updated with the new account.

ah oops, c2, not staging2. nevermind

link2xt:
c2 is also running now, with ECDSA cert.

missytake:
(Can't we use ecdsa-nistp256 for DKIM as well? super annoying to fit the RSA keys into a DNS record... that would improve the situation a lot, no?)

link2xt:
> Can't we use ecdsa-nistp256 for DKIM as well?

Ed25519 is standardized but is apparently not supported in practice:
- https://serverfault.com/questions/1023674/is-ed25519-well-supported-for-the-dkim-validation
- https://www.uriports.com/blog/dkim-ed25519-adoption/

We can still experiment with it, setup a chatmail relay with ed25519 key. It should work with other chatmail relays and for non-chatmail servers that don't support ed25519 there is SPF.

PR #691, opened 2025-10-24
try to limit index cache max size

### LINKED ISSUE #644: Restore in-memory index

hpk42:
i think this also closes #644 because it should limit cache index disc usage.


link2xt:
I will deploy it to CI server now and see if it breaks the tests.

PR #692, opened 2025-10-24
Add a script for creating DNS records in Cloudflare

Keonik1:
A small script that allows you to create all the necessary DNS records in Cloudflare.

### DIFF
--- a/scripts/create_cloudflare_records.sh
+++ b/scripts/create_cloudflare_records.sh
@@ -1,0 +1,173 @@
+#!/bin/bash
+# go to https://dash.cloudflare.com/profile/api-tokens
+# "create token" -> "Edit zone DNS"
+## optionaly: rename token
+## set your zone 
+# "continue to summary" -> "create token"
+# copy your created token
+
+CLOUDFLARE_API_KEY=${CLOUDFLARE_API_KEY}
+ZONE_ID=${ZONE_ID}
+
+CHATMAIL_FULL_DNS_NAME=${CHATMAIL_FULL_DNS_NAME}
+CHATMAIL_PUBLIC_IP=${CHATMAIL_PUBLIC_IP}
+
+IPV6_ENABLED=${IPV6_ENABLED:-false}
+CHATMAIL_PUBLIC_IPv6=${CHATMAIL_PUBLIC_IPv6}
+
+#####################
+# why 'proxied' is 'false'?
+# I suppose that if Cloudflare is blocked in a country, clients cannot use Deltachat without a VPN.
+#####################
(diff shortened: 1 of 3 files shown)


scripts/create_cloudflare_records.sh
```diff
@@ -0,0 +1,173 @@
...
+create_record "$(generate_post_data_cname_record "www.$CHATMAIL_FULL_DNS_NAME")"
+
+# MX records
+create_record "$(generate_post_data_mx_record "$CHATMAIL_FULL_DNS_NAME")"
+
+# TXT records
+create_record "$(generate_post_data_txt_record "$CHATMAIL_FULL_DNS_NAME" '\"v=spf1 a ~all\"')"
+create_record "$(generate_post_data_txt_record "_dmarc.$CHATMAIL_FULL_DNS_NAME" '\"v=DMARC1;p=reject;adkim=s;aspf=s\"')"
+create_record "$(generate_post_data_txt_record "_adsp._domainkey.$CHATMAIL_FULL_DNS_NAME" '\"dkim=discardable\"')"
+create_record "$(generate_post_data_txt_record "opendkim._domainkey.$CHATMAIL_FULL_DNS_NAME" '\"v=DKIM1;k=rsa;p=;s=email;t=s\"')"
```
link2xt, line 166: `p=` value is empty, so there is no actual DKIM key.

ccclxxiii:
> CAA record is simply not created, DKIM key is missing.
> 
> Also this script will need to be maintained if merged and we don't use cloudflare so cannot test it.
> 
> If we want to automate uploading DNS records to APIs, I would at least use existing solution like https://github.com/StackExchange/dnscontrol instead of rolling out our own that only supports cloudflare and is not used by anyone else.

this is what i'm using, i'm working on a hook that calls dnscontrol to push the records via api for common providers. if this is not preferred we may be able to make py infra do the work, but i'd need to research it a bit more since it's newer framework for me

link2xt:
> we may be able to make py infra do the work

pyinfra is mostly about SSHing into the server and installing stuff there. For DNS we only need to do things locally, like generating a zonefile or dnsconfig.js for dnscontrol.

ccclxxiii:
> > we may be able to make py infra do the work
> 
> pyinfra is mostly about SSHing into the server and installing stuff there. For DNS we only need to do things locally, like generating a zonefile or dnsconfig.js for dnscontrol.

unless this is already started, i'll likely pick it up and experiment this week if time allows

link2xt:
Let's close this PR since in the current state it is not finished and likely not the approach we want to take.

PR #695, opened 2025-10-26
Organize cmdeploy into install, configure, and activate stages

cliffmccarthy:
- Added a class hierarchy of Deployer objects which implement install(), configure(), and activate() methods for each of the components that comprise a chatmail server.
- This is intended to provide the same functionality as the current process, though the order of the individual steps has been changed to organize them into the three separate stages of deployment.
- For more information, see the individual commit messages.  The commits are intended to be individually reviewable logical steps in the transformation of the code.  Also see the discussion of the new code structure in ARCHITECTURE.md.
- So far, I have tested this code on fresh installs and some upgrade scenarios, but I have not tested every possible upgrade path from past releases of chatmail/relay.

### DIFF
--- a/doc/source/overview.rst
+++ b/doc/source/overview.rst
@@ -342,28 +342,3 @@ of building a general-purpose container image.
 Operations that start services for systemd-based deployments should
 only be called from the ``activate_impl()`` methods.  These methods
 will not be called in non-systemd container environments.
-
-Deployer objects
-~~~~~~~~~~~~~~~~
-
-One might ask why the deployers are implemented as object classes, as
-opposed to callable functions or the like.  There are various reasons
-why objects are a good fit for the deployment process.
-
-1. Objects provide a way to organize the install, configure, and
-deploy operations for each component that is installed, supporting a
-"driver" type of pattern.  This could be implemented in other ways
-without objects, such as function jump tables, but objects provide a
-clean and formalized way to do essentially the same thing.
-
-2. Class inheritance provides a natural way to define
-component-specific operations for the various stages of deployment, by
-overriding the no-op stub methods in the base class.  The base class
-handles policy decisions about which stages are to be executed,
-ensuring consistent handling of the stages in a central location.
-
-3. Some of the components track state between stages, basing decisions
-like whether to restart a service on whether the software or
-configuration of that service was changed in an earlier stage.
-Keeping track of state between method calls is an ideal use case for
-objects.


cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -507,168 +754,348 @@ def check_config(config):
...
+        files.put(
+            name="Add Deltachat OBS GPG key to apt keyring",
+            src=importlib.resources.files(__package__).joinpath(
+                "obs-home-deltachat.gpg"
+            ),
+            dest="/etc/apt/keyrings/obs-home-deltachat.gpg",
+            user="root",
+            group="root",
+            mode="644",
+        )
```
missytake: This can be completely removed as we don't install dovecot from OBS anymore :) We don't need to set `present=False` imo because the file doesn't hurt either.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -766,161 +1141,81 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
...
+        acmetool_deployer,
+        website_deployer,
+        chatmail_venv_deployer,
+        mtasts_deployer,
+        opendkim_deployer,
+        dovecot_deployer,
+        postfix_deployer,
+        nginx_deployer,
+        rspamd_deployer,
+        fcgiwrap_deployer,
```
missytake:
```suggestion
        fcgiwrap_deployer,
        nginx_deployer,
        rspamd_deployer,
```

I think we should activate fcgiwrap before nginx actually, otherwise there might be a short time where nginx routes to the fcgiwrap port but nothing listens.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -766,161 +1141,81 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
...
+    echobot_deployer = EchobotDeployer(
+        dovecot_deployer=dovecot_deployer, postfix_deployer=postfix_deployer
+    )
+    journald_deployer = JournaldDeployer()
+    mtail_deployer = MtailDeployer(mtail_address=config.mtail_address)
+
+    all_deployers = [
+        chatmail_deployer,
+        turn_deployer,
+        unbound_deployer,
```
missytake:
```suggestion
        chatmail_deployer,
        unbound_deployer,
        turn_deployer,
```


My head finds it more logical to do unbound before turn deployment, as DNS resolving is more system preparation.

Having turn being deployed next to iroh-relay makes sense as well.

I don't see a technical necessity to do it either way though.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -766,161 +1141,81 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
...
+        chatmail_venv_deployer,
+        mtasts_deployer,
+        opendkim_deployer,
+        dovecot_deployer,
+        postfix_deployer,
+        nginx_deployer,
+        rspamd_deployer,
+        fcgiwrap_deployer,
+        echobot_deployer,
+        journald_deployer,
```
missytake: I'm a bit surprised journald is activated so late in the process - isn't logging important to the entire process, too? I think it should be directly before or after `chatmail_deployer`.

ARCHITECTURE.md
```diff
@@ -1,3 +1,5 @@
+## Chatmail Server
+
 This diagram shows components of the chatmail server; this is a draft
```
missytake:
```suggestion
# Chatmail Relay

This diagram shows components of the chatmail relay server; this is a draft
```

Let's use chatmail relay or chatmail relay server ;) it does in some situations make sense to say "chatmail relay *server*" to clarify that it runs on servers; In general we like to say "relay" instead of "server" to point out that it doesn't really store messages and only relays them.

missytake (approved):
Love it! It makes the whole structure much clearer, and the commit structure made it very pleasant to review.

I have some small suggestions, none of them critical.

I like how it stays inconsistent whether a Deployer class is definied in `cmdeploy/__init__.py` or in a subdirectory (with acmetool). I think many of the Deployers could be in subdirectories, because they are pretty isolated, like turn, iroh, opendkim, journald, and mtail. For some it might make sense to have them in `__init__.py` because you can see at one glance whether they interact with other services (like acmetool, actually?).

I started the CI in https://github.com/chatmail/relay/pull/696, so we can notice errors, just in case.
We should probably have more than just my review for this, as it changes the structure deeply? @hpk42 @link2xt

missytake:
CI fails to download iroh-relay on the ipv4-only runner o.0 this did not happen before. Any ideas?

```
--> Starting operation: Download iroh-relay 
    [staging-ipv4.testrun.org]   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
    [staging-ipv4.testrun.org]                                  Dload  Upload   Total   Spent    Left  Speed
    [staging-ipv4.testrun.org] 
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (6) Could not resolve host: github.com
    [staging-ipv4.testrun.org] 
    [staging-ipv4.testrun.org] gzip: stdin: unexpected end of file
    [staging-ipv4.testrun.org] tar: This does not look like a tar archive
    [staging-ipv4.testrun.org] tar: ./iroh-relay: Not found in archive
    [staging-ipv4.testrun.org] tar: Exiting with failure status due to previous errors
    [staging-ipv4.testrun.org] Error: executed 0 commands
```

Maybe unbound needs to be activated before the installation step? This would mess with the general concept of course.

cliffmccarthy:
> CI fails to download iroh-relay on the ipv4-only runner o.0 this did not happen before. Any ideas?

The "Download chatmail-turn" step runs almost the exact same thing and it worked in that run.  The only thing that happened between that step and "Download iroh-relay" was installation of unbound, so that does suggest that unbound has something to do with it.  I wonder if we can do the same thing I did with nginx to prevent the service from starting up as soon as it is installed.

missytake:
Ah, then unbound is out of the box simply misconfigured for an ipv4 only host?

cliffmccarthy:
> Ah, then unbound is out of the box simply misconfigured for an ipv4 only host?

That seems possible, given what we have observed here.  I haven't investigated how it works very deeply.

I made the revision to use policy-rc.d like with nginx, and it works fine on a fresh install.  You can cherry-pick commit 32b0472c57d199a86c71ecdcf851ee6554eed1dc onto the CI pull request to see if that solves it for the IPv4 system.

missytake:
> > Ah, then unbound is out of the box simply misconfigured for an ipv4 only host?
> 
> That seems possible, given what we have observed here. I haven't investigated how it works very deeply.
> 
> I made the revision to use policy-rc.d like with nginx, and it works fine on a fresh install. You can cherry-pick commit [32b0472](https://github.com/chatmail/relay/commit/32b0472c57d199a86c71ecdcf851ee6554eed1dc) onto the CI pull request to see if that solves it for the IPv4 system.

that helped. But do we want unbound to be disabled during installation, or do we want it correctly configured?

cliffmccarthy:
> But do we want unbound to be disabled during installation, or do we want it correctly configured?

During the install and configure phases, we mostly install software and write files.  Can we do that without unbound running?  Certainly the initial installation of unbound itself is done without unbound running.  And this sequence of operations only affects the first 'cmdeploy run' invocation on a fresh install; for subsequent calls to 'run' (such as to apply changes or toggle the mail service), the unbound service should be already up and running.

cmdeploy/src/cmdeploy/__init__.py
hpk42: generally, i'd say it's better to remove files that we don't need anymore.  It's sometimes not easy to determine on the filesystem what is happening if there are a lot of garbage files. For example, in /etc/dovecot btw there are several unused files which is confusing if you want to investigate live configuration.

cliffmccarthy: I revised this to remove the file instead of creating it.  This is a no-op in a container image build, of course, but for the upgrade use case it will help as you indicate.

cmdeploy/src/cmdeploy/__init__.py
cliffmccarthy: I moved this to just after chatmail_deployer, and made the other reorderings you suggested.  chatmail_deployer does the 'apt' updates, so I think of it as needing to be the first thing in the sequence, though JournaldDeployer doesn't do any package operations, so I suppose it could probably be first if necessary.  But, the service doesn't get reset until the activate stage, so the order it appears in the install phase may not matter.

ARCHITECTURE.md
cliffmccarthy: Revised as suggested.

cliffmccarthy:
I pushed a few more commits to implement the feedback so far.  I also added a changelog entry, which is now a conflict, so I'll rebase when we think this is good to go and resolve the conflict.  (And I'll add the "conventional commits" tags that I forgot on the first pass.)

Right now, this has no functional conflicts with the main branch, but if there are any pending pull requests that this would conflict with, I'm fine with merging those first if desired to minimize disruption, and I can incorporate the revisions when I do the final rebase.

cliffmccarthy:
I rebased onto main and force-pushed to get this pull request up-to-date.  I retained the previous state of the pull request in my fork on the [cmdeploy-install-configure-activate-2025-10-30](https://github.com/cliffmccarthy/chatmail-relay/tree/cmdeploy-install-configure-activate-2025-10-30) branch, if anyone needs to refer back to it for any reason.

cliffmccarthy:
The latest revision includes an additional change I made after examining ChatmailVenvDeployer.  I noticed it was handling systemd.service() calls for doveauth, echobot, and turnserver, which are associated with other deployer classes.  For echobot and turnserver, the systemd.service() calls were duplicated in the corresponding classes, possibly restarting the service twice in each run of cmdeploy.  I refactored the code so that ChatmailVenvDeployer only handles services delivered by the chatmail relay project, and the restarting of doveauth, echobot, and turnserver is handled in the corresponding classes.  This also led to a simplification of those classes because they no longer need to keep state in an attempt to avoid restarting.  For more details, see the commit messages and diffs in the four commits leading up to 948efff70db3013f99339f0989d75d063cdc479d.

ARCHITECTURE.md
```diff
@@ -48,3 +50,66 @@ graph LR;
...
+
+One might ask why the deployers are implemented as object classes, as
+opposed to callable functions or the like.  There are various reasons
+why objects are a good fit for the deployment process.
+
+1. Objects provide a way to organize the install, configure, and
+deploy operations for each component that is installed, supporting a
+"driver" type of pattern.  This could be implemented in other ways
+without objects, such as function jump tables, but objects provide a
+clean and formalized way to do essentially the same thing.
```
hpk42: Modules would also provide a way to organize install, configure and deploy operations. 
You wouldn't need jump tables, but can also check via introspection if a "cmdeploy_configure", "cmdeploy_install", etc. is present. I don't object to using classes for arranging code.

ARCHITECTURE.md
```diff
@@ -48,3 +50,66 @@ graph LR;
...
+
+2. Class inheritance provides a natural way to define
+component-specific operations for the various stages of deployment, by
+overriding the no-op stub methods in the base class.  The base class
+handles policy decisions about which stages are to be executed,
+ensuring consistent handling of the stages in a central location.
```
hpk42: Are there real world use cases for inheritance?  Not in the current code at least, as far as i can see. 
n.b. I've seen systems become very complicated where people thought subclassing is a good extensibility idea at first ("nose" python testing framework got into intricate problems, limiting extensibility).

ARCHITECTURE.md
```diff
@@ -48,3 +50,66 @@ graph LR;
...
+
+3. Some of the components track state between stages, basing decisions
+like whether to restart a service on whether the software or
+configuration of that service was changed in an earlier stage.
+Keeping track of state between method calls is an ideal use case for
+objects.
```
hpk42: Are you talking about "need_restart"? Is there anything else? 

If need be, the state could also be passed as an object around to the deployer functions. 

On some level, modules with functions are like classes, except you can't instantiate multiple copies. Are there deployer classes that would exist more than once within one run? 

That all being said, i am fine to stick with classes for now.

cmdeploy/src/cmdeploy/deployer.py
```diff
@@ -0,0 +1,80 @@
+import os
+
+from pyinfra.operations import server
+
+
+class Deployer:
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
```
hpk42: this call is not neccessary, as there is no base class.

cmdeploy/src/cmdeploy/deployer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    #
+    # In any override, this method should return a list of 3-element
+    # (user, group, secondary-group-list) tuples.  If the group is None,
+    # no group is created corresponding to that user.  If the secondary
+    # group list is not None, the listed groups are created as well.
+    #
+    @staticmethod
+    def required_users():
+        return []
```
hpk42: i've seen no evidence that it can't just be a class attribute. Making it a static method with no input means that it could only depend on global state which is something to avoid.

cmdeploy/src/cmdeploy/__init__.py
```diff
@@ -766,163 +1118,79 @@ def deploy_chatmail(config_path: Path, disable_mail: bool) -> None:
...
+        mtasts_deployer,
+        opendkim_deployer,
+        dovecot_deployer,
+        postfix_deployer,
+        fcgiwrap_deployer,
+        nginx_deployer,
+        rspamd_deployer,
+        echobot_deployer,
+        mtail_deployer,
+    ]
```
hpk42, line 1127: you can inline all the deployer instantiations, halfing the number of lines. Not neccessary to create all these tmp variables.

ARCHITECTURE.md
cliffmccarthy: Yes, too much inheritance certainly leads to trouble.  Here, it seemed like a good way to provide default stub implementations, with classes only overriding the parts they need to.

ARCHITECTURE.md
cliffmccarthy: Yes, need_restart is the main stateful thing I saw between the stages of deployment.  (In the case of AcmetoolDeployer, this will become three state variables once I rebase onto the latest main.)

cmdeploy/src/cmdeploy/deployer.py
cliffmccarthy: I was going for "cooperative design" as described in https://rhettinger.wordpress.com/2011/05/26/super-considered-super/.  Following that pattern from the beginning is easier than trying to add it later.  But as discussed above, maybe we won't ever have enough inheritance for it to matter.  For the sake of simplicity, I'm fine with dropping these super() calls on the assumption there won't be any additional args or inherited constructors.

cmdeploy/src/cmdeploy/deployer.py
cliffmccarthy: Making it an attribute is a good idea.  I like knowing my data is just data and can't suddenly come to life and do something weird like connect to the internet.

cmdeploy/src/cmdeploy/__init__.py
cliffmccarthy: This is a great suggestion.  Before I made the simplification to the activate step for echobot, it needed to be able to access the postfix and dovecot deployers to check their state, but now that it operates independently of them, we don't need to keep the references for later, so I like your idea of doing it all inline and reducing the duplication.

cliffmccarthy:
Thanks for the great comments and revisions, @hpk42!  I'll merge your pull request into my branch soon, after I've read through all the diffs.  Then I'll need to rebase onto main to resolve the acmetool conflict.  I'll keep your commits in the stack for proper attribution, of course.

cliffmccarthy:
I rebased onto main again and force-pushed to get this pull request up-to-date.  I retained the previous state of the pull request in my fork on the [cmdeploy-install-configure-activate-2025-11-11](https://github.com/cliffmccarthy/chatmail-relay/tree/cmdeploy-install-configure-activate-2025-11-11) branch.  That branch includes the original commits from @hpk42 exactly as they were submitted.

cmdeploy/src/cmdeploy/deployer.py
hpk42: it all comes down to the question of using class inheritance as an organizing paradigm -- which i am forever doubtful about. 

FWIW currently Deployer objects have install/configure/activate methods, and can pass state like "need_restart" via  "self".  Inheritance can only change one of the three actions, and call their super() if they want. I'd like to see code that uses this base inheritance and then discuss further.  

Maybe it will be entirely different aspects we want to make customizable? For pytest, i co-developed the small [pluggy](https://pypi.org/project/pluggy/) plugin mechanism that allows to define "hook" functions and call them easily in defined protocols. There are now thousands of plugins, and many plugins from ten years ago continue to function fine without much changes. Pytest uses pluggy to allow deep customization of test collection, running, reporting and debugging, including adding new command line options, new env vars, new modes of running things etc.   See https://github.com/pytest-dev/pytest/blob/main/src/_pytest/hookspec.py#L75 for the current "hookspec".

doc/source/overview.rst
```diff
@@ -297,3 +297,73 @@ actually it is a problem with your TLS certificate.
...
+component-specific operations for the various stages of deployment, by
+overriding the no-op stub methods in the base class.  The base class
+handles policy decisions about which stages are to be executed,
+ensuring consistent handling of the stages in a central location.
+
+3. Some of the components track state between stages, basing decisions
+like whether to restart a service on whether the software or
+configuration of that service was changed in an earlier stage.
+Keeping track of state between method calls is an ideal use case for
+objects.
```
hpk42: i suggest to strike this section completely.  It's partly not fitting anymore, partly not agreeable, and not really adding much to what was said in the previous section.  
```suggestion

```

cliffmccarthy: I was starting to question the value of it myself.  I pushed a new revision.

cliffmccarthy:
For CI, we were using https://github.com/chatmail/relay/pull/696 previously; perhaps @missytake can update that with the latest to get another run that has the necessary permissions.

hpk42 (approved):
wonderful, i think this is mergeable. Thanks @cliffmccarthy for persevering!

Issue #431, opened 2024-10-17
Remove `/metrics`


link2xt:
This is a follow-up to #328 

We now have `mtail` metrics which are more flexible and can also count new registrations and similar metrics.

`/metrics` is created by a cronjob and is not authenticated, so we cannot expose much there and "pushing" metrics is not really compatible with prometheus model where it expects to pull metrics at any time and timestamps the data with the current timestamp.

For metrics like disk space admins can run node_exporter.

PR #703, opened 2025-10-29
feat: remove /metrics from the website

link2xt:
Similar data is already generated by fsreport
available for the relay operator
and metrics for prometheus are generated by mtail.

Closes <https://github.com/chatmail/relay/issues/431>


cliffmccarthy (commented):
This looks good to me, and appeals to my "negative lines of code" sense.

ccclxxiii:
@link2xt this was a point of confusion for some relay operators, glad it's being factored away.  the mtail works quite well as it is.

feld:
fsreport doesn't tell you how many accounts, only login stats. But it would be easy to update fsreport to add it.

link2xt:
> this was a point of confusion for some relay operators, glad it's being factored away

This was brought up on the forum again: https://support.delta.chat/t/userbase-information/4567/4

So rebasing it.

feld:
in the spirit of reducing unnecessary external metadata, let's just do this if there aren't any blockers

j4n:
I just stumbled across this today; but it did not seem the metrics where equivalent to the ones from mtail:
```
# HELP total number of accounts
# TYPE accounts gauge
accounts 12830
# HELP number of CI accounts
# TYPE ci_accounts gauge
ci_accounts 0
# HELP number of non-CI accounts
# TYPE nonci_accounts gauge
nonci_accounts 12830
```
vs mtail
```
# HELP created_accounts defined at delivered_mail.mtail:17:9-24
# TYPE created_accounts counterd_mail.mtail"} 41444
created_accounts{prog="delivered_mail.mtail"} 41444l.mtail:18:9-27
# HELP created_ci_accounts defined at delivered_mail.mtail:18:9-27
# TYPE created_ci_accounts counterd_mail.mtail"} 0
created_ci_accounts{prog="delivered_mail.mtail"} 0_mail.mtail:19:9-30
# HELP created_nonci_accounts defined at delivered_mail.mtail:19:9-30
# TYPE created_nonci_accounts counterd_mail.mtail"} 41444
created_nonci_accounts{prog="delivered_mail.mtail"} 41444-22
# HELP delivered_mail defined at delivered_mail.mtail:1:9-22
```
(and there seems to be something weird going on with that mtail output)

Maybe we should just output the metrics to `/var/lib/prometheus/node-exporter` instead of `/var/www/`?

link2xt:
Don't know why your output looks like this, something is probably wrong with the terminal. `created_accounts` metric is a counter since the start of mtail as it gets the info from logs, so you can use it to see how many accounts are created in some timespan by recording this in prometheus.

j4n:
I understand, mtail is providing the rate of account creation; I think it would be helpful to be able to monitor total accounts on the relay continuing removal of the public /metric. 

Incorporating it into fsreport would be sensible I suppose, ideally with openmetrics-compatible output. If I understand correctly, fsreport is already running by systemd-timer regularly?

Re output, I indeed can't reproduce this issue atm.

hpk42:
So far, /metrics allows to get an outside impression how many addresses are on a server. 

CLI/systemd fsreport run only gives it to admin. 

I skimmed the chain of referenced issues, but i didn't see a clear rationale 
why we don't want the number of addresses to be publically visible outside.

link2xt:
Why do we want to expose the number of addresses to everyone on the internet? What is the usecase for knowing the number of addresses on servers that are not run by you?

j4n:
I just pushed #881 which would allow us to remove metrics.py and still provide monitoring output for those who want it.

missytake:
> Why do we want to expose the number of addresses to everyone on the internet? What is the usecase for knowing the number of addresses on servers that are not run by you?

by now, I agree :)

j4n:
rebased, and I think we can merge this, the only question is if we should amend the systemd-unit for fsreport to output the metrics file (#881 now merged) to keep the user count public.

PR #706, opened 2025-10-30
Improve dns responses parsing

lig:
`dig` often includes failed attempts in fulfilling the user request with lines that start with ';' which is a commentary line. This lines could be safely ignored as long as there is an actual answer.

Requests like an example below could result in some connectivity (ipv6 or otherwise) errors reports followed by a legitimate answer from dig.

This change ignores such "comments" in the output and returns the first meaningful answer.

```shell
> dig @linode3.countzero.co -r -q mta-sts.countzero.co -t cname +short
;; communications error to 2a01:7e01::f03c:93ff:febc:6635#53: connection refused
;; communications error to 2a01:7e01::f03c:93ff:febc:6635#53: connection refused
;; communications error to 2a01:7e01::f03c:93ff:febc:6635#53: connection refused
countzero.co.
```

This case appears to be caused by Debian 12 having no ipv6 connectivity by default on a fresh install. So, any SOA accessible by ipv6 will most likely cause a break in the chatmail server default installation process.

### DIFF
--- a/cmdeploy/src/cmdeploy/tests/test_dns.py
+++ b/cmdeploy/src/cmdeploy/tests/test_dns.py
@@ -1,3 +1,5 @@
+from copy import deepcopy
+
 import pytest
 
 from cmdeploy import remote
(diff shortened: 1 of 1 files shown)


missytake:
Looks good! Can you maybe add a test to `cmdeploy/src/cmdeploy/tests/test_dns.py` with a mocked DNS test result with a comment? Otherwise I can do that later myself.

lig:
Will do

lig:
That will be a bit more complicated than adding a fixture as the method affected by the fix is actually mocked in tests. I'll figure something out.

lig:
@missytake I've parametrized the fixture and changed the function that is being mocked.

PR #707, opened 2025-10-31
Remove smtp_tls_security_level exception for nauta.cu

link2xt:
nauta.cu MX servers don't support STARTTLS
and we have maintained this exception to allow delivery to nauta.cu. Since nauta.cu has blocked largest chatmail relay instances, we can remove this exception.
Even if it works today for some chatmail servers,
they will likely be blocked once the number of users grows. Even now users from nauta.cu cannot practically participate in groups.

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/main.cf.j2
+++ b/cmdeploy/src/cmdeploy/postfix/main.cf.j2
@@ -25,7 +25,6 @@ smtp_tls_security_level=verify
 # <https://www.postfix.org/postconf.5.html#smtp_tls_servername>
 smtp_tls_servername = hostname
 smtp_tls_session_cache_database = btree:${data_directory}/smtp_scache
-smtp_tls_policy_maps = inline:{nauta.cu=may}
 smtp_tls_protocols = >=TLSv1.2
 smtpd_tls_protocols = >=TLSv1.2
 


link2xt:
I am not sure if there are nauta.cu users of bots that can migrate from one bot to another in case of blocks. nauta.cu users connecting to bots is still an usecase to support, then we can just close it.

missytake:
What are the downsides of keeping the exception? I don't see how it hurts.

link2xt:
> What are the downsides of keeping the exception? I don't see how it hurts.

We will have to keep it forever then even if nobody is using it because maybe someone depends on it. Currently mails from nauta.cu to nine.testrun.org work one way, mails sent from nine.testrun.org to nauta.cu simply disappear. Yet all the addresses participating in the group are sent in plaintext and visible by the hoster of chatmail server as soon as any nauta.cu address is added instead of clearly rejecting the mail.

If we remove it and someone notices that it does not work and complains we can add it back, but currently it seems nobody is using it and it is a good time to remove this.

@adbenitez said the bots that are actually used are running on gmail or so.

missytake:
I think it's fine if we keep one line in a config file forever.

The situation in groups with both nine and nauta.cu users will not get less annoying if we remove this line.

But - the situation will improve *a lot* with multi-transport. Users will simply use a different transport if nine doesn't work, e.g. one of the 30+ chatmail relays which are not blocked. If we remove this line, we also break it for all those other relays which will soon make it work again.

adbenitez:
this should at least be changed for nine that is blocked already, and exposed to chatmail config so operator can disable this, ex. also arcanechat.me is blocked so no need for this hack

there are some few poor people still using nauta.cu as a cheap way of getting stuff from the internet, this puts us in bad light with the ISP/Cuban-gov that considers such usage illegal activity / bypassing (but that bridge is burned anyways), that is why they more recently blocked arcanechat.me and other chatmails that were shared in the "Cuban DeltaChat" Telegram public room,

missytake:
> this should at least be changed for nine that is blocked already, and exposed to chatmail config so operator can disable this, ex. also arcanechat.me is blocked so no need for this hack
> 
> there are some few poor people still using nauta.cu as a cheap way of getting stuff from the internet, this puts us in bad light with the ISP/Cuban-gov that considers such usage illegal activity / bypassing (but that bridge is burned anyways), that is why they more recently blocked arcanechat.me and other chatmails that were shared in the "Cuban DeltaChat" Telegram public room,

Including the line for some chatmail relays but removing it for nine and arcanechat.me makes the maintenance even harder. Then it's not just one line we need to remember, but additional templating based on the `mail_domain`... then we can just leave it in for blocked hosts, it will simply have no effect.

PR #708, opened 2025-10-31
Remove acmetool cronjob

link2xt:
acmetool is always running in redirector mode and has a hook that restarts or reload services, there is no need to do the same manually.

### DIFF
--- a/cmdeploy/src/cmdeploy/acmetool/__init__.py
+++ b/cmdeploy/src/cmdeploy/acmetool/__init__.py
@@ -10,12 +10,9 @@ def deploy_acmetool(email="", domains=[]):
         packages=["acmetool"],
     )
 
-    files.put(
-        src=importlib.resources.files(__package__).joinpath("acmetool.cron").open("rb"),
-        dest="/etc/cron.d/acmetool",
-        user="root",
-        group="root",
-        mode="644",
+    files.file(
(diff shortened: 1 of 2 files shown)


link2xt:
This was added in 9c09d50e8feb52b725cc8b10665a3f6055bd79b9 when hook was already in place.

missytake:
> This was added in [9c09d50](https://github.com/chatmail/relay/commit/9c09d50e8feb52b725cc8b10665a3f6055bd79b9) when hook was already in place.

It was added because the hook didn't work :( in any case, it would be much nicer to have a systemd timer executing this. Then we can investigate when a cert job fails, and see whether it actually ran. I suggest to do this instead?

link2xt:
> It was added because the hook didn't work :( in any case, it would be much nicer to have a systemd timer executing this. Then we can investigate when a cert job fails, and see whether it actually ran. I suggest to do this instead?

The hook is there, if it does not work then it should have been fixed instead. Why do we hack our own solution with cron or timers when acmetool is supposed to be doing this?

I see the hook is installed in `/usr/lib/acme/hooks/nginx`, but the man page says it should go to `/etc/acme/hooks`.

link2xt:
It seems acmetool-redirector is not renewing certificates, so we need a timer.
I opened #719 instead.

PR #710, opened 2025-10-31
Fix quotes in page-layout.html

alxndr42:
I'm guessing that web browsers are flexible enough, but my text editor noticed. :sweat_smile:

### DIFF
--- a/www/src/page-layout.html
+++ b/www/src/page-layout.html
@@ -9,7 +9,7 @@
     <title>{{ config.mail_domain }} {{ pagename }}</title>
     <link rel="stylesheet" href="./main.css">
     <link rel="icon" href="/logo.svg">
-    <link rel=”mask-icon” href=”/logo.svg” color=”#000000">
+    <link rel="mask-icon" href="/logo.svg" color="#000000">
 </head>
 <body>
 


missytake:
Yeaah, we get a PR for this every once in a while... problem is, this will cause merge conflicts for many people, and that is not really worth fixing a cosmetic problem - say sorry to your IDE ;)

Theoretically, since a while you can set `www_folder=disabled` [in chatmail.ini](https://github.com/chatmail/relay/?tab=readme-ov-file#custom-web-pages). We could start detecting if `<<<<<<<<<`, `==========`, and `>>>>>>>>>>` are in the `www/src/*.md` files, and recommend to resolve the merge conflicts or set `www_folder=disabled` to skip www upload. This would be in any case a good change, as merge conflicts can also happen for other reasons, and e.g. https://mehl.cloud/info.html has a conflict deployed as we speak. Can you maybe add that to this PR? I understand if that's out of your scope :sweat_smile:

alxndr42:
> We could start detecting if <<<<<<<<<, ==========, and >>>>>>>>>> are in the www/src/*.md files, and recommend to resolve the merge conflicts or set www_folder=disabled to skip www upload.

Would this be done as part of the pyinfra deployment? I can try to look into that.

(Personally, I'm already using a separate location for `www_folder`, I just wanted to fix the quotes for anyone else who runs into this. Preferring to avoid a ton of merge conflicts is understandable, though.)

alxndr42:
It might be a bit confusing to repurpose this PR for merge conflict detection, so I created a separate issue for it: https://github.com/chatmail/relay/issues/713

PR #719, opened 2025-11-06
Replace acmetool cronjob with a timer

link2xt:
There is no && restart in the systemd unit, hook should restart the services.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- acmetool: replace cronjob with a systemd timer
+  ([#719](https://github.com/chatmail/relay/pull/719))
+
 - don't deploy the website if there are merge conflicts in the www folder
   ([#714](https://github.com/chatmail/relay/pull/714))
 


cmdeploy/src/cmdeploy/acmetool/__init__.py
```diff
@@ -10,21 +10,24 @@ def deploy_acmetool(email="", domains=[]):
...
     )
 
-    files.put(
-        src=importlib.resources.files(__package__).joinpath("acmetool.cron").open("rb"),
-        dest="/etc/cron.d/acmetool",
-        user="root",
-        group="root",
-        mode="644",
+    # Remove old acmetool cronjob, it is replaced with systemd timer.
+    files.file(
```
missytake:
```suggestion
    files.file(
        name="Remove old acmetool cronjob, it is replaced with systemd timer.",
```

This serves both as a comment and shows better output in pyinfra :)

cmdeploy/src/cmdeploy/acmetool/__init__.py
```diff
@@ -10,21 +10,24 @@ def deploy_acmetool(email="", domains=[]):
...
     files.put(
         src=importlib.resources.files(__package__).joinpath("acmetool.hook").open("rb"),
-        dest="/usr/lib/acme/hooks/nginx",
+        dest="/etc/acme/hooks/nginx",
         user="root",
         group="root",
         mode="744",
     )
+    # Remove the hook from the wrong location where it was previously installed.
+    files.file(
```
missytake:
```suggestion
    files.file(
        name="Remove the hook from the wrong location where it was previously installed."
```

same here :)

cmdeploy/src/cmdeploy/acmetool/acmetool-reconcile.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Renew TLS certificates with acmetool
+After=network.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/acmetool --batch reconcile
+
```
missytake, line 8:
```suggestion
Restart=always
RestartSec=86400

```

We could also drop the timer by just adding restarting the service once a day, right?
That's how we do it in https://codeberg.org/webxdc/pyinfra-xstore/src/branch/main/pyinfra_xstore/xstore-update.service.j2, could save us one file.

link2xt: Then it will not show up in `systemctl list-timers`.

Huge `RestartSec` values look like an ugly hack to me, basically scheduling tasks with large sleep values. This means it will not be run at exact time of the day, but will drift over time. For acmetool may not matter, but generally if you do your scheduling this way you will get unwanted synchronization over time, services will tend to run at the same time if you run the server long enough.

missytake: Fair :)

cmdeploy/src/cmdeploy/acmetool/acmetool.cron
```diff
@@ -1,4 +0,0 @@
-SHELL=/bin/sh
-PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin
-MAILTO=root
-20 16 * * * root /usr/bin/acmetool --batch reconcile && systemctl reload dovecot && systemctl reload postfix && systemctl reload nginx
```
hpk42, line 4: where does the former `reload dovecot && reload postfix && reload nginx` happen in the systemd version?

link2xt: It is in the hook, see the PR description.

PR #720, opened 2025-11-08
ci: pin jsok/serialize-workflow-action

link2xt:


link2xt:
It is running code from third-party repo that can be changed at any time.

See https://docs.zizmor.sh/audits/#unpinned-uses

PR #721, opened 2025-11-08
ci: delete old files when backing up acmetool state and DKIM keys

link2xt:
Took me a while to delete google.com and gogle.com requests that were restored over and over to merge https://github.com/chatmail/relay/pull/719

I also don't understand why rsync errors are ignored with `|| true` everywhere.

### DIFF
--- a/.github/workflows/test-and-deploy-ipv4only.yaml
+++ b/.github/workflows/test-and-deploy-ipv4only.yaml
@@ -35,11 +35,11 @@ jobs:
           chmod 600 ~/.ssh/id_ed25519
           ssh-keyscan staging-ipv4.testrun.org > ~/.ssh/known_hosts
           # save previous acme & dkim state
-          rsync -avz root@staging-ipv4.testrun.org:/var/lib/acme acme-ipv4 || true
-          rsync -avz root@staging-ipv4.testrun.org:/etc/dkimkeys dkimkeys-ipv4 || true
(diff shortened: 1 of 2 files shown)


missytake:
> Took me a while to delete google.com and gogle.com requests that were restored over and over to merge #719

Sorry for messing up our CI's acmetool state -.- I kinda forgot that it is kept, and failed to correctly clean up on ns.testrun.org after that.

> I also don't understand why rsync errors are ignored with `|| true` everywhere.

Because if a CI run fails before it acquires a /var/lib/acme and/or /etc/dkimkeys directory, it will make the next CI run fail; what we want in such a case is failing silently, so that later the last successful state from ns.testrun.org can be applied to the test VPS.

Also later we want to fail silently, without blocking the CI run - because if using the CI state which is cached on ns.testrun.org fails for some reason, we don't want the CI to stop, we want the CI run to create a new /var/lib/acme state, which is then later cached on ns.testrun.org during the next CI run.

What we could do - move the steps you modified to the very end of the CI workflow, so it only stores /var/lib/acme and /etc/dkimkeys on ns.testrun.org if the tests completed successfully. Then we could also use `rsync --delete` and give up on `|| true`.

In the beginning I developed it like this, because we didn't have state on ns.testrun.org yet; now that we have, we can move the steps to the end.

hpk42 (approved):
lgtm

hpk42:
Hum, i am not sure i am totally getting the problem that missytake outlined, so my approval of the PR might be premature.

PR #723, opened 2025-11-10
cmdeploy: fix status cmd after sshexec rework

missytake:
Broke with https://github.com/chatmail/relay/pull/659/files, reported in the mutual help chat.

tbh I think that the status command makes little sense as it is right now, not reporting failed services, only active ones... we could also remove it imho, or put 30 minutes of effort into making it useful.


hpk42:
the direct problem was that "cmdeploy status" had no automated test.  that's independent from improving what it does.  This PR should add a test.

PR #726, opened 2025-11-13
fix and update mermaid overview graph

hpk42:
- remove "cmdeploy" from relay diagram 
- replace cron with systemd and add edges
- be more detailed about filtermail-incoming versus outgoing 
- color internet reachable services red, color all filesystem-locations yellow
- fix postfix to use dovecot for SASL, instead of the wrong depicted direct use of doveauth
- add lastlogin component 
- add notification proxy

### DIFF
--- a/doc/source/overview.rst
+++ b/doc/source/overview.rst
@@ -171,7 +171,7 @@ Chatmail relay dependency diagram
         chatmail-expire-daily --- maildir;
         chatmail-fsreport-daily --- maildir;
         chatmail-metadata --- iroh-relay;
-        chatmail-metadata --- |encrypted device token| notification-proxy;
+        chatmail-metadata --- |encrypted device token| notifications.delta.chat;
         certs-nginx --> postfix;
         certs-nginx --> dovecot;
         style certs fill:#ff6;


doc/source/overview.rst
```diff
@@ -147,32 +146,43 @@ Component dependency diagram
...
+        dovecot --- |message delivery|maildir["maildir
+        /home/vmail/.../user"];
+        dovecot --- |lastlogin.socket|lastlogin;
+        dovecot --- chatmail-metadata;
+        lastlogin --- maildir;
+        doveauth --- maildir;
+        chatmail-expire-daily --- maildir;
+        chatmail-fsreport-daily --- maildir;
         chatmail-metadata --- iroh-relay;
+        chatmail-metadata --- |encrypted device token| notification-proxy;
```
missytake:
```suggestion
        chatmail-metadata --- |encrypted device token| notifications.delta.chat;
```

Maybe call it with its domain name to make clearer that it's an external service?

doc/source/overview.rst
```diff
@@ -147,32 +146,43 @@ Component dependency diagram
...
+        style website fill:#ff6;
+        style maildir fill:#ff6;
         style certs-nginx fill:#ff6;
-        style nginx-external fill:#fc9;
-        style nginx-right fill:#fc9;
+        style nginx-external fill:#f66;
+        style nginx-right fill:#f66;
+        style postfix fill:#f66;
+        style dovecot fill:#f66;
+        style notification-proxy fill:#f66;
```
missytake, line 185: There is not really a legend saying what the colors mean... can we put that in the caption or so?

doc/source/overview.rst
```diff
@@ -126,14 +126,13 @@ web page. Edit them before deploying to make your chatmail relay
 stand out.
 
 
-Component dependency diagram
---------------------------------------
+Chaitmail relay dependency diagram
+-------------------------------------------
```
missytake:
```suggestion
Chatmail relay dependency diagram
---------------------------------
```

Typo and too long underline ;)

doc/source/overview.rst
hpk42: did that, but the caption is then strangely formatted. It's not critical for understanding the diagram.

doc/source/overview.rst
hpk42: fixed

PR #729, opened 2025-11-14
acmetool: accept new Let's Encrypt Terms of Services

missytake:
Not sure this is suddenly necessary... a bit weird that we only noticed it now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- acmetool: accept new Let's Encrypt ToS: https://letsencrypt.org/documents/LE-SA-v1.6-August-18-2025.pdf
+  ([#729](https://github.com/chatmail/relay/pull/729))
+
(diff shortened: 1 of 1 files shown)


cmdeploy/src/cmdeploy/acmetool/response-file.yaml.j2
```diff
@@ -1,2 +1,2 @@
 "acme-enter-email": "{{ email }}"
-"acme-agreement:https://letsencrypt.org/documents/LE-SA-v1.5-February-24-2025.pdf": true
+"acme-agreement:acme-agreement:https://letsencrypt.org/documents/LE-SA-v1.6-August-18-2025.pdf": true
```
link2xt: Why `acme-agreement:` twice, copy-past error?

missytake: eeer, yes, 100%.

PR #730, opened 2025-11-15
Fixups for TLS configuration

link2xt:
There are two commit messages with a description, read them separately.

This mandatory/not mandatory distinction in postfix is a footgun, it has these settings but when they are actually used is described in a different place and not in the documentation for them.


hpk42:
soiunds good, but i'd have preferred to have a comment in the postfix config instead of in the commit.

PR #732, opened 2025-11-16
Add robots.txt to exclude all web crawlers

### LINKED ISSUE #715: Add a robots.txt to block searching for relays
A Google search with some phrases from the chatmail privacy policy, for instance, "This chatmail server neither asks for nor retains personal information" gives at least 36 results leading to websites of different chatmail relays. 

Not all of these relays were intended to be discovered in this way. This could lead to a faster and more efficient blocking of relays by a sophisticated censor (a.k.a Russia) 

Could be good to add a robots.txt or equivalent way to block collection in this way.

link2xt:
Close #715 

Contents of the file is copy-pasted from https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Robots_txt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## untagged
 
+- Add robots.txt to exclude all web crawlers
+  ([#732](https://github.com/chatmail/relay/pull/732))
+
 - acmetool: accept new Let's Encrypt ToS: https://letsencrypt.org/documents/LE-SA-v1.6-August-18-2025.pdf
   ([#729](https://github.com/chatmail/relay/pull/729))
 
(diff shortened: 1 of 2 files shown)


link2xt:
Seems to work, https://staging2.testrun.org/robots.txt is deployed.

missytake (approved):
Maybe a changelog item? Some might want their relay to be listed in search engines.

link2xt:
> Maybe a changelog item?

I'm adding changelog before merging because otherwise there are conflicts all the time.

I think we should make a release finally and switch to git-cliff afterwards, this manual changelog is a mess, someone adds to the top, someone to the bottom and now we have a huge unsorted "unreleased" section.

Issue #739, opened 2025-11-22
BUG: setup fails


maikelthedev:
- Server OS (Operating System) - preferably Debian 12: 
- On which OS you run cmdeploy: Debian12
- chatmail/relay version: 8f68672e3112682055b8fdb6e0e8d6c473730f8c

## Expected behavior

Chatmail server installs

## Actual behavior

Install fails

<img width="1262" height="721" alt="Image" src="https://github.com/user-attachments/assets/dfa068d1-503d-49f7-8ef0-e4eea5addab0" />

### Steps to reproduce the problem:

Literally follow https://chatmail.at/doc/relay/getting_started.html to the small print including server version. Exact same version of the server. The output given is the result of `scripts/cmdeploy run`

link2xt:
Duplicate of #666

maikelthedev:
Starting afresh with another Debian 12 install....

```bash
apt update && apt install python3-venv gcc git python3-dev libunwind8 libexttextcat-2.0-0  liblua5.4-0 liblua5.4-0  libstemmer0d libexttextcat-data 
```

Succcess!!

<img width="730" height="341" alt="Image" src="https://github.com/user-attachments/assets/f4e9b510-81b2-4016-9ca7-57b2aa806c45" />

So all the files needed are those from a blank install, not just libunwind8, and installing libunwind8 from a broken install as #666 suggested does not fix the issue, only installing all of the above on a **fresh** install

PR #740, opened 2025-11-22
Update getting_started.rst to reflect minimum requirements

maikelthedev:
As explained here in https://github.com/chatmail/relay/issues/739#issuecomment-3566815930 setup instructions won't work unless those packages are installed.


maikelthedev:
Will make part of bigger PR, so closing

Issue #641, opened 2025-09-18
CI: Printing echobot invite link after cmdeploy run sometimes fails


missytake:
- Server OS (Operating System) - preferably Debian 12: Debian 12
- On which OS you run cmdeploy: github CI
- chatmail/relay version: https://github.com/chatmail/relay/actions/runs/17808076868/job/50624784672?pr=639

Since https://github.com/chatmail/relay/pull/620 we print the echobot link at the end of CI. It seems that sometimes `echobot.service` hasn't started yet at this point, and  /var/lib/echobot/invite-link.txt doesn't exist yet. This makes the CI flaky.

missytake:
Until postfix + dovecot are enabled, echobot fails and tries to restart every 30 seconds. If `cmdeploy run` takes less than 30 seconds between starting postfix + dovecot and finishing, there is the possibility that echobot hasn't started yet, and the file doesn't exist. Then CI fails.

PR #741, opened 2025-11-22
Fix issues with echobot invite link in cmdeploy

mrdaemon:
The entire stack is setup to support `localhost` as a value for ssh_host, returning `LocalExec` when that is the case, but this last step made a new explicit `SSHExec` connection to `config.mail_domain`.

This changes that to simply use the `subprocess` module when `ssh_host` is local.

This also fixes the issue where the connection was made to `config.mail_domain` instead of the supplied `ssh_host` value, ensuring that remains consistent.

Additionally, the entire process will be skipped if `--dry-run` is used with cmdeploy, allowing a dry-run to complete without error.

Absolutely looking for feedback, as I mostly fixed this to get a deployment running in my environment, if more is needed, or a different approach is preferable, I'd be glad to rework this.

### DIFF
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -109,15 +109,24 @@ def run_cmd(args, out):
     try:
         retcode = out.check_call(cmd, env=env)
         if retcode == 0:
-            if not args.disable_mail:
+            if not args.disable_mail and not args.dry_run:
                 print("\nYou can try out the relay by talking to this echo bot: ")
-                sshexec = SSHExec(args.config.mail_domain, verbose=args.verbose)
-                print(
-                    sshexec(
-                        call=remote.rshell.shell,
-                        kwargs=dict(command="cat /var/lib/echobot/invite-link.txt"),
+                echobot_cmd = "cat /var/lib/echobot/invite-link.txt"
+                if ssh_host in ["localhost", "@local", "@docker"]:
+                    result = (
+                        subprocess.check_output(echobot_cmd, shell=True)
+                        .decode()
+                        .strip()
                     )
-                )
+                    print(result)
+                else:
+                    echo_sshexec = get_sshexec(ssh_host, verbose=args.verbose)
+                    print(
+                        echo_sshexec(
+                            call=remote.rshell.shell,
+                            kwargs=dict(command=echobot_cmd),
+                        )
+                    )
             out.green("Deploy completed, call `cmdeploy dns` next.")
         elif not remote_data["acme_account_url"]:
             out.red("Deploy completed but letsencrypt not configured")


hpk42 (commented):
lgtm, although it'd probably be good to revisit all the places where the split local/remote execution takes place, and see if the if-branching can be avoided.

mrdaemon:
I originally tried this but LocalExec and SSHExec don't seem to have the same semantics, with LocalExec only having a 'logged()' method, so I left it alone.

I wouldn't mind adding the required methods to LocalExec so it could just be used in place of SSHExec and avoiding the crappy if/else branching, if that sounds ok

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -109,15 +109,24 @@ def run_cmd(args, out):
...
+                        .strip()
+                    )
+                    print(result)
+                else:
+                    echo_sshexec = get_sshexec(ssh_host, verbose=args.verbose)
+                    print(
+                        echo_sshexec(
+                            call=remote.rshell.shell,
+                            kwargs=dict(command=echobot_cmd),
+                        )
```
missytake, line 128:
```suggestion
                invite_path = "/var/lib/echobot/invite-link.txt"
                if ssh_host in ["localhost", "@local", "@docker"]:
                    with open(invite_path) as f:
                        print(f.readline())
                else:
                    echo_sshexec = get_sshexec(ssh_host, verbose=args.verbose)
                    print(
                        echo_sshexec(
                            call=remote.rshell.shell,
                            kwargs=dict(command=f"cat {invite_path}"),
                        )
```

Thanks for the fix, just stumbled upon it ourselves. We don't even have to open a subprocess if we're on the same machine :) let's just do it in python.

j4n: This is actually too fast now and requires a time.sleep (5 seconds works for me) before the file exists. (Somewhat related to #641). 
```suggestion
                invite_path = Path("/var/lib/echobot/invite-link.txt")
                if ssh_host in ["localhost", "@local", "@docker"]:
                    sleep(5) # wait a bit for echobot to settle down
                    with invite_path.open() as f:
                        print(f.readline())
                else:
                    sshexec = SSHExec(args.config.mail_domain, verbose=args.verbose)
                    print(
                        localexec(
                            call=remote.rshell.shell,
                            kwargs=dict(command=f"cat {invite_path}"),
                        )
                    )
```

(still needs the `from time import sleep` at the top)

hpk42: `sleep(5)` is a questionable hack to avoid.  Why is it that "cmdeploy run" finishes but echobot is not running?  Is it possible to write a final "deployer" that waits e.g. for echobot to be up?  Not sure if that's easily possible with pyinfra (or a systemctl command).

hpk42: maybe "systemctl is-active --wait echobot.service" or something similar could ensure that "cmdeploy run" returns only after echobot runs?

missytake: `systemctl is-active --wait echobot.service` will only wait if state is `initializing` or `starting`, and we have either `failed` or `active` with `active` not necessarily meaning that configuration worked :/ I'll try something with `Requires=`: https://github.com/chatmail/relay/pull/746

Maybe we should move this question out of this PR.

missytake: I opened https://github.com/chatmail/relay/pull/751 so we can merge this improvement :)

PR #742, opened 2025-11-22
refactor: clean up, docs and Nixos compatibility

maikelthedev:
This commit does multiple things:
1. Puts Makefile on the project folder where it belongs as it can be reused for other stuff too.
2. Changes "make auto" to "make docs"
3. Make Nixos a possible development environment by the use of shell.nix
4. Avoids having to SSH anywhere to create a file in /etc/ to stop registrations, instead uses allow_registrations in chatmail.ini (see amended docs).
5. Updates the docs to work in both a blank Debian 12 server install and with Nixos as client.
6. Gets rid of not being nine.testrun.org being a reason to be declared Dev server.

### DIFF
--- a/doc/source/getting_started.rst
+++ b/doc/source/getting_started.rst
@@ -97,10 +97,16 @@ steps. Please substitute it with your own domain.
 
        scripts/cmdeploy run
 
-   This script will also check that you have all necessary DNS records.
+   This script will also check that you have most necessary DNS records.
    If DNS records are missing, it will recommend which you should
    configure at your DNS provider (it can take some time until they are
    public).
+
+6. It is highly recommended you run this post-setup to ensure **all needed** DNS records are there.
+   ::
+
+       scripts/cmdeploy dns
+
 
 Other helpful commands
 ----------------------


chatmaild/src/chatmaild/doveauth.py
```diff
@@ -23,8 +21,8 @@ def encrypt_password(password: str):
 
 def is_allowed_to_create(config: Config, user, cleartext_password) -> bool:
     """Return True if user and password are admissable."""
-    if os.path.exists(NOCREATE_FILE):
-        logging.warning(f"blocked account creation because {NOCREATE_FILE!r} exists.")
+    if config.allow_registrations != "yes":
+        logging.warning(f"blocked account creation because allow_registrations in chatmail.ini is not 'yes'.")
```
hpk42, line 25: this actually changes a run-time switch to a build-time .ini file switch,
and is backward compatible (for two years, `/etc/chatmail-nocreate` worked, and it might be in use in some closed instances).    So i doubt it's worth it.

doc/source/getting_started.rst
```diff
@@ -27,6 +29,28 @@ You will need the following:
    key is required due to an `upstream bug in
    paramiko <https://github.com/paramiko/paramiko/issues/2191>`_)
 
+Required dependencies for smooth install
+----------------------------------------
+
+On a Debian 12 cloud server
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
hpk42, line 37: this sounds like being about the target server, but it's really the client side, right?

maikelthedev:
Agreed

PR #746, opened 2025-12-01
echobot: require dovecot to be running

missytake:
The changes in https://github.com/chatmail/relay/pull/741/files#r2564775943 re-introduce #641. This might help - it worked for me on docker.testrun.org.


missytake:
I'm unsure whether this starts dovecot too early, though. With this PR, if you stop both dovecot and echobot, and you start echobot, it starts dovecot as well.

PR #747, opened 2025-12-03
Remove DKIM-Signature from incoming mail after checking 

j4n:
The original https://github.com/chatmail/relay/pull/533 attempted to remove the header through postfix, but that is too early. Instead, remove the headers in the OpenDKIM `final.lua` script after the validation.

### DIFF
--- a/cmdeploy/src/cmdeploy/opendkim/final.lua
+++ b/cmdeploy/src/cmdeploy/opendkim/final.lua
@@ -24,16 +24,15 @@ for i = 1, nsigs do
 	end
 end
 
-if not valid then
+if valid then
+	-- Strip all DKIM-Signature headers after successful validation
+	-- Delete in reverse order to avoid index shifting.
+	for i = nsigs, 1, -1 do
+		odkim.del_header(ctx, "DKIM-Signature", i)
+	end
+else
 	odkim.set_reply(ctx, "554", "5.7.1", "No valid DKIM signature found")
 	odkim.set_result(ctx, SMFIS_REJECT)
-	return nil
-end
-
--- Valid signature found. Strip all DKIM-Signature headers
--- Delete in reverse order to avoid index shifting.
-for i = nsigs, 1, -1 do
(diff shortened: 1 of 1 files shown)


doc/source/overview.rst
```diff
@@ -273,9 +273,11 @@ Incoming emails must have a valid DKIM signature with
 Signing Domain Identifier (SDID, ``d=`` parameter in the DKIM-Signature
 header) equal to the ``From:`` header domain. This property is checked
 by OpenDKIM screen policy script before validating the signatures. This
-correpsonds to strict :rfc:`DMARC <7489>` alignment (``adkim=s``).
+corresponds to strict :rfc:`DMARC <7489>` alignment (``adkim=s``).
 If there is no valid DKIM signature on the incoming email, the
 sender receives a “5.7.1 No valid DKIM signature found” error.
+After validating the DKIM signature, OpenDKIM strips all ``OpenDKIM:`` headers to
+reduce message size.
```
hpk42:
```suggestion
After validating the DKIM signature, 
the `final.lua` script strips all ``OpenDKIM:`` headers to reduce message size on disc. 
```

cmdeploy/src/cmdeploy/opendkim/final.lua
```diff
@@ -9,20 +9,31 @@ if nsigs == nil then
...
+	odkim.set_reply(ctx, "554", "5.7.1", "No valid DKIM signature found")
+	odkim.set_result(ctx, SMFIS_REJECT)
+	return nil
+end
+
+-- Valid signature found. Strip all DKIM-Signature headers
+-- Delete in reverse order to avoid index shifting.
+for i = nsigs, 1, -1 do
+	odkim.del_header(ctx, "DKIM-Signature", i)
 end
```
hpk42: this would read clearer to me (i don't know much lua)  if it were a "if valid-else" construct and at the end a single "return nil".  given there is no automated tests any gain in readability is good :)

j4n:
Implemented your feedback.

PR #751, opened 2025-12-06
echobot: print invite link without SSH if deployed locally/with docker

missytake:
superseding #741 - it would still be nice to wait until the file is created for the SSH code path, but then again it takes longer to establish, so the chance is higher that it just works.


missytake:
CI is broken because of https://github.com/chatmail/core/issues/7574.

PR #753, opened 2025-12-07
remove echobot from relay deployment 

hpk42:
fixes #752  
closes #751 
closes #746 
closes #741

also makes sure echobot is removed during "cmdeploy run"

image

PR #755, opened 2025-12-08
feat: Remove additional echobot-related code

cliffmccarthy:
feat: Remove echobot special cases

feat: Remove echobot user from deployment

feat: Remove echo from passthrough recipients

Also see #753.


cliffmccarthy:
Closing to re-submit from main repo so CI can run.

PR #756, opened 2025-12-08
feat: Remove additional echobot-related code

cliffmccarthy:
feat: Remove echobot special cases

feat: Remove echobot user from deployment

feat: Remove echo from passthrough recipients

Also see https://github.com/chatmail/relay/pull/753.


cliffmccarthy:
If any of these are still needed for any legacy compatibility reasons, let me know and I'll put them back.  (Along with a comment so we don't remove it again.)

adbenitez:
I think this is needed before tagging a new release since otherwise there will be leftovers after the removal of echobot, cc @missytake

hpk42 (approved):
thanks a lot, cliff, for cleaning up after my quick remove-echobot PR 😅

PR #757, opened 2025-12-08
dovecot: add tmpfs_index chatmail.ini parameter for storing index files in /dev/shm

### LINKED ISSUE #644: Restore in-memory index

missytake:
fix #644

### DIFF
--- a/chatmaild/src/chatmaild/ini/override-testrun.ini
+++ b/chatmaild/src/chatmaild/ini/override-testrun.ini
@@ -2,8 +2,6 @@
 
 tmpfs_index = true
 
-[privacy]
-
 passthrough_recipients = privacy@testrun.org echo@{mail_domain}
 
 privacy_postal =
--- a/chatmaild/src/chatmaild/config.py
+++ b/chatmaild/src/chatmaild/config.py
@@ -112,10 +112,10 @@ def get_default_config_content(mail_domain, **overrides):
 
     if mail_domain.endswith(".testrun.org"):
         override_inipath = inidir.joinpath("override-testrun.ini")
-        privacy = iniconfig.IniConfig(override_inipath)["privacy"]
+        params = iniconfig.IniConfig(override_inipath)["params"]
         lines = []
         for line in content.split("\n"):
-            for key, value in privacy.items():
+            for key, value in params.items():
                 value_lines = value.format(mail_domain=mail_domain).strip().split("\n")
                 if not line.startswith(f"{key} =") or not value_lines:
                     continue


missytake:
Maybe we can just use https://linuxvox.com/blog/devshm-in-linux/ instead of shipping our own tmpfs.

hpk42 (commented):
dovecot allows to set a separate index directory: https://doc.dovecot.org/2.3/configuration_manual/mail_location/#index-files which is preferable to creating symlinks 

how is dovecot handling it when the (tmp) file system gets full?

chatmaild/src/chatmaild/ini/override-testrun.ini
```diff
@@ -1,3 +1,6 @@
+[params]
+
+tmpfs_index = true
```
missytake, line 3: This is somehow not set for staging*.testrun.org.

cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
```diff
@@ -68,13 +68,11 @@ userdb {
...
+{% if config.tmpfs_index %}
+mail_location = maildir:{{ config.mailboxes_dir }}/%u:INDEX=/dev/shm/%u
+{% else %}
 mail_location = maildir:{{ config.mailboxes_dir }}/%u
-
-# index/cache files are not very useful for chatmail relay operations 
-# but it's not clear how to disable them completely. 
-# According to https://doc.dovecot.org/2.3/settings/advanced/#core_setting-mail_cache_max_size
-# if the cache file becomes larger than the specified size, it is truncated by dovecot 
-mail_cache_max_size = 500K
```
j4n, line 77: removing this limit will lead to no limit when we dont use the tmpfs option I presume?

j4n:
> how is dovecot handling it when the (tmp) file system gets full?

This is a good question, along with the usual combined size of the caches on a large server.
 FWIW, on docker shm defaults to 64 MiB, so that would need an override parameter exposed.

chatmaild/src/chatmaild/expire.py
```diff
@@ -145,6 +150,9 @@ def process_mailbox_stat(self, mbox):
             changed = True
         if changed:
             self.remove_file(f"{mbox.basedir}/maildirsize")
+        for file in mbox.extrafiles:
+            if "dovecot.index" in file.path.split("/")[-1] and file.size > 500 * 1024:
+                self.remove_file(file.path)
```
j4n, line 155: This matches broadly and according to my brief research triggers a full index rebuild where the old max_size flag just pruned the cache, I wonder if this won't mean a lot of I/O churn.

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -48,12 +48,8 @@
 # (space-separated, item may start with "@" to whitelist whole recipient domains)
 passthrough_recipients =
 
-# Use externally managed TLS certificates instead of built-in acmetool.
-# Paths refer to files on the deployment server (not the build machine).
-# Both files must already exist before running cmdeploy.
-# Certificate renewal is your responsibility; changed files are
-# picked up automatically by all relay services.
-# tls_external_cert_and_key = /path/to/fullchain.pem /path/to/privkey.pem
+# store index files in tmpfs (good for disk size and I/O, bad for ram)
```
j4n, line 51: This should be more informative I think. What are the performance implications / motivations to enable this? How big should your ram / tmpfs wrt. to number of chatmail accounts?

cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
missytake: the default limit is 1G.

chatmaild/src/chatmaild/expire.py
missytake: Yes, thinking about it again, we should only remove dovecot.index.cache files.

missytake: Not only that, on second thought it's the only useful thing in this PR. I'll close this one and create a new 3-line PR.

Issue #762, opened 2025-12-09
Can't message other Chatmail relays due to DKIM error


17451k:
- Server OS: Debian 12
- On which OS you run cmdeploy: Debian 12
- chatmail/relay version: b000213c

## Issue

Each message to `nine.testrun.org` (and other chatmail servers) returns with `5.7.1 No valid DKIM signature found`, messaging works only inside my chatmail server.

I assume that if `scripts/cmdeploy dns` returns "Great! All your DNS entries are verified and correct." then DKIM should work as expected. Am I wrong here?

### Logs

Message info from DeltaChat client:
```
Sent: 2025.12.09 16:28:54 by Me (REDACTED)
State: Failed, Encrypted
Error: Undelivered Mail Returned to Sender – This is the mail system at host REDACTED.

I'm sorry to have to inform you that your message could not
be delivered to one or more recipients. It's attached below.

For further assistance, please send mail to postmaster.

If you do so, please include this problem report. You can
delete your own text from the attached returned message.

The mail system

<5hp347qko@c2.testrun.org>: host c2.testrun.org[95.216.165.93] said: 500 Error:
(SMTPDataError) (554, b'5.7.1 No valid DKIM signature found') (in reply to
end of DATA command)
Message-ID: 9308c159-8a9b-4d58-a3e3-0039ab99d0c5@localhost
Server-URL: </INBOX/;UID=29855>

No Hop Info
```

scripts/cmdeploy dns
```
scripts/cmdeploy dns
Collecting initial DNS settings..............
Check expected zone file entries...........................................
Great! All your DNS entries are verified and correct.
```

scripts/cmdeploy test
```
...
SKIPPED [1] cmdeploy/src/cmdeploy/tests/online/test_1_basic.py:127: set CHATMAIL_DOMAIN2 to a second chatmail server
SKIPPED [2] cmdeploy/src/cmdeploy/tests/plugin.py:34: skipping slow test, use --slow to run
SKIPPED [1] cmdeploy/src/cmdeploy/tests/online/test_2_deltachat.py:110: set CHATMAIL_DOMAIN2 to a second chatmail server
SKIPPED [1] cmdeploy/src/cmdeploy/tests/online/test_2_deltachat.py:124: set CHATMAIL_DOMAIN2 to a second chatmail server
47 passed, 5 skipped in 111.35s

```

scripts/cmdeploy status
```
chatmail domain: REDACTED
no privacy settings
[...]

PR #764, opened 2025-12-10
opendkim: report DKIM error code in SMTP response

missytake:
This was very useful in debugging #762, does anything speak against making the rejection reason transparent?

### DIFF
--- a/cmdeploy/src/cmdeploy/opendkim/final.lua
+++ b/cmdeploy/src/cmdeploy/opendkim/final.lua
@@ -10,7 +10,7 @@ if nsigs == nil then
 end
 
 local valid = false
-local sigerrors = ""
+local error_msg = "No valid DKIM signature found."
 for i = 1, nsigs do
 	sig = odkim.get_sighandle(ctx, i - 1)
 	sigres = odkim.sig_result(sig)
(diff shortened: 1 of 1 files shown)


cmdeploy/src/cmdeploy/opendkim/final.lua
```diff
@@ -31,7 +34,7 @@ if valid then
 		odkim.del_header(ctx, "DKIM-Signature", i)
 	end
 else
-	odkim.set_reply(ctx, "554", "5.7.1", "No valid DKIM signature found")
+	odkim.set_reply(ctx, "554", "5.7.1", "No valid DKIM signature found. Search https://github.com/trusteddomainproject/OpenDKIM/blob/master/libopendkim/dkim.h#L108 for " .. sigerrors)
```
link2xt: `sigerrors` will be empty if there are no signatures at all.

cmdeploy/src/cmdeploy/opendkim/final.lua
```diff
@@ -21,6 +22,8 @@ for i = 1, nsigs do
 	-- means the message is acceptable.
 	if sigres == 0 then
 		valid = true
+    else
+        sigerrors = sigerrors .. " " .. tostring(sigres)
```
link2xt: I think it's better to create a variable for the error message at the top, set it to `No valid DKIM signature found`. Then on this line replace it with `DKIM signature is invalid, error code 4, search https://github.com/trusteddomainproject/OpenDKIM/blob/master/libopendkim/dkim.h#L108`.

And do not concatenate the error codes, just replace the error. It is unlikely that the sender has multiple DKIM signatures. I just don't trust the ability of OpenDKIM handling long error messages, it is C code that is trivial to segfault from Lua and I don't want to think about what happens if someone sends in a message with 10000 invalid signatures.

PR #765, opened 2025-12-10
postfix + dovecot: restart automatically on failure

### LINKED ISSUE #731: Dovecot is not restarted if service dies
on VPS with low RAM (ex. mine with 4GB) eventually filtermail and opendkim can use a lot of RAM, I suspect this was the reason Dovecot service got killed/died, in any case, this caused hours of downtime because it coincidentally happened while I was sleeping, if dovecot service would be configured to auto-restart, like filtermail service is, there would have been little downtime, since after filtermail got also killed and restarted the memory usage went down again

TL;DR: please set:

```
Restart=always
RestartSec=30
```

for dovecot service same way as it is done for filtermail service so if service dies unexpectedly the server can recover without big downtimes until admin notice or is available

missytake:
fix #731

### DIFF
--- a/cmdeploy/src/cmdeploy/dovecot/deployer.py
+++ b/cmdeploy/src/cmdeploy/dovecot/deployer.py
@@ -13,6 +13,8 @@ from cmdeploy.basedeploy import (
 
 
 class DovecotDeployer(Deployer):
+    daemon_reload = False
+
     def __init__(self, config, disable_mail):
         self.config = config
         self.disable_mail = disable_mail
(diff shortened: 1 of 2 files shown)


cmdeploy/src/cmdeploy/dovecot/deployer.py
```diff
@@ -27,7 +27,7 @@ def install(self):
 
     def configure(self):
         configure_remote_units(self.config.mail_domain, self.units)
-        self.need_restart = _configure_dovecot(self.config)
+        self.need_restart, self.daemon_reload = _configure_dovecot(self.config)
```
cliffmccarthy, line 32: I recommend assigning an initial value to `daemon_reload` in the constructor, so that the value exists even if activate() is called without calling configure() first.  (This applies to the postfix file too.)

hpk42: i add a default of None to instances.

PR #766, opened 2025-12-10
ci: run test_status_cmd at the end to avoid flakiness

missytake:
Let's see if this resolves errors like in https://github.com/chatmail/relay/actions/runs/20057002693/job/57540276823?pr=760 - it seem chatmail-metadata just takes a little bit longer to startup than the other units, and is not there yet when CI runs `test_status_cmd`.


missytake:
So we remove the test? Right now it's the reason tests are failing on main.

PR #768, opened 2025-12-11
feat: add --website-only option to run subcommand

### LINKED ISSUE #467: cmdeploy run --webdev command to only deploy www
It would be nice for the web development workflow if one could just run `cmdeploy run --webdev` and deploy only the www pages without the whole rest of the server.

adbenitez:
close #467


j4n:
Looks good to me!

PR #775, opened 2025-12-13
decoms development instance notice

ccclxxiii:
pr overview:
- in chats we discussed having operators remove the development instance notice
- some operators ignore the development instance banner even after their relay is stable
- this pr removes the notice, we can discuss pros/cons but i feel it doesn't serve a purpose


hpk42 (approved):
lgtm

PR #778, opened 2025-12-15
Make filtermail use normal 127.0.0.1 as the source address

feld:
On some OSes in you can use any address in the 127.0.0.0/8 range and it will work fine, but FreeBSD won't allow it.

### DIFF
--- a/chatmaild/src/chatmaild/filtermail.py
+++ b/chatmaild/src/chatmaild/filtermail.py
@@ -312,7 +312,7 @@ class IncomingBeforeQueueHandler:
         client = SMTPClient(
             "localhost",
             self.config.postfix_reinject_port_incoming,
-            source_address=("127.0.0.2", 0),
+            source_address=("127.0.0.1", 0),
         )
         client.sendmail(
             envelope.mail_from, envelope.rcpt_tos, envelope.original_content


chatmaild/src/chatmaild/filtermail.py
```diff
@@ -312,7 +312,7 @@ def sync_handle_DATA(self, envelope):
         client = SMTPClient(
             "localhost",
             self.config.postfix_reinject_port_incoming,
-            source_address=("127.0.0.2", 0),
+            source_address=("127.0.0.1", 0),
```
link2xt, line 315: See comment above, the reason for using 127.0.0.2 is to make opendkim verify DKIM signature. For messages that come from 127.0.0.1 it signs the message instead.

hpk42:
@feld how did you manage to get his freebsd relay running (with working cmping with other hosts, i assume)  regarding OpenDKIM ?

link2xt:
> @feld how did you manage to get his freebsd relay running (with working cmping with other hosts, i assume) regarding OpenDKIM ?

Relay probably signs incoming messages instead of verifying the signature if this change is applied.

feld:
My incoming messages all show

```
DKIM Results: Passed=true
```

However, that may be due to running in a traditional FreeBSD jail without a full network stack (full network stack jails are VNET jails). In this scenario 127.0.0.1 gets remapped automatically to the IP of the jail, which would be something other than 127.0.0.1 when OpenDKIM sees it.

That's likely what's going on, but I need to investigate further.

link2xt:
DKIM passes because you still sign outgoing mails. But now for incoming mails you also sign them instead of verifying them.

To make this less error-prone, one way would be to set `InternalHosts` to empty list (need to check if this actually disables signing) and then use `MacroList` or `MTA` configuration to make OpenDKIM sign instead of verifying where needed: http://www.opendkim.org/opendkim.conf.5.html

link2xt:
Let's close this, this change fails the test: https://github.com/chatmail/relay/pull/783
I opened https://github.com/chatmail/relay/pull/784 using macros instead.

PR #780, opened 2025-12-16
docs: update migration guide after nine migration

missytake:
Maybe this should be tested one more time, as the nine migration didn't go as smoothly as expected. Comments appreciated.

### DIFF
--- a/doc/source/migrate.rst
+++ b/doc/source/migrate.rst
@@ -26,7 +26,7 @@ this case, just run ``ssh-keygen -R "mail.example.org"`` as recommended.
    ::
 
        ssh -A root@$OLD_IP4
-       tar c - /home/vmail/mail | ssh root@$NEW_IP4 "tar x -C /"
+       tar c /home/vmail/mail | ssh root@$NEW_IP4 "tar x -C /"
 
    This saves us time during the downtime,
    at least the mailboxes are there already.
@@ -62,10 +62,10 @@ this case, just run ``ssh-keygen -R "mail.example.org"`` as recommended.
    ::
 
        ssh -A root@$OLD_IP4
-       tar c - /var/lib/acme /etc/dkimkeys /var/spool/postfix | ssh root@$NEW_IP4 "tar x -C /"
+       tar c /var/lib/acme /etc/dkimkeys /var/spool/postfix | ssh root@$NEW_IP4 "tar x -C /"
        rsync -azH /home/vmail/mail root@$NEW_IP4:/home/vmail/
 
-   This transfers all addresses, messages which have not been fetched yet, the TLS certificate,
+   This transfers all messages which have not been fetched yet, the TLS certificate,
    and DKIM keys (so DKIM DNS record remains valid).
    It also preserves the Postfix mail spool so any messages
    pending delivery will still be delivered.


doc/source/migrate.rst
```diff
@@ -8,65 +8,86 @@ you might need to adjust some of the steps to your environment.
 
 Let’s assume that your ``mail_domain`` is ``mail.example.org``, all
 involved machines run Debian 12, your old site’s IP address is
-``13.37.13.37``, and your new site’s IP address is ``13.12.23.42``.
```
j4n, line 11: too l33t for you :)? More seriously, I think it was useful to have an ssh-alias for cm-new and cm-old, then we dont have the IP addresses in the ssh invocations which probably makes it a whole lot clearer whats going on in times of panic.

doc/source/migrate.rst
```diff
@@ -8,65 +8,86 @@ you might need to adjust some of the steps to your environment.
...
+   let's transfer the current state of the mailboxes.
+   Login to your old machine (while forwarding your ssh-agent with ``ssh -A``)
+   so you can copy directly from the old to the new site with your SSH
+   key:
 
    ::
 
-       cmdeploy run --disable-mail --ssh-host 13.37.13.37
+       ssh -A root@13.12.13.12
+       tar c - /home/vmail/mail | ssh root@45.54.45.54 "tar x -C /"
```
j4n: did you test that invocation? when I tried, it did not work..
```
tar c - test/ | ssh localhost "tar x -C /tmp"
tar: -: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```

j4n: also maybe comment that this will transfer many a thousand files so verbose output might slow things down considerably. importantly, this should replicate hardlinks. probably useful to offer the rsync alternative here as well already.

doc/source/migrate.rst
```diff
@@ -8,65 +8,86 @@ you might need to adjust some of the steps to your environment.
...
+   You only need to change the ``A`` and ``AAAA`` records, for example:
+
+   ::
 
-   If other MTAs try to deliver messages to your chatmail domain they
-   may fail intermittently, as DNS catches up with the new site settings
-   but normally will retry delivering messages for at least a week, so
-   messages will not be lost.
+       mail.example.org.    IN A    45.54.45.54
+       mail.example.org.    IN AAAA 45:ac:1312:ab::1
```
j4n: ok, we should really start the guide with recommending to lower the TTL of these records ahead of the migration (60s) and raise it back up here.

j4n (approved):

doc/source/migrate.rst
hpk42: i replaced ip addresses with `$OLD_IP4` and `$NEW_IP4` 
you can literally set the environment variables, should work.

doc/source/migrate.rst
hpk42: this recommendation is now at the top.

doc/source/migrate.rst
hpk42: fixed the tar command but not sure it's needed to discuss the in's and out's of doing verbose mode, which we don't suggest anyway.

hpk42:
i continued working in #789 on revamping migration doc.  Would suggest to close this one here.

PR #783, opened 2025-12-18
feat: do not use 127.0.0.2 for OpenDKIM

link2xt:
Just want to see if CI catches this or if we need a new test.


link2xt:
This failed `test_reject_missing_dkim` as expected.

PR #784, opened 2025-12-18
feat: use daemon_name for OpenDKIM sign-verify decision instead of IP

link2xt:
On FreeBSD 127.0.0.2 is not assigned to any interface by default,
so 127.0.0.2 source address hack cannot be used to make OpenDKIM
verify the signature instead of signing.

This change sets InternalHosts to `-` so no IP addresses
make OpenDKIM sign the message. Instead of IP address,
OpenDKIM in the outgoing pipeline is explicitly told
to sign messages by setting `{daemon_name}` macro to `ORIGINATING`.

This is a replacement for #778

### DIFF
--- a/cmdeploy/src/cmdeploy/opendkim/opendkim.conf
+++ b/cmdeploy/src/cmdeploy/opendkim/opendkim.conf
@@ -65,3 +65,9 @@ PidFile			/run/opendkim/opendkim.pid
 # The trust anchor enables DNSSEC. In Debian, the trust anchor file is provided
 # by the package dns-root-data.
 TrustAnchorFile		/usr/share/dns/root.key
+
+# Sign messages when `-o milter_macro_daemon_name=ORIGINATING` is set.
+MTA ORIGINATING
+
+# No hosts are treated as internal, ORIGINATING daemon name should be set explicitly.
+InternalHosts -
--- a/cmdeploy/src/cmdeploy/postfix/master.cf.j2
+++ b/cmdeploy/src/cmdeploy/postfix/master.cf.j2
@@ -31,7 +31,6 @@ submission inet n       -       y       -       5000    smtpd
   -o smtpd_sender_restrictions=$mua_sender_restrictions
   -o smtpd_recipient_restrictions=
   -o smtpd_relay_restrictions=permit_sasl_authenticated,reject
-  -o milter_macro_daemon_name=ORIGINATING
   -o smtpd_client_connection_count_limit=1000
   -o smtpd_proxy_filter=127.0.0.1:{{ config.filtermail_smtp_port }}
 smtps     inet  n       -       y       -       5000    smtpd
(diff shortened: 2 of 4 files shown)


link2xt:
If this does not fail `test_reject_missing_dkim` then should be an alternative usable on FreeBSD.

I tested in https://github.com/chatmail/relay/pull/783 that the test works.

cmdeploy/src/cmdeploy/opendkim/opendkim.conf
```diff
@@ -65,3 +65,9 @@ PidFile			/run/opendkim/opendkim.pid
 # The trust anchor enables DNSSEC. In Debian, the trust anchor file is provided
 # by the package dns-root-data.
 TrustAnchorFile		/usr/share/dns/root.key
+
+# Sign messages when `-o milter_macro_daemon_name=ORIGINATING` is set.
+MTA ORIGINATING
+
+# No hosts are treated as internal, ORIGINATING daemon name should be set explicitly.
+InternalHosts -
```
link2xt, line 73: This `-` is used in some OpenDKIM tests:
https://github.com/trusteddomainproject/OpenDKIM/blob/master/opendkim/tests/t-sign-ss-macro.conf#L9

cmdeploy/src/cmdeploy/postfix/master.cf.j2
```diff
@@ -31,7 +31,6 @@ submission inet n       -       y       -       5000    smtpd
   -o smtpd_sender_restrictions=$mua_sender_restrictions
   -o smtpd_recipient_restrictions=
   -o smtpd_relay_restrictions=permit_sasl_authenticated,reject
-  -o milter_macro_daemon_name=ORIGINATING
```
link2xt, line 34: This is unused because there is no milter.

cmdeploy/src/cmdeploy/postfix/master.cf.j2
```diff
@@ -49,7 +48,6 @@ smtps     inet  n       -       y       -       5000    smtpd
   -o smtpd_recipient_restrictions=
   -o smtpd_relay_restrictions=permit_sasl_authenticated,reject
   -o smtpd_client_connection_count_limit=1000
-  -o milter_macro_daemon_name=ORIGINATING
```
link2xt, line 52: This was also unused, it's just a cleanup.

link2xt:
This is supposed to work: http://www.trusteddomain.org/pipermail/opendmarc-users/2013-June/000153.html

Here is the source code where `{daemon_name}` macro should be checked against MTA set and set `originok` to true:
https://github.com/trusteddomainproject/OpenDKIM/blob/5c539587561785a66c1f67f720f2fb741f320785/opendkim/opendkim.c#L11919-L11944

link2xt:
Seems the problem is that all the OUTGOING stuff works, but final.lua runs even for outgoing messages and this does not work anymore:
https://github.com/chatmail/relay/blob/7191329a9f5100d220a3770733fd8d9cc66f532a/cmdeploy/src/cmdeploy/opendkim/final.lua#L1-L5

We should not use `internal_ip` anymore.

PR #785, opened 2025-12-18
fix: expire messages also from DeltaChat IMAP subfolders

hpk42:
this relates to a debugging session with chatmail operators who saw messages on their relay in .DeltaChat folder, see also https://github.com/chatmail/core/issues/7623

### DIFF
--- a/chatmaild/src/chatmaild/expire.py
+++ b/chatmaild/src/chatmaild/expire.py
@@ -63,21 +63,28 @@ class MailboxStat:
             os.chdir(self.basedir)
         except FileNotFoundError:
             return
-        for name in os_listdir_if_exists("."):
+        try:
+            self.scandir(".")
+        finally:
+            os.chdir(old_cwd)
+
+    def scandir(self, dirname):
+        for name in os_listdir_if_exists(dirname):
+            relpath = f"{dirname}/{name}"
             if name in ("cur", "new", "tmp"):
-                for msg_name in os_listdir_if_exists(name):
-                    entry = get_file_entry(f"{name}/{msg_name}")
+                for msg_name in os_listdir_if_exists(relpath):
+                    entry = get_file_entry(f"{relpath}/{msg_name}")
                     if entry is not None:
                         self.messages.append(entry)
-
+            elif relpath == "./.DeltaChat":
+                self.scandir(name)
             else:
-                entry = get_file_entry(name)
+                entry = get_file_entry(relpath)
                 if entry is not None:
                     self.extrafiles.append(entry)
                     if name == "password":
                         self.last_login = entry.mtime
         self.extrafiles.sort(key=lambda x: -x.size)
-        os.chdir(old_cwd)
 
 
 def print_info(msg):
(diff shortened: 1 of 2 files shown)


chatmaild/src/chatmaild/expire.py
```diff
@@ -63,21 +63,28 @@ def __init__(self, basedir):
             os.chdir(self.basedir)
         except FileNotFoundError:
             return
-        for name in os_listdir_if_exists("."):
+        try:
+            self.scandir(".")
+        finally:
+            os.chdir(old_cwd)
```
link2xt, line 69: Somewhat related, would be nice to get rid of these `chdir` calls, they already caused bad error messages when we had disk problems and this `chdir` on exception does not look reliable to me as it may itself fail: https://github.com/chatmail/relay/issues/786

hpk42: Still better in a separate PR.

chatmaild/src/chatmaild/tests/test_expire.py
```diff
@@ -17,9 +17,7 @@
 from chatmaild.fsreport import main as report_main
 
 
-def fill_mbox(basedir):
-    basedir1 = basedir.joinpath("mailbox1@example.org")
-    basedir1.mkdir()
+def fill_mbox(basedir1):
```
link2xt, line 20: Maybe rename this `basedir1` to `mboxdir` at least. `basedir1` was not descriptive already, but at least it was clear that this avoided conflict with `basedir`, now this `1` is here for historical reasons only.

chatmaild/src/chatmaild/expire.py
hpk42: yes, but not in this PR. let's merge this (if you can agree) and then i'll do a separate PR for using absolute path.

chatmaild/src/chatmaild/expire.py
```diff
@@ -150,7 +157,7 @@ def process_mailbox_stat(self, mbox):
                 self.remove_file(message.relpath, mtime=message.mtime)
             elif message.size > 200000 and message.mtime < cutoff_large_mails:
                 # we only remove noticed large files (not unnoticed ones in new/)
-                if message.relpath.startswith("cur/"):
+                if "cur" in message.relpath.split("/"):
```
link2xt, line 160: Could this check for `message.relpath.split("/")[-2] instead? And catch `IndexError` I guess because there is no .get method for lists: https://discuss.python.org/t/add-safe-get-method-to-list/32555

chatmaild/src/chatmaild/expire.py
```diff
@@ -63,21 +63,28 @@ def __init__(self, basedir):
...
+            relpath = f"{dirname}/{name}"
             if name in ("cur", "new", "tmp"):
-                for msg_name in os_listdir_if_exists(name):
-                    entry = get_file_entry(f"{name}/{msg_name}")
+                for msg_name in os_listdir_if_exists(relpath):
+                    entry = get_file_entry(f"{relpath}/{msg_name}")
                     if entry is not None:
                         self.messages.append(entry)
-
+            elif relpath == "./.DeltaChat":
```
link2xt, line 79: If user logged in with Thunderbird this may also be "Sent" and "Drafts", would be better to just scan everything instead of hardcoding "DeltaChat".

link2xt (approved):

PR #790, opened 2025-12-24
ci: remove jsok/serialize-workflow-action dependency

link2xt:
Deployments to test servers will not be cancelled anymore, but it is not clear if we even want it.
This setup is much simpler because it only depends on GitHub Actions features and does not allocate
a runner just to sleep there and wait in the queue.


link2xt:
This is a follow-up to #750
This dependency was introduced in https://github.com/chatmail/relay/pull/361, but the problem can be solved by just not cancelling already started runs.

hpk42 (approved):
fine to test this.

Issue #761, opened 2025-12-09
Documentation for relay; chat.example.com or chat.example.org?


joho1968:
This documentation URL for chat relays: https://chatmail.at/doc/relay/getting_started.html mentions both `chat.example.com` and `chat.example.org`.

Maybe I'm the confused one, but shouldn't it be chat.example.org everywhere? 🤔 

<img width="778" height="712" alt="Image" src="https://github.com/user-attachments/assets/68284570-ab2c-4798-90e7-7519bac9fa03" />

missytake:
Thanks for the report, we should use chat.example.org consistently I think.

ccclxxiii:
addressed in #793

PR #793, opened 2025-12-28
use consistent naming schema in getting started

ccclxxiii:
pr overview:
- addresses #761 
- previously we had mixed example fqdns, now the proposal is to move to `chat.example.org`


link2xt (approved):

adbenitez (approved):
thanks!

PR #794, opened 2026-01-02
WIP: nginx: multiplex SSH over port 443 in case port 22 is blocked

missytake:
In some networks, port 22 might be blocked. If we can use SSH over port 443, that would mitigate this.


missytake:
If only it was this easy :/

I'm trying to get this work via this guide: https://www.gangsto.org/posts/ssh-on-port-443-part-1/
Accordings to this we need to set `ProxyCommand openssl s_client -connect %h:443 -alpn ssh -verify_return_error` in the `~/.ssh/config` to make it work. But I didn't manage connecting to it so far:

```
OpenSSH_9.6p1 Ubuntu-3ubuntu13.14, OpenSSL 3.0.13 30 Jan 2024
debug1: Reading configuration data /home/user/.ssh/config
debug1: /home/user/.ssh/config line 1: Applying options for *
debug1: /home/user/.ssh/config line 188: Applying options for staging2.testrun.org
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 19: include /etc/ssh/ssh_config.d/*.conf matched no files
debug1: /etc/ssh/ssh_config line 21: Applying options for *
debug1: Executing proxy command: exec openssl s_client -connect staging2.testrun.org:443 -alpn ssh -verify_return_error #-quiet -verify_quiet
debug1: identity file /home/user/.ssh/staging.testrun.org type 3
debug1: identity file /home/user/.ssh/staging.testrun.org-cert type -1
debug1: Local version string SSH-2.0-OpenSSH_9.6p1 Ubuntu-3ubuntu13.14
40F7AD4DCC780000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:../ssl/record/ssl3_record.c:354:
debug1: kex_exchange_identification: banner line 0: CONNECTED(00000003)
debug1: kex_exchange_identification: banner line 1: ---
debug1: kex_exchange_identification: banner line 2: no peer certificate available
debug1: kex_exchange_identification: banner line 3: ---
debug1: kex_exchange_identification: banner line 4: No client certificate CA names sent
debug1: kex_exchange_identification: banner line 5: ---
debug1: kex_exchange_identification: banner line 6: SSL handshake has read 5 bytes and written 332 bytes
debug1: kex_exchange_identification: banner line 7: Verification: OK
debug1: kex_exchange_identification: banner line 8: ---
debug1: kex_exchange_identification: banner line 9: New, (NONE), Cipher is (NONE)
debug1: kex_exchange_identification: banner line 10: Secure Renegotiation IS NOT supported
debug1: kex_exchange_identification: banner line 11: Compression: NONE
debug1: kex_exchange_identification: banner line 12: Expansion: NONE
debug1: kex_exchange_identification: banner line 13: No ALPN negotiated
debug1: kex_exchange_identification: banner line 14: Early data was not sent
debug1: kex_exchange_identification: banner line 15: Verify return code: 0 (ok)
debug1: kex_exchange_identification: banner line 16: ---
kex_exchange_identification: Connection closed by remote host
Connection closed by UNKNOWN port 65535
```

Seems like the client uses the wrong TLS version, but even with `-tls1_3` it doesn't work. Further debugging needed.

ccclxxiii:
> In some networks, port 22 might be blocked. If we can use SSH over port 443, that would mitigate this.

this is interesting idea, i've used this before.  the question(s) i would pose is:
- if the operator of the relay already has access, is it a problem? 
- if a network blocks `tcp/22` is that our area of responsibility/concern?  
- if we implement `tcp/22` into nginx via alpn, won't we create a coupling that might interfere with normal operations of configuration management? (ie an operator that uses ansible or some other system that requires ssh) if the nginx endpoint is down, then they could possibly not reach their server to make changes.  

just some thoughts i had on it.  if you want to chat through it, you know where my office is.

DarkCat09:
> In some networks, port 22 might be blocked

The port or the SSH protocol?

If it's the former, an admin should just modify SSH server config (they should do it anyway btw, I have no idea why cmdeploy doesn't accept a custom port, it's a usual security advice). Proxying SSH through nginx seems weird, to be honest.

If you're trying to bypass protocol detection, then wrapping it in TLS is probably okay, but I'd rather suggest relay admins to use proxying software that is *made for such cases* (xray, sing-box?) instead of crafting custom tunnels.

DarkCat09:
Like, it looks cool, but imho there are better tools for that, and an admin should already have a reliable access to their server before deploying a relay

missytake:
>     * if the operator of the relay already has access, is it a problem?

No, you can still just login via port 22 as well, no need to change or document this actually.

>     * if a network blocks `tcp/22` is that our area of responsibility/concern?

I was unsure whether I ran into this problem myself, that was my main reason for this PR.

>     * if we implement `tcp/22` into nginx via alpn, won't we create a coupling that might interfere with normal operations of configuration management? (ie an operator that uses ansible or some other system that requires ssh) if the nginx endpoint is down, then they could possibly not reach their server to make changes

As port 22 would continue to work, I don't think it causes a problem. It's just for people who know the workaround and configure it client-side - they can simply unconfigure it if nginx is down.

missytake:
> If you're trying to bypass protocol detection, then wrapping it in TLS is probably okay, but I'd rather suggest relay admins to use proxying software that is _made for such cases_ (xray, sing-box?) instead of crafting custom tunnels.

tbh I didn't fully understand so far if this approach also wraps it in TLS - I hope so, but then again the bug which is blocking me so far seems related to that question.

https://github.com/XTLS/Xray-core looks interesting, might look into it later. In any case it requires more thought than a simple tunnel we already use for other things.

ccclxxiii:
> No, you can still just login via port 22 as well, no need to change or document this actually.

i understand, for me this adds a bit of complexity to the relay overall operation and widens the attack surface.  an example would be if an operator allows `tcp/22` from their scoped cidr ie `5.5.2.34/32` and `127.0.0.1` only, potentially an attacker could try to use the `tcp/443` endpoint to connect to `tcp/22` via loopback.  

> I was unsure whether I ran into this problem myself, that was my main reason for this PR.

what i meant was is it the relay's responsibility to add an ssh feature for protocol/network filtering evasion?  i don't think so, personally.  the operator of a system should be able to harden administer their own system.

> As port 22 would continue to work, I don't think it causes a problem. It's just for people who know the workaround and configure it client-side - they can simply unconfigure it if nginx is down. 

for clarity, i like the idea (right up my alley), however i'm not sure it brings a lot of value compared to other initiatives that are in-flight or planned.  :thinking: 

if you want to walk through or pair on it, reach out for a chat :lock:

hpk42:
maybe it makes more sense to document/point to a way to let ssh listen on additional ports?

PR #798, opened 2026-01-05
feat(dkim): Replace OpenDKIM with DKIM Milter

j-g00da:


chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -121,3 +121,11 @@
...
 # postal address of the privacy supervisor
 privacy_supervisor =
+
+
+#
+# Experimental Options
+#
+
+# use dkim-milter instead of OpenDKIM
+dkim_milter = false
```
link2xt: In the end we don't need an option, it is fine to remove everything OpenDKIM-related. But existing users need a way to migrate. Ideally old key should just be reused, so no DNS changes are needed.

cmdeploy/src/cmdeploy/dkim_milter/deployer.py
```diff
@@ -0,0 +1,117 @@
...
+    def install(self):
+        """Builds and installs dkim-milter"""
+
+        # openssl is required to generate the signing key
+        apt.packages(
+            name="Install rustc, cargo and openssl required by dkim-milter",
+            packages=["rustc", "cargo", "openssl"],
+        )
+
+        cargo.packages(
```
link2xt: Instead of building on the server, the binaries are better built on CI similar to how chatmail-turn is doing this: https://github.com/chatmail/dkim-milter/issues/1

chatmaild/src/chatmaild/ini/chatmail.ini.f
j-g00da: One downside of this I see is that we will leave "opendkim" in names (e.g. selector) which may be a bit confusing - but I guess it's fine.

j4n:
Setting aside the binary deployment deployment strategy, looks good to me. Having worked on https://github.com/chatmail/relay/issues/530 I wonder if dkim-milter can do the stripping of validated DKIM-Signatures as well? I see the default config does strip the Authentication-Results already.

j-g00da:
I'm currently using self hosted binaries with custom patches (https://github.com/chatmail/dkim-milter/pull/3) to allow communication over unix domain socket.

What I think is mostly left to do in order to complete this is to port lua scripts. Since dkim-milter doesn't support custom scripts, I will probably just implement it directly on our fork.

hpk42:
On Mon, Jan 12, 2026 at 06:34 -0800, Jagoda Estera Ślązak wrote:
> j-g00da left a comment (chatmail/relay#798)
> 
> I'm currently using self hosted binaries with custom patches (https://github.com/chatmail/dkim-milter/pull/3) to allow communication over unix domain socket.
> 
> What I think is mostly left to do in order to complete this is to port lua scripts. Since dkim-milter doesn't support custom scripts, I will probably just implement it directly on our fork.

Sounds good to directly implement the lua scripts' logic, maybe in some extra module.

j-g00da:
This currently works, CI fails due to unrelated issues.

Few notes:
1. dkim-milter is less flexible than OpenDKIM, but I was able to get around it by just running two service instances for verification and signing. Advantage of that is a clear separation of these features, that will make it easy to move either verification or signing to filtermail later on.
2. dkim-milter's reject policies lets us enforce same policy within the config, so modification of the source to provide the lua script functionalities wasn't needed.

link2xt:
This can be closed in favor of https://github.com/chatmail/relay/pull/831 I think.

j-g00da:
> This can be closed in favor of https://github.com/chatmail/relay/pull/831 I think.

Depends on what we want to do with the signing side, we can either leave opendkim, replace it with dkim-milter or integrate signing into filtermail.

j-g00da:
This is stale, and we are more likely to implement signing in filtermail now. Closing.

PR #799, opened 2026-01-05
fix: Remove Authentication-Results header and correct header iteration

j4n:
Header indices start with 0; fix loop for purging DKIM headers, and also remove Authentication-Results header

### DIFF
--- a/cmdeploy/src/cmdeploy/opendkim/final.lua
+++ b/cmdeploy/src/cmdeploy/opendkim/final.lua
@@ -35,8 +35,10 @@ if valid then
 		odkim.del_header(ctx, "DKIM-Signature", i)
 	end
 
-	-- Delete first and presumably only occurence
-        odkim.del_header(ctx, "Authentication-Results", 0)
+	-- Delete up to 10 Authentication-Results headers
+	for i = 9, 0, -1 do
+		odkim.del_header(ctx, "Authentication-Results", 0)
+	end
 else
 	odkim.set_reply(ctx, "554", "5.7.1", error_msg)
 	odkim.set_result(ctx, SMFIS_REJECT)


cmdeploy/src/cmdeploy/opendkim/final.lua
```diff
@@ -31,9 +31,12 @@ end
 if valid then
 	-- Strip all DKIM-Signature headers after successful validation
 	-- Delete in reverse order to avoid index shifting.
-	for i = nsigs, 1, -1 do
+	for i = nsigs - 1, 0, -1 do
 		odkim.del_header(ctx, "DKIM-Signature", i)
 	end
+
+	-- Delete first and presumably only occurence
+        odkim.del_header(ctx, "Authentication-Results", 0)
```
hpk42: there can be multiple Authentication-Results headers IIRC see https://datatracker.ietf.org/doc/html/rfc7601#appendix-B.5 

It's best to remove all.

j4n: There is no mechanism to count these, as opposed to the DKIM headers; so the last commit just tries to delete up to 10 headers.

hpk42 (commented):
ok, looks good but a test is missing and would be useful also after we phase in the dkim-milter approach.  Test would need to: 

- use "maildomain2" to create an address and then send a message to mail_domain.  See test_securejoin in test_2_deltachat.py 

- use eg  "imap_tools" (grep for it) to retrieve the headers and checking if authentication-results or dkim signature headers are present in the mailbox (on maildomain). 

could you try to add such a test?

missytake:
https://github.com/chatmail/relay/pull/803 looks more promising, as it also cleans up internal messages and doesn't require lua code. Let's close this for now :)

PR #802, opened 2026-01-10
fix: properly make sure that postfix gets restarted on failure

hpk42:
found this fix in omidz4t fork 
 https://github.com/chatmail/relay/compare/main...omidz4t:relay-ir:main

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/deployer.py
+++ b/cmdeploy/src/cmdeploy/postfix/deployer.py
@@ -65,7 +65,7 @@ class PostfixDeployer(Deployer):
         restart_conf = files.put(
             name="postfix: restart automatically on failure",
             src=get_resource("service/10_restart.conf"),
-            dest="/etc/systemd/system/postfix.service.d/10_restart.conf",
+            dest="/etc/systemd/system/postfix@.service.d/10_restart.conf",
         )
         self.daemon_reload = restart_conf.changed
         self.need_restart = need_restart


hpk42:
hum, CI fails. it seems postfix isn't actually having a systemd service unit or so? 
my manual deploy to c2.testrun.org worked, however, and did not give this error.

link2xt:
There is a postfix systemd unit at `/lib/systemd/system/postfix.service`, this path is displayed with `systemctl status postfix`.

link2xt:
The error in CI log:
```
--> Starting operation: Start and enable Postfix 
    [staging2.testrun.org] Failed to restart postfix.service: Unit postfix.service has a bad unit file setting.
    [staging2.testrun.org] See system logs and 'systemctl status postfix.service' for details.
    [staging2.testrun.org] Error: executed 1 commands
```

j-g00da:
More context:
```
Service has Restart= set to either always or on-success, which isn't allowed for Type=oneshot services. Refusing.
```

adbenitez:
btw, does dovecot already has similar restarting service? it happened to me in the past that dovecot got killed in a small vps during a peak/bottleneck while I was sleeping and the service was down until I realize it, having some downtime hours unnecessarily, that could have been avoided if it would have just restarted

missytake:
> btw, does dovecot already has similar restarting service? it happened to me in the past that dovecot got killed in a small vps during a peak/bottleneck while I was sleeping and the service was down until I realize it, having some downtime hours unnecessarily, that could have been avoided if it would have just restarted

Yes, for dovecot we have it since https://github.com/chatmail/relay/pull/765, mostly because you reported that problem

missytake:
> hum, CI fails. it seems postfix isn't actually having a systemd service unit or so? my manual deploy to c2.testrun.org worked, however, and did not give this error.

postfix has a more complex systemd unit architecture, which is not easily transferrable from dovecot.

`postfix.service` is a oneshot service that triggers many smaller processes, and then exits. That's why it's working state is `active (exited)`, while dovecot is `active (running)`. The main postfix unit will not really fail during operation like that, because it isn't actually running during operation. So restarting on failure makes less sense.

~~We should remove the restarting logic from the postfix deployer, which anyway only tries to restart dovecot because of an earlier oversight, and then think about which problem we really want to solve with restarting postfix.~~

cmdeploy/src/cmdeploy/postfix/deployer.py
```diff
@@ -65,7 +65,7 @@ def configure(self):
         restart_conf = files.put(
             name="postfix: restart automatically on failure",
             src=get_resource("service/10_restart.conf"),
-            dest="/etc/systemd/system/dovecot.service.d/10_restart.conf",
+            dest="/etc/systemd/system/postfix.service.d/10_restart.conf",
```
missytake:
```suggestion
            dest="/etc/systemd/system/postfix@.service.d/10_restart.conf",
```

wait, this is what omid's fork is actually doing. Maybe that works.

missytake:
Looks like this is the systemd unit we actually want to restart:

```
root@staging2:~# systemctl status postfix@-
● postfix@-.service - Postfix Mail Transport Agent (instance -)
     Loaded: loaded (/lib/systemd/system/postfix@.service; enabled-runtime; preset: enabled)
    Drop-In: /etc/systemd/system/postfix@.service.d
             └─10_restart.conf
     Active: active (running) since Tue 2026-01-13 22:52:40 UTC; 8min ago
       Docs: man:postfix(1)
      Tasks: 6 (limit: 4530)
     Memory: 8.7M
        CPU: 2.755s
     CGroup: /system.slice/system-postfix.slice/postfix@-.service
             ├─31184 /usr/lib/postfix/sbin/master -w
             ├─31185 pickup -l -t unix -u -c
             ├─31186 qmgr -l -t unix -u
             ├─31407 tlsmgr -l -t unix -u -c
             ├─32447 smtpd -n smtp -t inet -u -c -o stress= -s 2 -o smtpd_tls_security_level=encrypt -o "smtpd_tls_mandatory_protocols=>=TLSv1.2" -o smtpd_proxy_filter=127.0.0.1:10081
             └─32449 anvil -l -t unix -u -c

Jan 13 22:54:15 staging2 postfix/smtps/smtpd[31408]: disconnect from unknown[172.182.226.233] ehlo=1 auth=1 mail=62 rcpt=61/62 data=61 rset=1 commands=187/188
Jan 13 22:54:17 staging2 postfix/smtps/smtpd[31405]: disconnect from unknown[172.182.226.233] ehlo=1 auth=1 mail=1 rcpt=1 data=1 commands=5
Jan 13 22:57:37 staging2 postfix/anvil[31410]: statistics: max connection rate 51/60s for (smtps:172.182.226.233) at Jan 13 22:54:04
Jan 13 22:57:37 staging2 postfix/anvil[31410]: statistics: max connection count 12 for (smtps:172.182.226.233) at Jan 13 22:53:24
Jan 13 22:57:37 staging2 postfix/anvil[31410]: statistics: max cache size 3 at Jan 13 22:53:52
Jan 13 22:59:10 staging2 postfix/smtpd[32447]: connect from unknown[158.94.210.39]
Jan 13 22:59:10 staging2 postfix/smtpd[32447]: disconnect from unknown[158.94.210.39] ehlo=1 auth=0/1 rset=0/1 quit=1 commands=2/4
Jan 13 23:00:45 staging2 postfix/smtpd[32447]: connect from unknown[31.129.22.226]
Jan 13 23:00:45 staging2 postfix/smtpd[32447]: lost connection after EHLO from unknown[31.129.22.226]
Jan 13 23:00:45 staging2 postfix/smtpd[32447]: disconnect from unknown[31.129.22.226] ehlo=0/1 commands=0/1
root@staging2:~# systemctl cat postfix@-
# /lib/systemd/system/postfix@.service
[Unit]
Description=Postfix Mail Transport Agent (instance %i)
Documentation=man:postfix(1)
PartOf=postfix.service
Before=postfix.service
ReloadPropagatedFrom=postfix.service
After=network-online.target nss-lookup.target
Wants=network-online.target

[Service]
Type=forking
GuessMainPID=no
ExecStartPre=/usr/lib/postfix/configure-instance.sh %i
ExecStart=/usr/sbin/postmulti -i %i -p start
ExecStop=/usr/sbin/postmulti -i %i -p stop
ExecReload=/usr/sbin/postmulti -i %i -p reload

[Install]
WantedBy=multi-user.target

# /etc/systemd/system/postfix@.service.d/10_restart.conf
[Service]
Restart=always
RestartSec=30
```

Now it should work. Can a fresh pair of eyes look over it?

PR #803, opened 2026-01-13
Strip DKIM-Signature header before LMTP

### LINKED ISSUE #530: Remove DKIM signature after checking it

feld:
Currently we strip the DKIM-Signature header in the OpenDKIM final.lua script after validation of the signature. We sign all messages upon submission, but we do not verify messages which are from a local account and delivered to another local account.

This corrects the problem and ensures that the plaintext headers of a local to local delivery are sanitized the same as a message received from another server.

The functionality in final.lua to strip the DKIM-Signature header can now be retired.

Fixes #530

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
+++ b/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
@@ -1,1 +1,2 @@
 /^DKIM-Signature:/            IGNORE
+/^Authentication-Results:/            IGNORE


cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -0,0 +1 @@
+/^DKIM-Signature:/            IGNORE
```
missytake, line 1:
```suggestion
/^DKIM-Signature:/            IGNORE
/^Authentication-Results:/            IGNORE
```

This can also be stripped, right?

missytake: Nevermind, internal messages (between addresses on the same relay) don't have the `Authentication-Results` header anyway, and it isn't about incoming messages from other relays (as dkim-milter will take care of it), right?

feld: We have never stripped that header from what I can tell, but we could if we wanted to. I don't believe it has any sensitive information we should be concerned about. (unlike the DKIM-Signature non-repudiation problem we're attempting to eliminate)

missytake: Ah right. I thought it was about saving storage space.

missytake:
Looking at the notes of #799 and #530, this looks like the most promising approach for this. But a test would be good and useful, also because we phase in the dkim-milter approach and need to know whether it still works then. Test would need to:

- use "maildomain2" to create an address and then send a message to mail_domain. See test_securejoin in test_2_deltachat.py
- use eg "imap_tools" (grep for it) to retrieve the headers and checking if authentication-results or dkim signature headers are present in the mailbox (on maildomain).

could you try to add such a test? I can also try, if you have no time.

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
missytake: In any case it also isn't needed by the client, so we can strip it, #799 also tried that.

feld:
> could you try to add such a test? I can also try, if you have no time.

I'm a little short on time this week and I also don't have any experience with the test framework in this project repo, so if someone who has familiarity with this could step in that would be great

PR #806, opened 2026-01-14
feat(dovecot): add config flag to export statistics

j4n:
This adds exporting of some dovecot event metrics to help debugging slow IMAP login and hibernation. We're hijacking the `mtail_address` config flag here and configure the port of the dovecot exporter to be the one over mtail's, `3904`. Renaming that flag to, e.g., `openmetrics_address`, would be sensible in the future. Next, we hope to test these settings in production shortly and follow up with documentation and an example Grafana dashboard.


missytake (approved):
Thanks :) any reason to leave out `dovecot-metrics.sh` from https://github.com/deltachat/sysadmin-pyinfra/pull/161 ?

j4n:
just trying to keep things modular.

PR #808, opened 2026-01-15
feat(filtermail): Replace filtermail with rust reimplementation

j-g00da:
~~WIP, for testing purposes it downloads a binary I self-host and sets the logging level to debug.~~


cmdeploy/src/cmdeploy/filtermail/deployer.py
```diff
@@ -0,0 +1,48 @@
...
+        # TODO: hash/arch checks, get binary from proper github release
+        url = "https://kamiokan.de/bin/filtermail"
+        server.shell(
+            name="Download filtermail",
+            commands=[
+                f"curl -L {url} -o {self.bin_path}.new && mv {self.bin_path}.new {self.bin_path}",
+                f"chmod 755 {self.bin_path}",
+            ],
+        )
+        self.need_restart = True
```
j4n:
```suggestion
        self.need_restart |= files.download(
            name="Download filtermail if updated",
            src=url,
            dest=self.bin_path,
            mode="755",
        ).changed
```
Use built-in download function will do more checks.

chatmaild/src/chatmaild/tests/test_filtermail_blackbox.py
j4n, line 1: This test should probably be kept, as it works just as well with the new implementation, except for hanging on startup as the old implementation (`pip install pytest_localserver`):
```
(venv) ~/delta/relay/chatmaild [j-g00da/filtermail-rs] % pytest src/chatmaild/tests/test_filtermail_blackbox.py -v
====================== test session starts =======================
platform linux -- Python 3.10.12, pytest-9.0.1, pluggy-1.6.0 -- /home/ja/delta/relay/venv/bin/python3
cachedir: .pytest_cache
Deltachat core=v2.23.0 sqlite=3.46.1 journal_mode=wal
rootdir: /home/ja/delta/relay/chatmaild
configfile: pyproject.toml
plugins: deltachat-2.23.0, localserver-0.10.0, cmdeploy-0.2, deltachat-rpc-client-2.28.0, xdist-3.8.0, anyio-4.11.0, typeguard-4.4.4
collected 2 items                                                

src/chatmaild/tests/test_filtermail_blackbox.py::test_one_mail[outgoing] PASSED [ 50%]
src/chatmaild/tests/test_filtermail_blackbox.py::test_one_mail[incoming] PASSED [100%]

======================= 2 passed in 0.51s ========================
(venv) ~/delta/relay/chatmaild [j-g00da/filtermail-rs] % 
```
I had to patch out waiting for the "entering serving loop" startup message of the old implementation:
```
-    line = popen.stderr.readline().strip()
-    if b"loop" not in line:
-        print(line.decode("ascii"), file=sys.stderr)
-        pytest.fail("starting filtermail failed")
+
+    # Wait for filtermail to start accepting connections
+    import socket
+    import time
+    for _ in range(50):  # 5 second timeout
+        try:
+            sock = socket.create_connection(("127.0.0.1", smtp_inject_port), timeout=0.1)
+            sock.close()
+            break
+        except (ConnectionRefusedError, OSError):
+            time.sleep(0.1)
+    else:
+        pytest.fail("filtermail failed to start accepting connections")
```
so maybe its worthwhile to bring that output back for full backwards-compatibility?

j4n (changes requested):
Looks great to me generally based on preliminary testing, two suggestions :)

cmdeploy/src/cmdeploy/filtermail/deployer.py
j-g00da: I will do this properly after initial release of the binary since this will also require e.g. choosing architecture.

missytake:
Some operators might not have set `filtermail_smtp_port_incoming`, `filtermail_smtp_port`, `postfix_reinject_port`, and `postfix_reinject_port_incoming` in their chatmail.ini - can we set a default value for them in config.py, and in filtermail-rs? The values missing crashed our service on updating

j-g00da:
All known issues now fixed.

.github/workflows/ci.yaml
```diff
@@ -14,7 +14,8 @@ jobs:
         # Otherwise `test_deployed_state` will be unhappy.
         with:
           ref: ${{ github.event.pull_request.head.sha }}
-
+      - name: download filtermail
+        run: curl -L https://github.com/chatmail/filtermail/releases/download/v0.1.2/filtermail-x86_64 -o /usr/local/bin/filtermail && chmod +x /usr/local/bin/filtermail
```
link2xt, line 18: This step is not needed anymore, right?

j-g00da: I think it's still needed, offline chatmaild tests don't run cmdeploy so the binary will be missing

link2xt: ok, so it's for test_filtermail_blackbox.py

link2xt:
test_exceed_rate_limit fails again. It was fine with just the increase from 5 to 25.

j-g00da:
> test_exceed_rate_limit fails again. It was fine with just the increase from 5 to 25.

it was fine only locally, still failing in CI, I guess I have to increase it further.

link2xt:
Follow-up issue for the test: https://github.com/chatmail/relay/issues/821

PR #581, opened 2025-06-04
Use static binary from official mtail release instead of Debian package

link2xt:
Debian has outdated version that does not actually work with logs from stdin. It gets stuck after some time.

Closes #579

Issue #579, opened 2025-06-02
mtail gets stuck


link2xt:
Debian-installed mtail gets stuck and does not update counters after some time.

https://github.com/chatmail/relay/issues/484 was not really fixed with https://github.com/chatmail/relay/pull/502
After restarting, mtail works for some time, then metric counters stop.

We need to update mtail to at least [v3.0.0-rc55](https://github.com/google/mtail/releases/tag/v3.0.0-rc55).

Debian bookworm (current stable) packages 3.0.0~rc50-1+b6
There are no backports.

Debian trixie (current testing) packages 3.0.9-1+b6.

We can try to directly install http://ftp.de.debian.org/debian/pool/main/m/mtail/mtail_3.0.9-1+b6_amd64.deb
(URL from https://packages.debian.org/trixie/amd64/mtail/download)
or just install mtail into /usr/local/bin by unpacking https://github.com/google/mtail/releases/download/v3.0.8/mtail_3.0.8_linux_amd64.tar.gz

Probably better just unpack into /usr/local/bin, we do the same for iroh-relay anyway.

missytake:
+1 for installing mtail directly to /usr/local/bin and not pulling it from the Debian repository; too likely that having two different Debian versions in our apt sources fucks up Debian's dependency management, and an actually critical service gets disrupted.

PR #812, opened 2026-01-19
mtail: explicitly read from /dev/stdin

feld:
I wish I could understand why this was happening, but even though my deployment with my Chatmail cookbook produced a 100% identical ExecStart command, the mtail process was trying to read a file named /- which did not exist.

This is a harmless change and should ensure nobody else runs into this mysterious edge case.

### DIFF
--- a/cmdeploy/src/cmdeploy/mtail/mtail.service.j2
+++ b/cmdeploy/src/cmdeploy/mtail/mtail.service.j2
@@ -3,7 +3,7 @@ Description=mtail
 
 [Service]
 Type=simple
-ExecStart=/bin/sh -c "journalctl -f -o short-iso -n 0 | /usr/local/bin/mtail --address={{ address }} --port={{ port }} --progs /etc/mtail --logtostderr --logs -"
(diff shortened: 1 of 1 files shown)


missytake:
Can you rebase this on top of https://github.com/chatmail/relay/pull/813 so CI can test this PR?

link2xt:
It could be you have old `mtail` version that does not support `-`. Changing to `/dev/stdin` then will not fix it, it will just get stuck after some time. See https://github.com/chatmail/relay/pull/581 and https://github.com/chatmail/relay/issues/579

feld:
> It could be you have old `mtail` version that does not support `-`.

that doesn't make sense to me, "-" is a shell feature not something that mtail would need to natively support, right? 

edit: huh! It's not a shell feature? TIL after all these years... 

anyway, I'm not using an old mtail version I'm installing 3.0.9 which is the latest. It's not getting stuck for me, seems to be working great now with /dev/stdin

link2xt:
I think new version is not getting stuck since about the same time as they added `-`. Likely added `-`, tried it, realized it's broken, then fixed. So using `-` instead of `/dev/stdin` ensures that the version is at least somewhat new.

missytake:
@feld do you still need this change? Otherwise I'd like to close the PR, as we have too many open PRs atm and it's hard to keep an overview.

PR #813, opened 2026-01-20
cmdeploy: fail on errors in postfix/dovecot config

missytake:
This should fail CI, at least the commit failed when deployed manually.

When we confirmed that CI fails on broken config, we can fix the dovecot config and merge to main.

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/deployer.py
+++ b/cmdeploy/src/cmdeploy/postfix/deployer.py
@@ -1,4 +1,4 @@
-from pyinfra.operations import apt, files, systemd
+from pyinfra.operations import apt, files, server, systemd
 
 from cmdeploy.basedeploy import Deployer, get_resource
 
@@ -77,6 +77,14 @@ class PostfixDeployer(Deployer):
             dest="/etc/systemd/system/postfix@.service.d/10_restart.conf",
         )
         self.daemon_reload = restart_conf.changed
+
+        # Validate postfix configuration before restart
+        if need_restart:
+            server.shell(
+                name="Validate postfix configuration",
+                # Extract stderr and quit with error if non-zero
+                commands=["""bash -c 'w=$(postconf 2>&1 >/dev/null); [[ -z "$w" ]] || { echo "$w"; false; }'"""],
+            )
         self.need_restart = need_restart
 
     def activate(self):
--- a/cmdeploy/src/cmdeploy/dovecot/deployer.py
+++ b/cmdeploy/src/cmdeploy/dovecot/deployer.py
@@ -37,9 +37,7 @@ class DovecotDeployer(Deployer):
         restart = False if self.disable_mail else self.need_restart
 
         systemd.service(
-            name="disable dovecot for now"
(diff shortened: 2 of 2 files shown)


missytake:
Fails with:

```
Jan 20 18:40:43 staging2 systemd[1]: dovecot.service: Scheduled restart job, restart counter is at 17.
Jan 20 18:40:43 staging2 systemd[1]: Stopped dovecot.service - Dovecot IMAP/POP3 email server.
Jan 20 18:40:43 staging2 systemd[1]: Starting dovecot.service - Dovecot IMAP/POP3 email server...
Jan 20 18:40:43 staging2 dovecot[30980]: doveconf: Fatal: Error in configuration file /etc/dovecot/dovecot.conf: event filter: syntax error, unexpected invalid token, expecting '=' or '>' or '<'
Jan 20 18:40:43 staging2 systemd[1]: dovecot.service: Main process exited, code=exited, status=89/n/a
Jan 20 18:40:43 staging2 systemd[1]: dovecot.service: Failed with result 'exit-code'.
Jan 20 18:40:43 staging2 systemd[1]: Failed to start dovecot.service - Dovecot IMAP/POP3 email server.
```

.github/workflows/test-and-deploy.yaml
```diff
@@ -74,7 +74,9 @@ jobs:
       - name: run deploy-chatmail offline tests 
         run: pytest --pyargs cmdeploy 
 
-      - run: cmdeploy init staging2.testrun.org
+      - run: |
+          cmdeploy init staging2.testrun.org
+          sed -i 's/# mtail_address/mtail_address/' chatmail.ini
```
j4n: maybe be lenient on the whitespace?
```
sed -i 's/#\s*mtail_address/mtail_address/' chatmail.ini 
```

j4n:
Ok, addressed the syntax error (I sincerely wonder how this slipped through, I suppose it only manifests on a full dovecot restart); and added a config check before restarting so syntax errors don't bring down the service again.

j4n:
maybe it makes sense to do such a check on postfix as well?

missytake:
> maybe it makes sense to do such a check on postfix as well?

trickier, `postconf` returns exit code 0 on warnings. But I found a way.

cmdeploy/src/cmdeploy/postfix/deployer.py
```diff
@@ -77,6 +77,13 @@ def configure(self):
             dest="/etc/systemd/system/postfix@.service.d/10_restart.conf",
         )
         self.daemon_reload = restart_conf.changed
+
+        # Validate postfix configuration before restart
+        if need_restart:
+            server.shell(
+                name="Validate postfix configuration",
+                commands=["postconf -n >/dev/null"],
```
missytake: this will not fail on warnings, only fatal errors; e.g. wrong config keys only cause warnings. But that's fine I guess, at least the service will start normally.

j4n (approved):
improved postconf checks (not as nice as doveconf)

PR #816, opened 2026-01-20
Aggressive LMTP header cleanup

feld:
This will remove all headers possible during LMTP delivery.

From header: required or clients do not process the message correctly

Message-Id header: required for clients to know which messages have been downloaded

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
+++ b/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
@@ -2,12 +2,16 @@
 # that must be retained for functionality and compatibility reasons 
 /^From:/            DUNNO
 /^Message-Id:/      DUNNO
-/^Chat-Is-Post-Message:/      DUNNO
+/^Chat-/            DUNNO
 /^Content-Type:/    DUNNO
 
 # For receiving clear-text messages (still supported in May 2026)
 /^Subject:/         DUNNO
 /^Date:/            DUNNO
+/^To:/              DUNNO
+/^CC:/              DUNNO
+/^References:/      DUNNO
+/^In-Reply-To:/     DUNNO
 
 # Senders might support Autocrypt 1 but not RFC9788 (Header Protection) 
 /^Autocrypt:/            DUNNO
--- a/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
+++ b/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
@@ -1,7 +1,23 @@
+# List of headers for incoming messages 
+# that must be retained for functionality and compatibility reasons 
 /^From:/            DUNNO
 /^Message-Id:/      DUNNO
 /^Chat-Is-Post-Message:/      DUNNO
 /^Content-Type:/    DUNNO
+
+# For receiving clear-text messages (still supported in May 2026)
 /^Subject:/         DUNNO
 /^Date:/            DUNNO
+
+# Senders might support Autocrypt 1 but not RFC9788 (Header Protection) 
+/^Autocrypt:/            DUNNO
+
+# SecureJoin V2 protocol headers (for backward compatibility) 
+/^Secure-Join:/             DUNNO
+/^Secure-Join-Invitenumber:/ DUNNO
+/^Secure-Join-Auth:/        DUNNO
+/^Secure-Join-Fingerprint:/ DUNNO
+/^Secure-Join-Group:/       DUNNO
+
+# Ignore all other headers
 /.*/                IGNORE
--- a/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
+++ b/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
@@ -1,3 +1,7 @@
-/^DKIM-Signature:/            IGNORE
-/^Authentication-Results:/            IGNORE
-/^Received:/            IGNORE
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-Is-Post-Message:/      DUNNO
+/^Content-Type:/    DUNNO
+/^Subject:/         DUNNO
+/^Date:/            DUNNO
+/.*/                IGNORE
--- a/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
+++ b/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
@@ -14,14 +14,10 @@
 /^In-Reply-To:/     DUNNO
 
 # Senders might support Autocrypt 1 but not RFC9788 (Header Protection) 
-/^Autocrypt:/            DUNNO
+/^Autocrypt:/       DUNNO
 
 # SecureJoin V2 protocol headers (for backward compatibility) 
-/^Secure-Join:/             DUNNO
-/^Secure-Join-Invitenumber:/ DUNNO
-/^Secure-Join-Auth:/        DUNNO
-/^Secure-Join-Fingerprint:/ DUNNO
-/^Secure-Join-Group:/       DUNNO
+/^Secure-Join/      DUNNO
 
 # Ignore all other headers
 /.*/                IGNORE


missytake:
We should deploy this on ci-chatmail.testrun.org before merging, to run core CI against it. @link2xt do we need to do anything for this other than `cmdeploy run`?

link2xt:
This will break message previews in the new core because of unencrypted `Chat-Is-Post-Message`.

feld:
I don't want to merge this until several core members are very confident this will not break anything. We can just continue iterating and thinking about whether the test coverage we have is sufficient and what the downsides may be as we continue the journey to remove all possible plaintext metadata from relay storage.

Does anyone know why we need the "From" header?

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -1,3 +1,4 @@
-/^DKIM-Signature:/            IGNORE
-/^Authentication-Results:/            IGNORE
-/^Received:/            IGNORE
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-Is-Post-Message:/      DUNNO
```
hpk42: Content-Type is also needed.

hpk42:
could you add content-type and rebase, so we can see what fails in the cmlxc runs?

feld:
> could you add content-type and rebase, so we can see what fails in the cmlxc runs?

done, let's see how it goes

feld:
>  assert 'Subject: [...]' in

tests are going to fail due to this, so I'll add Subject back in to see if it gets further. Maybe a "refute Subject in..." will be appropriate if everything else works

hpk42 (commented):
i think it's worthwhile to go for this change of having an explicit allow-list of headers, instead of blocking all unwanted ones.  But i'd be more in a "if in doubt, allow it" mode for now.  In particular, i think that Subject and Date better remain allowed for now, to not disrupt cleartext receival as a side effect of this allowlist approach.

feld:
I've just pushed an update with Date also retained

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -1,3 +1,7 @@
...
+/^Content-Type:/    DUNNO
+/^Subject:/         DUNNO
+/^Date:/            DUNNO
+/.*/                IGNORE
```
hpk42, line 23:
```suggestion
# List of headers for incoming messages 
# that must be retained for functionality and compatibility reasons 
/^From:/                     DUNNO
/^Message-Id:/               DUNNO
/^Content-Type:/             DUNNO

# Reserve a namespace in case chatmail introduces headers like 
# recently the Chat-Is-Post-Message one. 
/^Chat-/                     DUNNO

# Legacy headers not needed by current chatmail core (May 2026)
# Relays still have some support for delivering cleartext messages
/^Subject:/                  DUNNO
/^Date:/                     DUNNO
/^References:/       DUNNO 
/^In-Reply-To:/       DUNNO 
/^To:/                       DUNNO
/^Cc:/                       DUNNO

# Senders might support Autocrypt 1 but not RFC 9788 (Header Protection) 
/^Autocrypt:/                DUNNO

# For receiving automated replies 
/^Auto-Submitted:/           DUNNO 

# Legacy SecureJoin V2 headers, not used with V3 (released March 2026)
/^Secure-Join:/              DUNNO
/^Secure-Join-Invitenumber:/ DUNNO
/^Secure-Join-Auth:/         DUNNO
/^Secure-Join-Fingerprint:/  DUNNO
/^Secure-Join-Group:/        DUNNO
# Ignore all other headers
/.*/                         IGNORE
```

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -1,3 +1,7 @@
-/^DKIM-Signature:/            IGNORE
-/^Authentication-Results:/            IGNORE
-/^Received:/            IGNORE
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-Is-Post-Message:/      DUNNO
```
link2xt: If we introduced this change earlier, we would not have been able to add `Chat-Is-Post-Message`. So if we will need another header later, we will not be able to do it and have to put new data into some existing header.

feld: could we preserve one header like `Chat-Meta` and allow it to have a list of values?

Chat-Meta: Is-Post-Message, Something-Else

Or would that approach not be compatible?

link2xt: Just Chat- is fine.

If the PR is about removing metadata like IP addresses that spam filters and similar software on outgoing hosts may insert, it is unlikely that they will add any Chat- headers.

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -1,3 +1,23 @@
...
+# List of headers for incoming messages 
+# that must be retained for functionality and compatibility reasons 
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-Is-Post-Message:/      DUNNO
+/^Content-Type:/    DUNNO
+
+# For receiving clear-text messages (still supported in May 2026)
+/^Subject:/         DUNNO
+/^Date:/            DUNNO
```
hpk42, line 10: To, CC, References and in-Reply-To is also neededi think. 
see https://github.com/chatmail/relay/pull/816/changes#r3238110239

hpk42: i think you didn't take the latest version opf my suggestion, or maybbe github messed up.

feld: I just clicked the button to accept your changes. I think Github failed here...

feld:
Should we do the same for Secure Join and allow `/^Secure-Join` so the entire namespace is preserved as well?

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -1,3 +1,27 @@
...
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-/            DUNNO
+/^Content-Type:/    DUNNO
+
+# For receiving clear-text messages (still supported in May 2026)
+/^Subject:/         DUNNO
+/^Date:/            DUNNO
+/^To:/              DUNNO
+/^CC:/              DUNNO
```
link2xt, line 12: Are these checks case-sensitive? Is `Cc:` accepted?

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -1,3 +1,27 @@
...
-/^Authentication-Results:/            IGNORE
-/^Received:/            IGNORE
+# List of headers for incoming messages 
+# that must be retained for functionality and compatibility reasons 
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-/            DUNNO
+/^Content-Type:/    DUNNO
+
+# For receiving clear-text messages (still supported in May 2026)
```
link2xt, line 8: This is not only for cleartext, but any incoming encrypted messages that don't have header protection for these headers.

link2xt:
> Should we do the same for Secure Join and allow `/^Secure-Join` so the entire namespace is preserved as well?

Yes, I think it makes sense to do the same as with `Chat-` instead of listing the headers one by one.

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
hpk42: yes, header comparisons are generally case-insensitive, and postfix defaults to case-insensitive.  you have to do something special like "/^xal:/i    DUNNO" if you want to have case sensitive IIRC.

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
hpk42: yes, maybe best if you suggest a wording? 

sidenote:  ultimately, end-to-end encryption should require header protection, and anything else is roughly considered cleartext, as are messages today that are only encrypted and not signed.   we don't need to cargo-carry forward some Autocrypt L1 scope from 8 years ago.

link2xt:
There is also a comment in master.cf.j2 that needs an update:
https://github.com/chatmail/relay/blob/39d1ecaa03b994eb2fc9e58dc42bfd3f20ef4dba/cmdeploy/src/cmdeploy/postfix/master.cf.j2#L94-L96

With this change we will cleanup Received headers for incoming mails, and it is not a problem because DKIM is checked now in filtermail-incoming.

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
feld: Do we want to reword anything here or is this understood?

link2xt: I'd suggest something like "For messages that don't use protected headers (RFC 9788)" and group it together with the "Autocrypt" header below.

For example Thunderbird supports Autocrypt and is not likely to support protected headers soon. There is some legacy header protection support but mostly for Subject, and on the sender side I think they put Autocrypt only in the unprotected part. I know it's not really interesting to chat with Thunderbird, but we try to keep interoperability with Thunderbird and Proton Mail just on the level of being able to import public key and exchange messages in 1:1 chat, mostly to support OpenPGP as a standard and not ending up doing our own entirely incompatible client.

From this point of view not dropping `Date` and `Autocrypt` is mostly for the same reason.

link2xt (approved):

PR #820, opened 2026-01-22
fix(mtail): separate metrics for incoming and outgoing messages

link2xt:


cmdeploy/src/cmdeploy/mtail/delivered_mail.mtail
```diff
@@ -44,20 +44,36 @@ counter warning_count
...
+
+counter incoming_unencrypted_mail_count
+/Incoming: Filtering unencrypted mail\./ {
+  incoming_unencrypted_mail_count++
+  filtered_incoming_mail_count++
+}
+
+
 counter rejected_unencrypted_mail_count
 /Rejected unencrypted mail\./ {
```
j-g00da: Can we also fix this one? In rs version I include address from.
Just removing the dot should work.

PR #824, opened 2026-01-26
docs: fix link to filtermail

j-g00da:
Maybe we should remove it completely (?), but I think this provides some more clarity over just removing it from docs.

### DIFF
--- a/doc/source/overview.rst
+++ b/doc/source/overview.rst
@@ -42,6 +42,11 @@ The deployed system components of a chatmail relay are:
 -  Dovecot_ is the Mail Delivery Agent (MDA) and
    stores messages for users until they download them
 
+-  `filtermail <https://github.com/chatmail/filtermail>`_
+   prevents unencrypted email from leaving or entering the chatmail
+   service and is integrated into Postfix’s outbound and inbound mail
+   pipelines.
+
 -  Nginx_ shows the web page with privacy policy and additional information
 
 -  `acmetool <https://hlandau.github.io/acmetool/>`_ manages TLS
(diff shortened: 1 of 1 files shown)


doc/source/overview.rst
```diff
@@ -85,8 +85,9 @@ short overview of ``chatmaild`` services:
    <https://doc.dovecot.org/2.3/configuration_manual/authentication/dict/#complete-example-for-authenticating-via-a-unix-socket>`_
    to authenticate logins.
 
--  `filtermail <https://github.com/chatmail/relay/blob/main/chatmaild/src/chatmaild/filtermail.py>`_
-   prevents unencrypted email from leaving or entering the chatmail
+-  `filtermail <https://github.com/chatmail/filtermail>`_
+   no longer a part of chatmaild, development moved to a separate repository.
+   Prevents unencrypted email from leaving or entering the chatmail
```
missytake:
```suggestion
   prevents unencrypted email from leaving or entering the chatmail
```

This is sufficient :) we don't need the history in the docs.

j-g00da: The problem with this is that it will suggest that filtermail is included in chatmaild, since this is under "chatmaild implements various systemd-controlled services" section

missytake: Then let's move the line up `cmdeploy`, before nginx? https://github.com/chatmail/relay/pull/824/files#diff-c2f128f8f9aa4888a83d2b4074d395e492cdd73e89ef7e21aec3b5fd12b4a9b1R45

PR #829, opened 2026-01-29
Enhance getting started guide with package installation

timursagdenov:
Updated instructions to include package installation step.


ccclxxiii:
@timursagdenov thanks, this is helpful

ccclxxiii:
@hpk42 @missytake

ccclxxiii:
> Yeah, it's not perfect that this assumes a Debian-derivative for the local PC, but if people run something special that's on them.
> 
> Maybe we should just re-frame the whole getting started section so it is executed on the server, not on the local PC.

i'll start working on this as mentioned in team chat.  for now i will take notes from this pr, and combine it with document enhancements.  @timursagdenov thanks for your contribution, ill take in what you wrote and add it in new pr.

PR #831, opened 2026-01-30
feat: replace DKIM verification with filtermail v0.5

j-g00da:
Upgrade to filtermail v0.5, which has a built-in DKIM verifier
and disable OpenDKIM on reinject_incoming.


chatmaild/src/chatmaild/tests/test_filtermail_blackbox.py
```diff
@@ -58,6 +60,9 @@ def test_one_mail(
 
     popen = make_popen(["filtermail", path, filtermail_mode])
     line = popen.stderr.readline().strip()
+    if b"WARN" in line:
```
missytake: Can we look for the specific warning? There might be other warnings which would be helpful in the pytest output.

cmdeploy/src/cmdeploy/opendkim/deployer.py
```diff
@@ -37,24 +37,6 @@ def configure(self):
...
-        )
-        need_restart |= screen_script.changed
-
-        final_script = files.put(
-            src=get_resource("opendkim/final.lua"),
-            dest="/etc/opendkim/final.lua",
-            user="root",
-            group="root",
-            mode="644",
-        )
```
missytake, line 55: These calls should not be removed, but include a `present=False` line.

link2xt:
Failing testcase is at https://github.com/chatmail/filtermail/pull/52

j-g00da:
> Unfortunately we got some invalid rejections

Solved in https://github.com/chatmail/filtermail/pull/53

Emails that failed DKIM verification are now stored in `/tmp/filtermail-rejected/dkim-verify/<timestamp>.eml`
(https://github.com/chatmail/filtermail/pull/55)

Both included in filtermail 0.5

j-g00da:
Further edge cases addressed in filtermail 0.5.1 (https://github.com/chatmail/filtermail/pull/62 and https://github.com/chatmail/filtermail/pull/61).

Ready again :)

missytake:
> Emails that failed DKIM verification are now stored in `/tmp/filtermail-rejected/dkim-verify/<timestamp>.eml` ([chatmail/filtermail#55](https://github.com/chatmail/filtermail/pull/55))

This seems to work, but it would be much less annoying if it would be saved to `/tmp/filtermail-rejected/dkim-verify/<domain>.eml` and would overwrite older files. The latest report per domain is enough to debug it, and then you only have to delete one file if it works.

missytake (approved):

link2xt:
> This seems to work, but it would be much less annoying if it would be saved to /tmp/filtermail-rejected/dkim-verify/<domain>.eml and would overwrite older files. The latest report per domain is enough to debug it, and then you only have to delete one file if it works.

There are now some false-positive mailer-daemon messages that will easily overwrite a really problematic mail because #873 
I think it is good enough as is. We may want to setup tmpfiles.d to cleanup old files after some time though.

Issue #738, opened 2025-11-21
Fix ipv4-only for unbound


feld:
It's possible to deploy a Chatmail relay server that misbehaves because it tries to deliver to mail servers that have an AAAA and just fails. We should control sysctls and guard any setting for Postfix/Dovecot in the templates that enable IPv6 functionality 

edit: this is a rough placeholder issue as the reporter's GitHub account was banned and they need to make a new one. More information will be coming.

link2xt:
It is currently opt-out, but there is a setting to disable it:
https://github.com/chatmail/relay/blob/8f68672e3112682055b8fdb6e0e8d6c473730f8c/chatmaild/src/chatmaild/ini/chatmail.ini.f#L63-L64

missytake:
I'd like to close this issue for now, we can re-open it if the original poster re-appears and gives more information. Personally, I'm sceptical to make IPv6 opt-in, and opt-out is already supported - so to keep this open we need more arguments. And if it was actually a problem with `disable_ipv6 = True`, we need more information what went wrong.

ccclxxiii:
@missytake here is the core issue i see.  

when unbound deploys by default it will listen on `ipv4` and `ipv6`.  the machine it was deployed to couldn't bind unbound to the interface since the expectation was to have `ipv6` enabled.  this caused the `cmdeploy run` to fail with a systemd journal entry noting it could not bind the interface.

as a patch i created `/etc/unbound/unbound.conf.d/custom-interfaces.conf`

as:

````
server:
    interface: 127.0.0.1
    do-ip6: no
    interface-automatic: no
````

sighup'd unbound, and then re-deployed and it worked.

it seems the case for `disable_ipv6 = True` has no effect on the unbound service whether `True || False`

as for postfix, i've deployed a new relay and it failed without changing `mynetworks` in `/etc/postfix/main.cf`

`mynetworks = 127.0.0.0/8`

py infra will overwrite this each time you perform a `cmdeploy run` so it may be something we consider with `.j2` templating for `postfix` and `unbound`  

likely a pr i can make if there's interest in fixing it.

cc: @feld

adbenitez:
that was also my experience when deploying chatmail in a server that had IPv6 disabled by default (Contabo), even if I was providing the `disable_ipv6=True` in `chatmail.ini` this was still failing

ccclxxiii:
fixed in #832

PR #832, opened 2026-01-30
remediates ipv6 boolean not being respected during operations

ccclxxiii:
pr overview:
- addresses #738 
- implements improved templating for `unbound` `postfix` and `dovecot` service units  
- when `disable_ipv6 = False` (default behavior), service interfaces listen on both ipv4/v6  
- when `disable_ipv6 = True` (non-default behavior), service interfaces are forced to listen on ipv4 address which is desired behavior  
- this assumes the operator either has no ipv6 interface available or has disabled it themselves in `/etc/sysctl.conf`  
- this has been smoke-tested manually on an existing relay, and both positive/negative cases have been smoke-tested on new relay instances with a positive outcome

### DIFF
--- a/cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
+++ b/cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2
@@ -1,7 +1,7 @@
 ## Dovecot configuration file
 
 {% if disable_ipv6 %}
-listen = *
+listen = 0.0.0.0
 {% endif %}
 
 protocols = imap lmtp


ccclxxiii:
> Looks good :) CI is happy as well, so it's not only smoke tests :)
> 
> ~Although I don't fully remember what network protocols github actions support - I think last time I looked the runners can't do IPv6 anyway 🙃~ edit: nevermind, staging-ipv4.testrun.org is a Hetzner machine without an IPv6:
> image

thanks @missytake !

Issue #833, opened 2026-01-30
Resolv.conf experiences improper resolver injection


ccclxxiii:
in deploying a new relay (provider is racknerd, but has also happened on cloudfanatic and hetzner)
i have seen `/etc/resolv.conf` become populated as

````
# Generated by SolusVM
nameserver 8.8.8.8
nameserver 8.8.4.4nameserver 9.9.9.9
````

this causes name resolution to fail and the deploy sequence to bail out.

- server: `debian 12`  
- osexec: `debian 12`  
- version-hash: `e42139e97b85abb7273ff073711f5a849f39530f` 
 
## Expected behavior

- deployment should continue without exceptions raised for bad ns reoslution

## Actual behavior

- `/etc/resolv.conf` is populated with an improper concat at times which results in

````
# Generated by SolusVM
nameserver 8.8.8.8
nameserver 8.8.4.4nameserver 9.9.9.9
````

PR #834, opened 2026-01-30
fix: remediates issue with improper concat on resolver injection

ccclxxiii:
pr overview:
- addresses issue #833 
- in some hosting providers when we inject `9.9.9.9` from `cmdeploy/src/cmdeploy/deployers.py` the injection of `9.9.9.9` can have an improper concat which causes name resolution to fail and the deploy to bail out
- this branch has been tested on racknerd and cloudfanatic where i was experiencing the problem
- full end to end smoke test on clean and existing relays demonstrates the behavior is corrected

----
before (current main branch)

````
# Generated by SolusVM
nameserver 8.8.8.8
nameserver 8.8.4.4nameserver 9.9.9.9
````

after (373/fix-dns-resolver-injection branch)

````
# Generated by SolusVM
nameserver 8.8.8.8
nameserver 8.8.4.4
nameserver 9.9.9.9
````

### DIFF
--- a/cmdeploy/src/cmdeploy/deployers.py
+++ b/cmdeploy/src/cmdeploy/deployers.py
@@ -550,22 +550,11 @@ def deploy_chatmail(config_path: Path, disable_mail: bool, website_only: bool) -
         return
 
     if host.get_fact(Port, port=53) != "unbound":
-        server.shell(
-            name="Ensure resolv.conf ends with newline",
-            commands=[
-                "python3 - <<'PY'\n"
-                "from pathlib import Path\n"
-                "path = Path('/etc/resolv.conf')\n"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmdeploy/src/cmdeploy/deployers.py
Comment (missytake) [inline cmdeploy/src/cmdeploy/deployers.py]:
```suggestion
            line="\nnameserver 9.9.9.9",
```

phew, isn't this a bit easier?

ccclxxiii: @missytake good callout, was starting to follow some other pattern, let me refactor

ccclxxiii:
@missytake feedback addressed in `c0718325ef911b9d04c19ed470b4a8e47b46c52e`

missytake (approved):

PR #836, opened 2026-02-07
fix: multiple bug fixes across chatmaild and cmdeploy

Retengart:
### Bug fixes (round 1)
- **fsreport.py**: fix `AttributeError` — `FileEntry` has `path`, not `relpath`. Crashes when `--mdir` flag is used
- **rdns.py**: fix `TypeError` — `get_dkim_entry` returns bare `None` instead of `(None, None)` tuple on `CalledProcessError`, crashing `cmdeploy dns` on fresh servers
- **deployers.py**: fix silent deploy failure — `build_webpages` returns `None` on exception, causing `rsync("None/", ...)`
- **cmdeploy.py**: remove dead code and potential `NameError` in `run_cmd` — `check_call` always returns 0 or raises, making `retcode != 0` branches unreachable; `remote_data` undefined with `--skip-dns-check`
- **rshell.py**: fix `IndexError` in `dovecot_recalc_quota` — empty line at end of `doveadm` output produces `parts=[]`

### Security fixes (round 2)
- **newemail.py**: use `secrets.choice` instead of `random.choices` for username generation — per Python docs, `secrets` should be used for security-sensitive data
- **nginx.conf.j2**: remove deprecated TLS 1.0/1.1 — RFC 8996, aligns with postfix (≥1.2) and dovecot (1.3) in same stack
- **doveauth.py**: validate localpart characters to `[a-z0-9._-]` — prevents filesystem issues from crafted usernames via IMAP/SMTP auth
- **doveauth.py**: fix TOCTOU race in account creation — use `filelock` to serialize concurrent creation for same address
- **turnserver.py**: add 5s socket timeout — hung TURN daemon would block dict proxy thread indefinitely
- **metadata.py**: handle `turn_credentials()` exceptions — return `N` (not found) instead of crashing the dict proxy connection

### Formatting
- **chore**: ruff formatting fixes in acmetool, dovecot, postfix deployers

## Test plan

- [x] `pytest chatmaild/src/` — 50 passed
- [x] `pytest cmdeploy/src/cmdeploy/tests/ --ignore=online` — 16 passed
- [x] `ruff check` + `ruff format --check` on all changed Python files — clean
- [x] CI staging deploy + online tests

### DIFF
--- a/chatmaild/src/chatmaild/doveauth.py
+++ b/chatmaild/src/chatmaild/doveauth.py
@@ -149,12 +149,16 @@ class AuthDictProxy(DictProxy):
             return
 
         lock = filelock.FileLock(str(user.password_path) + ".lock", timeout=5)
-        with lock:
-            userdata = user.get_userdb_dict()
-            if userdata:
-                return userdata
-            user.set_password(encrypt_password(cleartext_password))
-            print(f"Created address: {addr}", file=sys.stderr)
+        try:
+            with lock:
+                userdata = user.get_userdb_dict()
+                if userdata:
+                    return userdata
+                user.set_password(encrypt_password(cleartext_password))
+                print(f"Created address: {addr}", file=sys.stderr)
+        except filelock.Timeout:
+            logging.warning("lock timeout creating %s, skipping", addr)
+            return
         return user.get_userdb_dict()
 
 
--- a/cmdeploy/src/cmdeploy/tests/test_rshell.py
+++ b/cmdeploy/src/cmdeploy/tests/test_rshell.py
@@ -20,10 +20,8 @@ def test_dovecot_recalc_quota_normal_output():
 
 def test_dovecot_recalc_quota_empty_output():
     """Empty doveadm output (trailing newline) must not IndexError."""
-    call_count = [0]
 
     def mock_shell(cmd):
-        call_count[0] += 1
         if "recalc" in cmd:
             return ""
         # quota get returns only empty lines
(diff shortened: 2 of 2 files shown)


missytake:
Thanks for this giant PR! I will create a separate branch so the CI can deploy it to our test servers.

Usually we like to have one PR per change, so we can decide about them individually^^ let's see, if all of it is uncontroversial, it's probably fine, and at least the commits are properly separated :) I will take a look at your commits later today (or latest, on Tuesday).

cmdeploy/src/cmdeploy/tests/test_dns.py
```diff
@@ -60,6 +60,29 @@ def mockdns(request, mockdns_base, mockdns_expected):
...
+        """CalledProcessError must return (None, None), not bare None."""
+        from subprocess import CalledProcessError
+
+        def failing_shell(command, fail_ok=False, print=print):
+            raise CalledProcessError(1, command)
+
+        monkeypatch.setattr(remote.rdns, "shell", failing_shell)
+        result = remote.rdns.get_dkim_entry("some.domain", "", dkim_selector="opendkim")
+        assert result == (None, None)
+        assert result[0] is None and result[1] is None
```
missytake: Out of interest: Is there a scenario in which the first statement passes and the second doesn't?

Retengart: I guess I overlooked it... Of course, you're right — it's redundant. `None` is a singleton so `== None` and `is None` are equivalent. Removed in the next push.

Retengart:
Two follow-up fixes based on self-review:

1. **`doveauth.py`**: `filelock.Timeout` was uncaught — if the lock couldn't be acquired in 5s, the exception would propagate to DictProxy's generic handler and kill the entire Dovecot dict connection (affecting all in-flight requests). Now caught locally: logs a warning and returns `N` (not found) so only the single lookup fails gracefully and the client can retry.

2. **`test_rshell.py`**: removed unused `call_count` variables from 3 tests (declared and incremented but never asserted).

missytake:
I'll rebase it on main in a separate branch, so we can run CI and merge it. Hopefully I get to it later today :)

missytake:
merged in #883 :)

PR #840, opened 2026-02-09
feat: metadata service: make turnserver socket path configurable

### LINKED ISSUE #779: chatmaild turnserver hardcodes the socket path
https://github.com/chatmail/relay/blob/1188aed0616744c34568882194790bf6f9784dc9/chatmaild/src/chatmaild/turnserver.py#L7

Can we make this configurable? Thanks!

feld:
also add tests for the turnserver metadata

Fixes #779

### DIFF
--- a/chatmaild/src/chatmaild/metadata.py
+++ b/chatmaild/src/chatmaild/metadata.py
@@ -79,12 +79,13 @@ class Metadata:
 
 
 class MetadataDictProxy(DictProxy):
-    def __init__(self, notifier, metadata, iroh_relay=None, turn_hostname=None):
+    def __init__(self, notifier, metadata, iroh_relay=None, turn_hostname=None, turn_socket_path=None):
         super().__init__()
         self.notifier = notifier
         self.metadata = metadata
         self.iroh_relay = iroh_relay
         self.turn_hostname = turn_hostname
+        self.turn_socket_path = turn_socket_path
 
     def handle_lookup(self, parts):
         # Lpriv/43f5f508a7ea0366dff30200c15250e3/devicetoken\tlkj123poi@c2.testrun.org
@@ -101,7 +102,7 @@ class MetadataDictProxy(DictProxy):
                             return f"O{self.iroh_relay}\n"
                         case "turn":
                             try:
-                                res = turn_credentials()
+                                res = turn_credentials(self.turn_socket_path)
                             except Exception:
                                 logging.exception("failed to get TURN credentials")
                                 return "N\n"
@@ -135,6 +136,7 @@ def main():
     config = read_config(config_path)
     iroh_relay = config.iroh_relay
     mail_domain = config.mail_domain
+    socket_path = config.turn_socket_path
 
     vmail_dir = config.mailboxes_dir
     if not vmail_dir.exists():
@@ -152,6 +154,7 @@ def main():
         metadata=metadata,
         iroh_relay=iroh_relay,
         turn_hostname=mail_domain,
+        turn_socket_path=socket_path,
     )
 
     dictproxy.serve_forever_from_socket(socket)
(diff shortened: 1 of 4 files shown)


feld:
There were no real tests for the turnserver metadata stuff that I could see, so I included that, but I have no real experience with python test framework so that part was vibed but it looks like how I'd do things in other languages.

If someone with actual experience in Python tests could review it that would be great

chatmaild/src/chatmaild/tests/test_turnserver.py
```diff
@@ -0,0 +1,120 @@
...
+    # Create a config with custom socket path
+    config_path = temp_dir / "chatmail.ini"
+    write_initial_config(config_path, "test.example.org", {
+        "turn_socket_path": str(temp_socket_path)
+    })
+    config = read_config(config_path)
+
+    # Allow time for server to start
+    import time
+    time.sleep(0.01)
```
adbenitez: this doesn't look like a serious way to wait for a service to start, might cause flaky tests 🤔

adbenitez (approved):
lgtm, I didn't test but the change is simple and straightforward: just adding a new configuration option for a path instead of using a hard-coded value, can't say much about the tests tho better if someone with more deep understanding of the current tests machinery also take  a look at it, the use of sleep give me a bad feeling, I didn't check if other tests also do this

missytake (commented):
I'm slightly conflicted about this PR.

The additional config option looks innocent enough, it's a bit annoying to add new config options for small things which operators are only expected to edit with alternative deployment options, but fine with me if the chef cookbook needs this. I think my preference is saying "please rebase this on top of https://github.com/chatmail/relay/pull/853 so we don't have to put this option into the default-generated chatmail.ini, and can just set a default in config.py for everyone who doesn't need to think about this".

But then there is also the LLM-generated tests; you don't understand what they do, I don't understand what they do, they have some sketchy sleep statement in it... If they were in a separate PR, this one would be long approved.

Also, please fix the lint errors :) https://github.com/chatmail/relay/actions/runs/22113484762/job/63915452488?pr=840

chatmaild/src/chatmaild/metadata.py
```diff
@@ -76,12 +76,13 @@ def get_tokens_for_addr(self, addr):
...
 
 class MetadataDictProxy(DictProxy):
-    def __init__(self, notifier, metadata, iroh_relay=None, turn_hostname=None):
+    def __init__(self, notifier, metadata, iroh_relay=None, turn_hostname=None, config=None):
         super().__init__()
         self.notifier = notifier
         self.metadata = metadata
         self.iroh_relay = iroh_relay
         self.turn_hostname = turn_hostname
+        self.config = config
```
hpk42: we avoid passing config around in favor of explicit values. 
in this case it's the turn_socket_path that should be passed in.

chatmaild/src/chatmaild/tests/test_turnserver.py
```diff
@@ -0,0 +1,120 @@
+"""Tests for turnserver functionality, particularly metadata integration."""
```
hpk42: i think it's fine to just do offline config tests (that the config setting exists and has a proper default) and not try to test turn_credentials with mock servers etc. for now. it didn't have these tests before.

hpk42 (changes requested):
thanks. please simplify the PR along the review comments.

chatmaild/src/chatmaild/tests/test_turnserver.py
missytake: There are now tests for turnserver functionality here: https://github.com/chatmail/relay/pull/883/changes/835b5e77066348ed577dcbf705fb934f272c5e5c#diff-fc3c6a3f43aaa8266eb10ce6dad7f018493f151fca34cecea33318d6ce719afa

Maybe we can take the turnserver tests out of this PR and use those instead?

ccclxxiii:
@feld looks like merge conflicts you may want to remediate, and bring in latest default into trunk

feld:
rebased, simplified. It looks right at first glance but needs a real review.

hpk42:
i took your idea and cleaned up the turnserver tests see https://github.com/chatmail/relay/pull/968

hpk42:
k, https://github.com/chatmail/relay/pull/968 is merged.  If you need further discussion, feel free to chat me up :)

PR #841, opened 2026-02-10
fix(dovecot): install also if dovecot.service=False in SystemdEnabled Fact

missytake:
Somehow, installing dovecot was skipped during our migration; after we set this check to True, it at least tried installing dovecot.

### DIFF
--- a/cmdeploy/src/cmdeploy/dovecot/deployer.py
+++ b/cmdeploy/src/cmdeploy/dovecot/deployer.py
@@ -22,7 +22,7 @@ class DovecotDeployer(Deployer):
 
     def install(self):
         arch = host.get_fact(Arch)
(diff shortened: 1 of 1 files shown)


j4n (approved):

missytake:
This will lead to dovecot being re-installed when it is disabled, though. Not sure if that is a problem

j4n:
Merging this, though I do think this check is harmful still as it prevents updates from being installed, I will propose a better guard separately.

missytake:
I found out why this is a problem - when upgrading dovecot, it somehow tries to restart the dovecot systemd service, even if it was *disabled* before. This makes cmdeploy fail if there is no cert yet, e.g. because acmetool failed on the first try.

PR #845, opened 2026-02-11
feat(postfix): enforce sending from public IP addresses

j4n:
Add chatmail.ini settings for smtp_bind_address, smtp_bind_address6, and smtp_bind_address_enforce to configure the outbound postfix address.


link2xt:
This is too low-level, basically 3 options for 3 options in postfix that you can configure to whatever. Can as well ask users to write postfix config addition.

We should already know everything needed because we generate zone file and check A and AAAA records, no need to fill in the same IP addresses manually.

missytake:
As it looks like now, users in NATed environments don't need to disable this even, they probably run with --skip-dns-check anyway.

j4n:
Fixing an unrelated bug, I added a check for --skip-dns for the acme config line, that broke when running with that flag and the deploy fails (second commit).

PR #846, opened 2026-02-11
postfix: accept self-signed certificates for IP-only relays

missytake:
Together with omitting DKIM checks for IP-only relays that should make us interoperable with madmail.


missytake:
Bad news - this doesn't work. I think the only way to federate would be dropping TLS verification for all recipients, or better, delivering via opportunistic HTTPS in filtermail instead of SMTP.

cmdeploy/src/cmdeploy/postfix/smtp_tls_policy_map
```diff
@@ -0,0 +1,2 @@
+/^\[([0-9]{1,3}\.){3}[0-9]{1,3}\]$/ encrypt
+/nauta\.cu/    may
```
link2xt: Probably should be `/^nauta\.cu$/` for exact match.

cmdeploy/src/cmdeploy/postfix/smtp_tls_policy_map
```diff
@@ -0,0 +1,2 @@
+/^\[([0-9]{1,3}\.){3}[0-9]{1,3}\]$/ encrypt
```
link2xt: Could it be `/^\[[^]]+\]$/` for simplicity and to also support IPv6?

missytake: image

Found a slightly better version which doesn't match for domains enclosed in brackets, but should accept IPv6.

link2xt: What is the problem with matching `[example.org]`? You will not be able to even establish a TCP connection to it anyway.

missytake: hm, point. Still, this makes it slightly clearer what's the purpose of these exceptions...

j-g00da: Syntax for ipv6 domain literals is `[IPv6:<ipv6>]` (with a prefix) so it wouldn't match these

missytake: ah thanks, good catch. That makes it even easier to see what's the purpose of the regex

link2xt: I still don't understand why this need to be more complicated than checking for `[` and `]`, square brackets are only allowed for address literals: https://www.rfc-editor.org/rfc/rfc5321#section-4.1.3
Now we already discovered one bug.

At the same time second regexp (apparently intended for IPv4 literals) matches pretty much everything, e.g. `echo '[IPv6:2a01:4f9:fff1:59::0.0.0.1]' | awk '/^\[[^]]+\.[0-9]{1,3}\]$/ { print }'` (current IPv6 address of nine.testrun.org in  IPv6-comp notation), but requires that address ends with a dot and numbers.

cmdeploy/src/cmdeploy/postfix/smtp_tls_policy_map
```diff
@@ -0,0 +1,3 @@
+/^\[IPv6\:[^]]+\:[0-9,a-f]{0,4}\]$/ encrypt
```
link2xt: `:` does not need to be escaped.

cmdeploy/src/cmdeploy/postfix/smtp_tls_policy_map
missytake: alright, I'm convinced. I think to me different regexes are complicated than to you; but that's what commit messages are for, so should be fine.

link2xt:
I have tested by creating a profile on IP-based server, exporting "Me" vCard to saved messages, importing it into staging server profile and sending a message - it arrived.

j-g00da (approved):

PR #849, opened 2026-02-13
feat: Strip Received headers before delivery

feld:


feld:
I have been using this without issues

PR #852, opened 2026-02-16
feat(cmdeploy): add ssh_host chatmail.ini option to deploy remotely, recommend localhost deployment as default

missytake:
pyinfra, stemming from ansible, has a build machine / deployment server separation which we don't actually need in this project. We are mostly using it to script the installation, not to document our deployment; and I don't know of anyone importing cmdeploy as a pyinfra deployment module (other than https://github.com/deltachat/pyinfra-borgbackup/ for example).

Let's make it easier for future operators by simply assuming that you install cmdeploy on the relay itself. This also helps with people having weird laptop OSes.

The idea is to add a chatmail.ini config parameter, `ssh_host`, which will be set to "localhost" on `cmdeploy init` to be the default for new operators, but which will default to `mail_domain` if unset, so existing deployments keep working with their chatmail.ini files. This could become a problem if some is used to setup chatmail relays remotely without keeping the chatmail.ini file around, though.

### DIFF
--- a/chatmaild/src/chatmaild/ini/chatmail.ini.f
+++ b/chatmaild/src/chatmaild/ini/chatmail.ini.f
@@ -3,6 +3,9 @@
 # mail domain (MUST be set to fully qualified chat mail domain)
 mail_domain = {mail_domain}
 
+# Where to deploy the relay - if unspecified, mail_domain will be used.
+ssh_host = localhost
+
 #
 # If you only do private test deploys, you don't need to modify any settings below
 #
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -96,7 +96,7 @@ def _warn_unused_settings(unused_keys, out):
 def run_cmd(args, out):
     """Deploy chatmail services on the remote server."""
 
-    ssh_host = args.ssh_host if args.ssh_host else args.config.mail_domain_bare
+    ssh_host = args.ssh_host if args.ssh_host else args.config.ssh_host
     sshexec = get_sshexec(ssh_host)
     require_iroh = args.config.enable_iroh_relay
     strict_tls = args.config.tls_cert_mode == "acme"
@@ -116,7 +116,7 @@ def run_cmd(args, out):
     pyinf = "pyinfra --dry" if args.dry_run else "pyinfra"
 
     cmd = f"{pyinf} --ssh-user root {ssh_host} {deploy_path} -y"
-    if ssh_host == "localhost":
+    if ssh_host in ["localhost", "@local"]:
         cmd = f"{pyinf} @local {deploy_path} -y"
 
     if version.parse(pyinfra.__version__) < version.parse("3"):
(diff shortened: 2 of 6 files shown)


hpk42:
is this PR still needed?

missytake:
> is this PR still needed?

I still think it is a great improvement, but it's not a priority for the first half of May. I'll continue to work on it in a few weeks

missytake:
`cmlxc` fails, because the default now says to deploy in the builder container. Should we change cmlxc to deploy on the relay itself, or remove the `ssh_host` option in the generated chatmail.ini?

I'll proceed with hardcoding the `ssh_host` option to `mail_domain` for now.

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -3,6 +3,9 @@
 # mail domain (MUST be set to fully qualified chat mail domain)
 mail_domain = {mail_domain}
 
+# Where to deploy the relay - if unspecified, mail_domain will be used.
+ssh_host = localhost
```
j-g00da, line 7: I'm still a bit worried, this will confuse some people (although, it indeed makes more sense).

cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -116,7 +116,7 @@ def run_cmd(args, out):
     pyinf = "pyinfra --dry" if args.dry_run else "pyinfra"
 
     cmd = f"{pyinf} --ssh-user root {ssh_host} {deploy_path} -y"
-    if ssh_host == "localhost":
+    if ssh_host in ["localhost", "@local"]:
```
j-g00da, line 119: Do we rely on `@local` working here too? I would prefer to keep a single correct way to do this if we can.

doc/source/getting_started.rst
```diff
@@ -246,25 +242,3 @@ The deploy will verify that both files exist on the server.
...
-
-Run safety checks before a new deployment:
-
-::
-
-   ./scripts/cmdeploy dns
-   ./scripts/cmdeploy status
-
-If you keep multiple build machines (ie laptop and desktop), keep ``chatmail.ini`` in sync between
-them.
```
j-g00da, line 270: Maybe we can change this to a section about migrating from a separate build machine to running deploy scripts on relay itself?

j-g00da (approved):
LGTM, left some nits

chatmaild/src/chatmaild/ini/chatmail.ini.f
missytake: Yeah, I couldn't come up with a better path forward. If you have a better wording, let me work :)

cmdeploy/src/cmdeploy/cmdeploy.py
missytake: Both usages are in fluctuation I think, as `@local` is pyinfra convention.

chatmaild/src/chatmaild/ini/chatmail.ini.f
j-g00da: It's mostly confusing because mail_domain is the default... except it's not because it is also overwritten by default by localhost.

Maybe something like:
```suggestion
# Where to deploy the relay.
# Change the following line to deploy on a remote machine instead,
# or remove it to deploy on mail_domain.
```

Not sure :/

j4n (approved):
lgtm :) we've been deploying nine like this for a while and its just much faster in the very least.

missytake:
I started a small poll in the operators community to see if this breaks anyone's workflows or is viewed critically in any way.

missytake:
The change is not uncontroversial: 
image

Let's talk about it next week.

PR #853, opened 2026-02-17
feat(config): load default values from Config(), not chatmail.ini.f

missytake:
We should not rely on the chatmail.ini.f default config file for filling the empty values, but do this in the python3 Config() constructor. This enables https://github.com/chatmail/relay/pull/852 to have different default values for older chatmail.ini files than for more recent configs, and is generally more consistent.


chatmaild/src/chatmaild/config.py
```diff
@@ -23,16 +18,18 @@ def __init__(self, inipath, params):
...
-        self.username_max_length = int(params["username_max_length"])
-        self.password_min_length = int(params["password_min_length"])
-        self.passthrough_senders = params["passthrough_senders"].split()
-        self.passthrough_recipients = params["passthrough_recipients"].split()
+        self.max_mailbox_size = params.get("max_mailbox_size", "500M")
+        self.max_message_size = int(params.get("max_message_size", 31457280))
+        self.delete_mails_after = params.get("delete_mails_after", "20")
+        self.delete_large_after = params.get("delete_large_after", "7")
+        self.delete_inactive_users_after = int(
+            params.get("delete_inactive_users_after", 100)
```
j4n: this increased to 100 from 90

chatmaild/src/chatmaild/config.py
```diff
@@ -9,12 +9,7 @@
 def read_config(inipath):
     assert Path(inipath).exists(), inipath
     cfg = iniconfig.IniConfig(inipath)
-    params = cfg.sections["params"]
-    default_config_content = get_default_config_content(params["mail_domain"])
```
j4n, line 14: leaving this out means minimal chatmail.ini's won't get the testrun.org privacy override, but thats probably fine because we always deploy the default ini anyway.

j4n (approved):
lgtm; one issue I see is that now we have two places there are defaults; probably we should generate the chatmail.ini.f from the config.py defaults?

chatmaild/src/chatmaild/config.py
missytake: Right, that was a mistake. reducing it to 90.

missytake:
> lgtm; one issue I see is that now we have two places there are defaults; probably we should generate the chatmail.ini.f from the config.py defaults?

Hm, I think the better approach is commenting out the values in chatmail.ini.f, then the defaults in config.py count if you don't explicitly uncomment them. This is how many other config formats are doing it as well.

The commented out values still document the default values for new users, but if we forget to update them, nothing breaks. And it's clear to us that the commented out line isn't the meaningful one.

chatmaild/src/chatmaild/config.py
missytake: it isn't fine, it makes the CI fail :sweat_smile: 

but - true, we don't really need this privacy override anymore. nine is anyway deployed from a specific existing chatmail.ini file, and the other .testrun.org instances are test servers and don't need a privacy policy.

...and then we don't need the whole override mechanism. The only other point where it's used is when testing the deprecated `mailboxes_dir` config option.

But removing the config overrides is out of scope for this PR I guess <.< unless more people agree we could take it out here?

missytake: Nevermind, we need the overrides for more tests than just for testing the deprecated `mailboxes_dir` option. We can't get rid of it, maybe it can be simplified.

missytake: The `mailboxes_dir` option can also be tested without the testrun-specific overrides; I removed the testrun-specific overrides completely.

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -64,18 +64,18 @@
...
-filtermail_smtp_port = 10080
-postfix_reinject_port = 10025
+#filtermail_smtp_port = 10080
+#postfix_reinject_port = 10025
 
 # SMTP incoming filtermail and reinjection 
-filtermail_smtp_port_incoming = 10081
-postfix_reinject_port_incoming = 10026
+#filtermail_smtp_port_incoming = 10081
+#postfix_reinject_port_incoming = 10026
```
hpk42: i'd remove them entirely from chatmail.ini
filtermail_http_port_incoming and filtermail_lmtp_port_transport are already missing. 
If someone really wants to fiddle at this level, they need to read the source for config.py.

hpk42 (approved):
lgtm!  would remove the port configuration from chatmail.ini.f

PR #855, opened 2026-02-18
Introduce support for self-signed chatmail relays

hpk42:
This PR introduces support for running chatmail relays using self-signed TLS certificates by setting tls_cert = self in chatmail.ini. This is useful for testing purposes and deploying functional chatmail instances on private networks, local machines, or anywhere where obtaining a public domain and Let's Encrypt certificate is not readily available or feasible. 

Note that relays configured with self-signed certificates will only interoperate with other relays that also use opportunistic encryption (tls_cert = self); they will not be able to federate with standard chatmail relays using Let's Encrypt (tls_cert = acme) because the default configuration enforces strict certificate validation. Enabling interoperability between self-signed and signed relays would require relaxing default security settings, which is out of scope of this PR, and requires more thorough discussion. 

the commits can also be reviewed separately.

### DIFF
--- a/.github/workflows/test-and-deploy-ipv4only.yaml
+++ b/.github/workflows/test-and-deploy-ipv4only.yaml
@@ -87,7 +87,7 @@ jobs:
           ssh root@ns.testrun.org nsd-checkzone staging-ipv4.testrun.org /etc/nsd/staging-ipv4.testrun.org.zone
           ssh root@ns.testrun.org systemctl reload nsd
 
-      - name: cmdeploy test acme
+      - name: cmdeploy test
         run: CHATMAIL_DOMAIN2=ci-chatmail.testrun.org cmdeploy test --slow
 
       - name: cmdeploy dns


cmdeploy/src/cmdeploy/tests/online/test_0_qr.py
```diff
@@ -11,15 +11,28 @@ def test_gen_qr_png_data(maildomain):
...
     for i in range(3):
         account_id = rpc.add_account()
-        rpc.set_config_from_qr(account_id, url)
+        if chatmail_config.tls_cert == "self":
+            # deltachat core's rustls rejects self-signed HTTPS certs during
+            # set_config_from_qr, so fetch credentials via requests instead
+            verify = False
+            res = requests.post(f"https://{maildomain}/new", verify=verify)
+            data = res.json()
+            rpc.set_config(account_id, "addr", data["email"])
```
link2xt: This `set_config` followed up by `configure` way of setting up is deprecated. UIs are not using it already: https://github.com/deltachat/deltachat-android/pull/3676

New APIs are `add_transport_from_qr(account_id, url)` and `add_transport({"addr": ...})`.

www/src/dclogin.js
```diff
@@ -0,0 +1,19 @@
...
+    function generateProfile() {
+        fetch('/new')
+            .then(function(r) { return r.json(); })
+            .then(function(data) {
+                var url = data.dclogin_url;
+                var link = document.getElementById('dclogin-link');
+                link.href = url;
+                var container = document.getElementById('qr-code');
+                var qr = new QRCode({content: url, width: 300, height: 300, padding: 1, join: true});
+                container.innerHTML = '<a href="' + url + '">' + qr.svg() + '</a>';
```
link2xt: Better create another container just for SVG with `a` outside and only change `href`, so it does not break if URL has characters that should be escaped.

doc/source/faq.rst
```diff
@@ -14,7 +14,12 @@ goes beyond what classic email servers offer:
 -  **Instant/Realtime:** sub-second message delivery, realtime P2P
    streaming, privacy-preserving Push Notifications for Apple, Google, and `Ubuntu Touch <https://docs.ubports.com/en/latest/appdev/guides/pushnotifications.html>`_;
 
--  **Security Enforcement**: only strict TLS, DKIM and OpenPGP with minimized metadata accepted
+-  **Security Enforcement**: only strict TLS, DKIM and
+   OpenPGP with minimized metadata accepted.
+   Relays using ``tls_cert = self``
```
link2xt: This does not explain the difference. It says normally relays use TLS, DKIM and OpenPGP, and with `tls_cert = self` they also use TLS and OpenPGP.

cmdeploy/src/cmdeploy/selfsigned/deployer.py
```diff
@@ -0,0 +1,34 @@
...
+        apt.packages(
+            name="Install openssl",
+            packages=["openssl"],
+        )
+
+    def configure(self):
+        server.shell(
+            name="Generate self-signed TLS certificate if not present",
+            commands=[
+                f"[ -f {self.cert_path} ] || openssl req -x509 -newkey rsa:4096"
```
link2xt: We have switched to ECC (P-256) in https://github.com/chatmail/relay/pull/689
They are almost instant to generate and much smaller.
This should be `-newkey ec -pkeyopt ec_paramgen_curve:P-256`, otherwise self-signed certs will be RSA 4096.

cmdeploy/src/cmdeploy/selfsigned/deployer.py
```diff
@@ -0,0 +1,34 @@
...
+                f" -nodes -days 3650"
```
link2xt: `-nodes` is renamed into `-noenc`

link2xt: `-days 3650` is 10 years, i'd add one more 0 to make it expire in 100 years instead

cmdeploy/src/cmdeploy/selfsigned/deployer.py
```diff
@@ -0,0 +1,34 @@
...
+                f" -keyout {self.key_path}"
+                f" -out {self.cert_path}"
+                f' -subj "/CN={self.mail_domain}"'
+                f' -addext "subjectAltName=DNS:{self.mail_domain},DNS:www.{self.mail_domain},DNS:mta-sts.{self.mail_domain}"',
```
link2xt, line 31: maybe also add `-addext "extendedKeyUsage=serverAuth,clientAuth"`, normally certificates have EKU

cmdeploy/src/cmdeploy/selfsigned/deployer.py
```diff
@@ -0,0 +1,34 @@
...
+            ],
```
link2xt, line 32: Combined, command should be something like this:
```
openssl req -x509 \
        -newkey ec -pkeyopt ec_paramgen_curve:P-256 \
        -noenc -days 36500 -keyout privkey -out cert \
        -subj "/CN=example.org" \
        -addext "extendedKeyUsage=serverAuth,clientAuth" \
        -addext "subjectAltName=DNS:example.org,DNS:www.example.org,DNS:mta-sts.example.org"
```

cmdeploy/src/cmdeploy/tests/online/test_0_qr.py
```diff
@@ -11,15 +11,28 @@ def test_gen_qr_png_data(maildomain):
...
     """Test configuring accounts by scanning a QR code works."""
     url = f"DCACCOUNT:https://{maildomain}/new"
     for i in range(3):
         account_id = rpc.add_account()
-        rpc.set_config_from_qr(account_id, url)
+        if chatmail_config.tls_cert == "self":
+            # deltachat core's rustls rejects self-signed HTTPS certs during
+            # set_config_from_qr, so fetch credentials via requests instead
+            verify = False
+            res = requests.post(f"https://{maildomain}/new", verify=verify)
```
link2xt: Seems more clear to say `verify=False` here instead of defining `verify=False` and then using it as `verify=verify`.

cmdeploy/src/cmdeploy/tests/online/test_2_deltachat.py
```diff
@@ -180,6 +181,11 @@ def test_hide_senders_ip_address(cmfactory):
...
     user2._evtracker.wait_next_incoming_message()
-    user2.direct_imap.select_folder("Inbox")
-    msg = user2.direct_imap.get_all_messages()[0]
-    assert public_ip not in msg.obj.as_string()
+    addr = user2.get_config("addr")
+    host = addr.split("@")[1]
+    pw = user2.get_config("mail_pw")
+    mailbox = imap_tools.MailBox(host, ssl_context=ssl_context)
+    mailbox.login(addr, pw)
+    msgs = list(mailbox.fetch())
```
link2xt: Better use `fetch(seen=False)`, by default fetch marks messages as seen and it is unnecessary. Does not matter for this test, but it will be copy-pasted to other tests and may result in difficult to debug problems.

cmdeploy/src/cmdeploy/tests/plugin.py
```diff
@@ -297,19 +316,27 @@ def cache_maybe_store_configured_db_files(self, acc):
...
         def read_path(self, path):
             return
 
     am = ACFactory(request=request, tmpdir=tmpdir, testprocess=testproc, data=Data())
 
+    if chatmail_config.tls_cert == "self":
+        # Skip upstream's init_imap which creates a DirectImap with
+        # strict SSL that fails on self-signed certs.  Chatmail tests
+        # use fresh throwaway accounts, so folder cleanup is unnecessary.
+        am._acsetup.init_imap = lambda acc: None
```
link2xt: Can this be done unconditionally? If there are no more tests using `direct_imap`, can as well disable it completely.

cmdeploy/src/cmdeploy/selfsigned/deployer.py
```diff
@@ -0,0 +1,34 @@
...
+from cmdeploy.basedeploy import Deployer
+
+
+class SelfSignedTlsDeployer(Deployer):
+    """Generates a self-signed TLS certificate for all chatmail endpoints."""
+
+    def __init__(self, mail_domain):
+        self.mail_domain = mail_domain
+        self.cert_path = "/etc/ssl/certs/mailserver.pem"
+        self.key_path = "/etc/ssl/private/mailserver.key"
```
missytake, line 12: Instead of writing a whole module which generates TLS certificates, why not just use the certs dovecot ships? https://github.com/chatmail/relay/commit/fae556887345d9bd1206ce00f2e8b61c7a996d9a

(edit: ah, it seems to expire after a year and not be re-generated.)

.github/workflows/test-and-deploy-ipv4only.yaml
```diff
@@ -71,12 +71,21 @@ jobs:
...
+        run: |
           cmdeploy init staging-ipv4.testrun.org
           sed -i 's#disable_ipv6 = False#disable_ipv6 = True#' chatmail.ini
           sed -i 's/#\s*mtail_address/mtail_address/' chatmail.ini
+          sed -i 's/tls_cert = acme/tls_cert = self/' chatmail.ini
+          cmdeploy run --verbose --skip-dns-check
+          cmdeploy test --slow
 
-      - run: cmdeploy run --verbose --skip-dns-check
+      - name: deploy and test tls_cert=acme
```
missytake:
```suggestion
      - name: deploy tls_cert=acme
```

the test only happens below.

.github/workflows/test-and-deploy-ipv4only.yaml
```diff
@@ -71,12 +71,21 @@ jobs:
...
+        run: |
+          cmdeploy init staging-ipv4.testrun.org
```
missytake, line 75:
```suggestion
          rm chatmail.ini
          cmdeploy init staging-ipv4.testrun.org
```

`cmdeploy init` doesn't overwrite existing chatmail.ini files, so we would only test against `tls_cert = self`.

chatmaild/src/chatmaild/newemail.py
```diff
@@ -23,13 +24,26 @@ def create_newemail_dict(config: Config):
     return dict(email=f"{user}@{config.mail_domain}", password=f"{password}")
 
 
+def create_dclogin_url(email, password):
+    """Build a dclogin: URL with credentials and self-signed cert acceptance.
+
+    Uses ic=3 (AcceptInvalidCertificates) so Delta Chat clients
+    can connect to servers with self-signed TLS certificates.
+    """
+    return f"dclogin:{quote(email, safe='')}?p={quote(password, safe='')}&v=1&ic=3"
```
missytake:
```suggestion
    Uses ic=3 and sc=3 (AcceptInvalidCertificates) so Delta Chat clients
    can connect to servers with self-signed TLS certificates.
    """
    return f"dclogin:{quote(email, safe='')}?p={quote(password, safe='')}&v=1&ic=3&sc=3"
```

The SMTP certificates need to be accepted as well.

chatmaild/src/chatmaild/tests/test_newmail.py
```diff
@@ -25,3 +41,19 @@ def test_print_new_account(capsys, monkeypatch, maildomain, tmpdir, example_conf
...
+    config = make_config("chat.example.org", {"tls_cert": "self"})
+    monkeypatch.setattr(chatmaild.newemail, "CONFIG_PATH", str(config._inipath))
+    print_new_account()
+    out, err = capsys.readouterr()
+    lines = out.split("\n")
+    dic = json.loads(lines[2])
+    assert "dclogin_url" in dic
+    url = dic["dclogin_url"]
+    assert url.startswith("dclogin:")
+    assert "ic=3" in url
```
missytake, line 58:
```suggestion
    assert "ic=3" in url
    assert "sc=3" in url
```

And let's also test for it :)

chatmaild/src/chatmaild/tests/test_newmail.py
```diff
@@ -15,6 +19,18 @@ def test_create_newemail_dict(example_config):
     assert ac1["password"] != ac2["password"]
 
 
+def test_create_dclogin_url():
+    url = create_dclogin_url("user@example.org", "p@ss w+rd")
+    assert url.startswith("dclogin:")
+    assert "v=1" in url
+    assert "ic=3" in url
```
missytake, line 26:
```suggestion
    assert "ic=3" in url
    assert "sc=3" in url
```

cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
```diff
@@ -42,6 +42,9 @@ stream {
 }
 
 http {
+{% if config.tls_cert == "self" %}
+	limit_req_zone $binary_remote_addr zone=newaccount:10m rate=2r/s;
+{% endif %}
```
missytake: why only for instances with self-signed certs?

hpk42: because i didn't want to modify main line behaviour on the side, it's probably its own discussion.  might be we indeed want to generally think about rate limiting for new account creation.

chatmaild/src/chatmaild/newemail.py
link2xt: There is no separate option for SMTP, `sc` is ignored so this only makes the QR code larger.

cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
link2xt: This does not however limit account creation, only prevents spawning CGI processes.

link2xt: I also made a PR to switch to the new dcaccount scheme that creates accounts directly: https://github.com/chatmail/relay/pull/858
This is also how it works already for the default relay in the apps, they don't rely on the HTTP endpoint.

cmdeploy/src/cmdeploy/tests/online/test_2_deltachat.py
hpk42: imap_tools do not support seen flag here.  Do you maybe rather refer to core semantics?

link2xt:
> Note that relays configured with self-signed certificates will only interoperate with other relays that also use opportunistic encryption (tls_cert = self); they will not be able to federate with standard chatmail relays using Let's Encrypt (tls_cert = acme) because the default configuration enforces strict certificate validation. Enabling interoperability between self-signed and signed relays would require relaxing default security settings, which is out of scope of this PR, and requires more thorough discussion.

It's fine as is if treated as experimental option for getting roughly the same setup as real chatmail in testing VMs without the need to have a real domain.

Documentation commit however tries to motivate this with https://datatracker.ietf.org/doc/html/rfc7435 (informational RFC from December 2014, when only 60% of incoming and <80% of mails were encrypted for Gmail according to https://transparencyreport.google.com/safer-email/overview) which back then argued for accepting unauthenticated encryption even if it uses insecure algorithms and self-signed certificates over rejecting it and risking dropping to cleartext.

If we successfully promote this option like FAQ currently does, as an easy way to setup without security implications and just as secure because of OpenPGP, this will result in a separate network of relays which can federate with anything in contrast to default policy TLS relays which cannot federate with mail servers using self-signed certificates. As the option to enable federation is tied to using self-signed certificates, this will force other relays that want to federate with relays using self-signed certificate to also switch to self-signed certificates, or we will have to decouple `smtp_tls_security_level` from this setting and just set it to "encrypt". Promoting the option as easy setup in the FAQ only makes sense if the goal is to be eventually forced to disable TLS checks between all relays.

Recent changes to support federation for IP-based relays in https://github.com/chatmail/relay/pull/846 and https://github.com/chatmail/relay/pull/831 (filtermail 0.4.0) do not have this problem because IP-based relays exist in a separate namespace. Only relays with IP-addresses use self-signed certificates, so we know in advance from the recipient address whether to expect a self-signed or WebPKI-signed certificate.

One way to put domain-based relays with self-signed certificates in its own namespace is to use domains that cannot be used to obtain WebPKI certificates. One such option is domains that start with `_`. I have tested and they open in Firefox just like normal website and can have A/AAAA records, but it is not possible to get valid TLS certificates (except for maybe wildcard on the higher level, but then its really on you if you get wildcard certificate and create a subdomain with `_`, cannot happen accidentally) for them because [CA/Browser Forum](https://github.com/cabforum/servercert/blob/24f38fd4765e019db8bb1a8c56bf63c7115ce0b0/docs/BR.md) requirements explicitly require that 

> The Fully-Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of P-Labels or Non-Reserved LDH Labels joined together by a U+002E FULL STOP (".") character.

"P-Labels" are punycode (and start with xn--), LDH means starting with letter, digit or hyphen. Domains starting with `_` are explicitly excluded.
If we do this and say the option is for experiments, we don't even need an explicit `chatmail.ini` setting but can derive it from the domain name by checking for _ right after loading the config. And we can do the same in chatmail core easily as it should not expect a valid certificate for domain starting with `_` anyway.
And then instead of switching global `smtp_tls_security_level`, we can add line `/^_/ encrypt` to https://github.com/chatmail/relay/blob/36eb63faa1bb5dac8b02a7902dd71b6f3003ad11/cmdeploy/src/cmdeploy/postfix/smtp_tls_policy_map for all relays unconditionally and have federation with such relays already.

cmdeploy/src/cmdeploy/tests/online/test_2_deltachat.py
link2xt: Sorry, it's `mark_seen`, not `seen`. The flag is in the README even: https://pypi.org/project/imap-tools/
See https://github.com/chatmail/core/pull/7424#discussion_r2515281480, this resulted in the problems in the past in core tests.
The reason is the IMAP standard, if you fetch body of the message, message is marked as seen. To avoid this you need to fetch `BODY.PEEK[]` and this is what `mark_seen=False` does.

More direct link: https://github.com/ikvk/imap_tools/issues/179

cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
hpk42: indeed, i added the rate limit bvecause of spawning of CGI processes. it takes ~0.3 seconds on c2 for the web page to render

cmdeploy/src/cmdeploy/selfsigned/deployer.py
hpk42: also it's not just dovecot but also postfix, nginx etc.  The default place for something like this is /etc/ssl as far as i gather.  

I prefer this tls deployer approach to special-casing ACME deployer.   Everything around TLS certificates is already muddied so i thought a bit explicit naming and handling can not hurt.

cmdeploy/src/cmdeploy/tests/online/test_0_qr.py
hpk42: changed.

www/src/dclogin.js
hpk42: tried to follow advise.

doc/source/faq.rst
hpk42: i removed the mentioning completely.  it's not relegated more as a note on testing or special circumstances elsewhere.

cmdeploy/src/cmdeploy/selfsigned/deployer.py
hpk42: done

cmdeploy/src/cmdeploy/tests/online/test_0_qr.py
hpk42: did so.

cmdeploy/src/cmdeploy/tests/online/test_2_deltachat.py
hpk42: did so.

hpk42:
thanks, i went with your Underscore domains suggestion, removed the chatmail.ini option, and demoted the mentioning of self-signed relays generally.  There is no reference to RFC7435 left but IMHO that RFC has a certain timelessness in some of its views which i share.  But this PR here is not the place to discuss this further.   PR is now scoped towards testing and special circumstances (mesh networks, local setups) and not in encouraging potentially questionable security practises.

cmdeploy/src/cmdeploy/tests/plugin.py
hpk42: did it.

cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
hpk42: as to #858 that doesn't worked with self-signed, as far as i gather.

hpk42: lastly, how to limit account creation is a separate topic because it's indeed not about rate-limiting the http request here. I suggest to keep the rate-limiting of cgi-bin invocations here for now.

.github/workflows/test-and-deploy-ipv4only.yaml
```diff
@@ -75,8 +75,7 @@ jobs:
           cmdeploy init staging-ipv4.testrun.org
           sed -i 's#disable_ipv6 = False#disable_ipv6 = True#' chatmail.ini
           sed -i 's/#\s*mtail_address/mtail_address/' chatmail.ini
-
-      - run: cmdeploy run --verbose --skip-dns-check
+          cmdeploy run --verbose --skip-dns-check
```
missytake, line 78: So now we don't test self-signed certs anymore? I'm fine with this for now - kind of planning to move staging-ipv4.testrun.org to test IP-only relays, so then we would move to self-signed certs ;) But let's discuss then.

Is there a reason we don't have a separate CI step for `cmdeploy run` anymore? It made the output clearer to distinguish imho.

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -66,6 +66,9 @@
 # if set to "True" IPv6 is disabled
 disable_ipv6 = False
 
+
+
+
```
missytake:
```suggestion
```

These empty lines don't really serve a purpose.

j4n:
looks good to me - I was working on the third option, externally generated certs (for traefik, #662) in https://github.com/chatmail/relay/pull/857 -- probably makes sense to modify this to adapt the CHATMAIL_NOACME part on the work here once this is merged.

.github/workflows/test-and-deploy-ipv4only.yaml
hpk42: yes, i dropped it because the whole staging thing is already muddied enough, with the zone file shenenigans. It's not even clear that testing self-signed here with a zone file that might have valid TLS cert makes sense etc.   As it's now only for testing purposes it's maybe not strictly needed to test it in CR.

cmdeploy/src/cmdeploy/tests/online/test_0_qr.py
```diff
@@ -8,18 +9,34 @@ def test_gen_qr_png_data(maildomain):
...
     url = f"DCACCOUNT:https://{maildomain}/new"
     for i in range(3):
         account_id = rpc.add_account()
-        rpc.set_config_from_qr(account_id, url)
+        if chatmail_config.tls_cert_mode == "self":
+            # deltachat core's rustls rejects self-signed HTTPS certs during
+            # set_config_from_qr, so fetch credentials via requests instead
+            res = requests.post(f"https://{maildomain}/new", verify=False)
+            data = res.json()
+            rpc.add_or_update_transport(account_id, {
```
link2xt, line 33: You don't need to run `configure` after running `add_or_update_transport`. By running `configure` you configure the account twice, it should be only in the `else` branch.

chatmaild/src/chatmaild/ini/chatmail.ini.f
hpk42: yes, i am doing a last formatting fix push, there were also other whitespace changes and one omission/commit accident.

cmdeploy/src/cmdeploy/tests/online/test_0_qr.py
```diff
@@ -8,18 +9,34 @@ def test_gen_qr_png_data(maildomain):
...
+                "addr": data["email"],
+                "password": data["password"],
+                "imapServer": maildomain,
+                "smtpServer": maildomain,
+                "certificateChecks": "acceptInvalidCertificates",
+            })
+        else:
+            rpc.set_config_from_qr(account_id, url)
```
link2xt: It is unrelated refactoring, but if you do `rpc.add_transport_from_qr(account_id, url)`, you don't need `configure` in this branch either.

.github/workflows/test-and-deploy-ipv4only.yaml
```diff
@@ -88,7 +87,7 @@ jobs:
           ssh root@ns.testrun.org nsd-checkzone staging-ipv4.testrun.org /etc/nsd/staging-ipv4.testrun.org.zone
           ssh root@ns.testrun.org systemctl reload nsd
 
-      - name: cmdeploy test
+      - name: cmdeploy test acme
```
link2xt: At this point `chatmail.ini` is already generated, do we need to specify `acme` manually? Looks like it can be derived from the hostname.

link2xt: Does not look like it is used in the rest of the code, probably can be just removed.

PR #858, opened 2026-02-18
feat: create accounts without HTTP request

link2xt:
This is supported since chatmail core version 2.23.0.

### DIFF
--- a/cmdeploy/src/cmdeploy/genqr.py
+++ b/cmdeploy/src/cmdeploy/genqr.py
@@ -7,7 +7,7 @@ from PIL import Image, ImageDraw, ImageFont
 
 
 def gen_qr_png_data(maildomain):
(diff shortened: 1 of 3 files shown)


link2xt:
We can decide to merge this later, but this will anyway take some time to propagate to relays and it is not critical that users who don't upgrade will be able to scan the QR code.

hpk42 (approved):
lgtm.  worst case is that someone using a pre-feb-2026 core version fails to use a relay that offers the new scheme on the relay web page.

link2xt:
Will also not work for relays which tweaked username or password length requirements.

missytake:
> Will also not work for relays which tweaked username or password length requirements.

Then let's get rid of those values for creating new addresses, while keeping old addresses working? addresses are hidden nowadays anyway, the length is not really a factor. And password length should be a safe default anyway. Less config options that way :)

PR #860, opened 2026-02-19
feat: support externally managed TLS via tls_external_cert_and_key option

hpk42:
Add tls_external_cert_and_key config option for chatmail servers that manage their own TLS certificates outside of ACME. When set, the deployer verifies the cert and key files exist on the server and installs a systemd path unit that watches the certificate via inotify — when it changes, dovecot and nginx are automatically reloaded (postfix reads certs per handshake so needs no reload).  Includes unit tests, an e2e test script, docs, and a CI workflow.

this is based on the prior work of self-signed TLS certs in #855 

PR replaces and closes #662

### DIFF
--- a/cmdeploy/src/cmdeploy/external/deployer.py
+++ b/cmdeploy/src/cmdeploy/external/deployer.py
@@ -63,11 +63,5 @@ class ExternalTlsDeployer(Deployer):
             restarted=self.need_restart,
             daemon_reload=self.need_restart,
         )
-        # Trigger the oneshot service so services pick up the current cert.
-        # The path unit handles future changes via inotify.
-        systemd.service(
-            name="Reload TLS services for current certificate",
-            service="tls-cert-reload.service",
-            running=True,
-            daemon_reload=False,
-        )
+        # No explicit reload needed here: dovecot/nginx read the cert
+        # on startup, and the .path watcher handles live changes.
--- a/cmdeploy/src/cmdeploy/external/tls-cert-reload.path.f
+++ b/cmdeploy/src/cmdeploy/external/tls-cert-reload.path.f
@@ -1,6 +1,10 @@
 # Watch the TLS certificate file for changes.
 # When the cert is updated (e.g. renewed by an external process),
-# this triggers tls-cert-reload.service to restart the affected services.
+# this triggers tls-cert-reload.service to reload the affected services.
+#
+# NOTE: changes to the certificates are not detected if they cross bind-mount boundaries. 
+# After cert renewal, you must then trigger the reload explicitly:
+#   systemctl start tls-cert-reload.service
 [Unit]
 Description=Watch TLS certificate for changes
 
--- a/cmdeploy/src/cmdeploy/external/tls-cert-reload.service
+++ b/cmdeploy/src/cmdeploy/external/tls-cert-reload.service
@@ -11,5 +11,5 @@ Description=Reload TLS services after certificate change
 
 [Service]
 Type=oneshot
-ExecStart=/bin/systemctl reload dovecot
-ExecStart=/bin/systemctl reload nginx
+ExecStart=/bin/systemctl try-reload-or-restart dovecot
+ExecStart=/bin/systemctl try-reload-or-restart nginx
--- a/doc/source/getting_started.rst
+++ b/doc/source/getting_started.rst
@@ -229,7 +229,11 @@ The deploy will verify that both files exist on the server.
    You are responsible for certificate renewal.
    When the certificate file changes on disk,
    all relay services pick up the new certificate automatically
-   (via a systemd path watcher installed during deploy).
+   via a systemd path watcher installed during deploy.
+   The watcher uses inotify, which does not cross bind-mount boundaries.
+   If you use such a setup, you must trigger the reload explicitly after renewal::
+
+      systemctl start tls-cert-reload.service
 
 
 Migrating to a new build machine


chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -48,6 +48,11 @@
 # (space-separated, item may start with "@" to whitelist whole recipient domains)
 passthrough_recipients =
 
+# To use externally managed TLS certificates (e.g. from Traefik or certbot),
```
link2xt: Not clear from the description (have not read the code): is the path local to the deployer or is it on the server?

hpk42: it's all on the server. the docs are clearer on this, but not yet this comment. I'll collect feedback and address it in batch.

hpk42: i fixed the comment.

j4n:
Looks good, will try it with traefik soon.

chatmaild/src/chatmaild/tests/test_config.py
```diff
@@ -87,3 +87,36 @@ def test_config_tls_self(make_config):
...
+
+def test_config_tls_external_overrides_underscore(make_config):
+    config = make_config(
+        "_test.example.org",
+        {
+            "tls_external_cert_and_key": "/certs/fullchain.pem /certs/privkey.pem",
+        },
+    )
+    assert config.tls_cert_mode == "external"
+    assert config.tls_cert_path == "/certs/fullchain.pem"
```
link2xt, line 112: Could check `tls_key_path` as well.

chatmaild/src/chatmaild/config.py
```diff
@@ -60,10 +60,24 @@ def __init__(self, inipath, params):
...
+
+        if external:
+            parts = external.split()
+            if len(parts) != 2:
+                raise ValueError(
+                    "tls_external_cert_and_key must have two space-separated"
+                    " paths: CERT_PATH KEY_PATH"
+                )
+            self.tls_cert_mode = "external"
+            self.tls_cert_path = parts[0]
```
link2xt: Can use multi-assigment to double-check that we only have two parts:
```
self.tls_cert_path, self.tls_key_path = parts
```

cmdeploy/src/cmdeploy/external/deployer.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, cert_path, key_path):
+        self.cert_path = cert_path
+        self.key_path = key_path
+
+    def configure(self):
+        server.shell(
+            name="Verify external TLS certificate and key exist",
+            commands=[
+                f"test -f {self.cert_path} && test -f {self.key_path}",
```
link2xt: Maybe better use facts https://docs.pyinfra.com/en/3.x/facts/files.html#files-file and return real exception here explaining what went wrong.

cmdeploy/src/cmdeploy/external/deployer.py
```diff
@@ -0,0 +1,69 @@
...
+            ],
+        )
+
+        # Deploy the .path unit (templated with the cert path).
+        source = get_resource("tls-cert-reload.path.f", pkg=__package__)
```
link2xt, line 33: We don't specify `pkg=__package__` explicitly in any other place and it is the default, is it needed?

cmdeploy/src/cmdeploy/external/deployer.py
```diff
@@ -0,0 +1,69 @@
...
+        content = source.read_text().format(cert_path=self.cert_path).encode()
+
+        import io
```
link2xt: Could be moved to the top of the file?

cmdeploy/src/cmdeploy/external/deployer.py
```diff
@@ -0,0 +1,69 @@
...
+            src=io.BytesIO(content),
+            dest="/etc/systemd/system/tls-cert-reload.path",
+            user="root",
+            group="root",
+            mode="644",
+        )
+
+        service_unit = files.put(
+            name="Upload tls-cert-reload.service",
+            src=get_resource("tls-cert-reload.service", pkg=__package__),
```
link2xt, line 47: Same, no need to specify `__package__` explicitly?

cmdeploy/src/cmdeploy/external/deployer.py
```diff
@@ -0,0 +1,69 @@
...
+            service="tls-cert-reload.path",
+            running=True,
+            enabled=True,
+            restarted=self.need_restart,
+            daemon_reload=self.need_restart,
+        )
+        # Always trigger a reload so services pick up the current cert.
+        # The path unit handles future changes via inotify.
+        server.shell(
+            name="Reload TLS services for current certificate",
```
link2xt: Can this be done via `systemd.service` with `running=True`? I think it will try starting oneshot service every time just the same.

cmdeploy/src/cmdeploy/deployers.py
```diff
@@ -536,6 +537,20 @@ def activate(self):
         )
 
 
+def get_tls_deployer(config, mail_domain):
```
link2xt, line 544: High-level problem with this is that if you have a server with acmetool and want to reconfigure with external certificate, acmetool does not get uninstalled.

I think what we need is not selecting one deployer, but run all deployers with some option like `enabled=...` that tells the deployer if it should deploy or un-deploy acme or tls-cert-reload services.

hpk42: this is addressed in the separate https://github.com/chatmail/relay/pull/869 (the first commit, the second does some follow up refactoring) 

all other review comments were fixed in the branch here.

.github/workflows/test-and-deploy-ipv4only.yaml
```diff
@@ -103,11 +103,3 @@ jobs:
...
         run: ssh root@staging-ipv4.testrun.org "cd relay && scripts/cmdeploy dns -v --ssh-host localhost"
 
-  test-tls-external:
-    needs: deploy
-    uses: ./.github/workflows/reusable-test-tls-external.yaml
-    with:
-      domain: staging-ipv4.testrun.org
-    secrets:
-      STAGING_SSH_KEY: ${{ secrets.STAGING_SSH_KEY }}
-
```
hpk42: didn't want to further debug the workflow. Thorough TLS-mode testing is better done whne we have a more flexible provisioning of VPS and DNS for PRs.

cmdeploy/src/cmdeploy/tests/plugin.py
```diff
@@ -381,10 +381,12 @@ def iter_output(self, logcmd=""):
...
             res = line.decode().strip().lower()
-            if res:
-                yield res
-            else:
+            if not res:
                 break
+            if ready is not None:
+                ready()
+                ready = None
+            yield res
```
hpk42, line 418: this is a test fix on the side for a flaky test, took me a while to figure out.  There are concurrency issues with getting log lines and triggering the send message.  I think this is now reliable.

cmdeploy/src/cmdeploy/tests/setup_tls_external.py
```diff
@@ -0,0 +1,340 @@
+"""Setup and verify external TLS certificates for a chatmail server.
+
+Generates a self-signed TLS certificate, uploads it to the chatmail
+server via SCP, runs ``cmdeploy run``, and then probes all TLS-enabled
+ports (nginx, postfix, dovecot) to verify the certificate is actually
+served.  After probing, checks remote service logs for errors.
+
```
hpk42: this test is only manually run, and partially machine generated.

cmdeploy/src/cmdeploy/external/deployer.py
```diff
@@ -0,0 +1,73 @@
...
+            name="Enable tls-cert-reload path watcher",
+            service="tls-cert-reload.path",
+            running=True,
+            enabled=True,
+            restarted=self.need_restart,
+            daemon_reload=self.need_restart,
+        )
+        # Trigger the oneshot service so services pick up the current cert.
+        # The path unit handles future changes via inotify.
+        systemd.service(
```
j4n: On a fresh deploy (or Docker container start), those services aren't running yet, so the reload fails. We can probably just remove this as dovecot/nginx read the cert on startup, and the .path watcher handles live cert changes via inotify.

j4n:
One caveat turned up: inotify doesn't cross bind-mount boundaries and if certificates are modified outside of the container, the reload must be triggered explicitly.

hpk42:
@j4n i addressed your three comments in https://github.com/chatmail/relay/commit/3f05a9a8b4b26637e44e35f73cc1c5bf628b49c7

PR #861, opened 2026-02-19
feat: reconfigure acmetool from redirector to proxy mode

### LINKED ISSUE #809: Change AcmeTool to not listen on port 80
Sometimes relay operators report certs aren't getting renewed and it tends to be an issue with acmetool wanting to bind to port 80 but Nginx already has port 80. This can be solved by just reverse proxying to acmetool instead.

```
    location /.well-known/acme-challenge/ {
      proxy_pass http://127.0.0.1:402;
    }
```

then in the acme config you tell it to listen on port 402 instead

https://hlandau.github.io/acmetool/userguide#web-server-configuration-challenges

link2xt:
This elimitates the problem of acmetool failing
to start when nginx is installed already and uses port 80.

This also makes nginx redirect HTTP requests to HTTPS for setups that don't have acmetool.

Closes #809

### DIFF
--- a/cmdeploy/src/cmdeploy/acmetool/acmetool-redirector.service
+++ b/cmdeploy/src/cmdeploy/acmetool/acmetool-redirector.service
@@ -3,7 +3,7 @@ Description=acmetool HTTP redirector
 
 [Service]
 Type=notify
-ExecStart=/usr/bin/acmetool redirector --service.uid=daemon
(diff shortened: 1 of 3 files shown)


ccclxxiii:
> This elimitates the problem of acmetool failing to start when nginx is installed already and uses port 80.
> 
> This also makes nginx redirect HTTP requests to HTTPS for setups that don't have acmetool.
> 
> Closes #809

just deployed on a relay.  this is helpful.  thank you

ccclxxiii (approved):

PR #862, opened 2026-02-19
fix(cmdeploy): Rewrite dovecot install logic, update

j4n:
The old code did not install updates when the service was running; check
installed version instead of systemd status. Also, rewrite install logic
to extract dovecot version and hashes as module-level constants.

### DIFF
--- a/cmdeploy/src/cmdeploy/basedeploy.py
+++ b/cmdeploy/src/cmdeploy/basedeploy.py
@@ -1,6 +1,7 @@
 import importlib.resources
 import io
 import os
+from contextlib import contextmanager
 
 from pyinfra.operations import files, server, systemd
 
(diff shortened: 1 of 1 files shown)


missytake:
There was some reason why we didn't want to download and install this on every run, @link2xt do you remember?

j4n:
> There was some reason why we didn't want to download and install this on every run, @link2xt do you remember?

While I still do have a hunch I may be barking up the wrong tree here, regarding this specifically, it wouldn't, as long as the chatmail repo and the hashes/versions are not touched it won't do it.

missytake:
Ah right, we don't want to upgrade dovecot if it is not running:

https://github.com/chatmail/relay/pull/841#issuecomment-3968154506

j4n:
> Ah right, we don't want to upgrade dovecot if it is not running:
> 
> [#841 (comment)](https://github.com/chatmail/relay/pull/841#issuecomment-3968154506)

brough in the fix from  https://github.com/chatmail/relay/pull/884/changes#r2935020218 and rebased.

PR #869, opened 2026-02-23
automate and improve change-tracking in deployers

hpk42:
Automate change tracking for added/remove files and systemd services.  This removes a lot of boilerplate from the deployers and makes them more readable and less error-prone.


cmdeploy/src/cmdeploy/basedeploy.py
```diff
@@ -150,3 +143,113 @@ def configure(self):
...
+        self.daemon_reload = False
+
+    def ensure_systemd_unit(self, src, **kwargs):
+        dest_name = src.split("/")[-1].replace(".j2", "")
+        dest = f"/etc/systemd/system/{dest_name}"
+        if src.endswith(".j2"):
+            return self.put_template(src, dest, **kwargs)
+        return self.put_file(src, dest)
+
+    def put_file(self, src, dest, mode="644"):
```
j4n, line 169:
```suggestion
    def put_file(self, src, dest, mode="644", reason=None):
```
I think it would be valuable to  preserve the ability to explain changes for put_file, put_template, remove_file, and embed it in braces into the name parameter.

j4n, line 169: also, this takes mode as an argument, but template doesnt, but takes owner - maybe this should be symmetrical to be less confusing.

cmdeploy/src/cmdeploy/basedeploy.py
```diff
@@ -150,3 +143,113 @@ def configure(self):
...
+            daemon_reload=self.daemon_reload,
+        )
+        self.daemon_reload = False
+
+    def ensure_systemd_unit(self, src, **kwargs):
+        dest_name = src.split("/")[-1].replace(".j2", "")
+        dest = f"/etc/systemd/system/{dest_name}"
+        if src.endswith(".j2"):
+            return self.put_template(src, dest, **kwargs)
+        return self.put_file(src, dest)
```
j4n, line 167: maybe we should pass kwargs to the non-template-case, to honor `dest_name` for example.

j4n (approved):
Looks good to me, a few minor suggestions.

Tested deploying, re-deploying, and redeploy with modified system units, all good.

It makes me wish to have a `pyinfra_helpers.py` we could use in other repos as well, with just the wrapper functions :)

cmdeploy/src/cmdeploy/basedeploy.py
hpk42: The new API is meant to only consists of things that map the previous usage, nothing more. If some other argument is needed, it is easy enough to add when it is actually needed.

hpk42:
a more general API or `pyinfra_helpers.py` is a separate concern (which i am not sure about how useful it is).

PR #870, opened 2026-02-23
fix: DNS check timeout with IPv6-broken authoritative NS

### LINKED ISSUE #851: DNS check during deployment is not working with desec.io as upstream DNS(SEC) provider
- Server OS (Operating System): `Debian 12.12`
- On which OS you run cmdeploy: `Manjaro Linux`
- chatmail/relay version: `87022e3681b4de62eac81ad1f5b5552ac24d3bec`

## Expected behavior

The DNS checks before deployment don't time out when the SOA is not answering (or there should be a fallback).

## Actual behavior

I get a timeout after 60s and the deployment script stops.

### Steps to reproduce the problem:

1. `cmdeploy run`
2. Wait for the following error message:

```
[ssh] login to mydomain.de
[ssh] Collecting initial DNS settings.
.........Traceback (most recent call last):
  File "/home/user/relay/venv/bin/cmdeploy", line 6, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/user/relay/cmdeploy/src/cmdeploy/cmdeploy.py", line 411, in main
    res = args.func(args, out, **kwargs)
  File "/home/user/relay/cmdeploy/src/cmdeploy/cmdeploy.py", line 95, in run_cmd
    remote_data = dns.get_initial_remote_data(sshexec, args.config.mail_domain)
  File "/home/user/relay/cmdeploy/src/cmdeploy/dns.py", line 10, in get_initial_remote_data
    return sshexec.logged(
           ~~~~~~~~~~~~~~^
        call=remote.rdns.perform_initial_checks, kwargs=dict(mail_domain=mail_domain)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/user/relay/cmdeploy/src/cmdeploy/sshexec.py", line 79, in logged
    return self(call, kwargs, log_callback=print_stderr)
  File "/home/user/relay/cmdeploy/src/cmdeploy/sshexec.py", line 65, in __call__
    code, data = self._remote_cmdloop_channel.receive(timeout=self.timeout)
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/relay/venv/lib/python3.14/site-packages/execnet/gateway_base.py", line 931, in receive
    raise self.TimeoutError("no item after %r seconds" % timeout) from None
execnet.gateway_base.TimeoutError: no item after 60 seconds
....
```

### What I already checked
- The DNS entries are in place
- No firewall is blocking the requests
- The "error" seems to be on desec.io server side (maybe this is expected behaviour by them)

### Logs
Executing the following two commands (which are executed by the deploy script on the remote host) I get the following results:

`dig -r -q mydomain.de -t SOA +noall +authority +answer`
```
mydomain.de.		42	IN	SOA	get.desec.io. get.desec.io. 2026024337 86400 3600 2419200 3600
```

`dig @get.desec.io. -r -q mydomain.de -t A +short`
```
;; communications error to 2a01:4f8:10a:1044:deec:642:ac10:80#53: timed out
;; communications error to 2a01:4f8:10a:1044:deec:642:ac10:80#53: timed out
;; communications error to 2a01:4f8:10a:1044:deec:642:ac10:80#53: timed out
1.2.3.4 <-- COMMENT: This is actually the right IP and I think it is resolved using IPv4 but I'm not sure
```

### Workaround
Deploying with `cmdeploy run --skip-dns-check` works, but I think the script should have a fallback to either use IPv4 then or provide an error message which tells people to use the `--skip-dns-check` parameter

hpk42:
Add IPv4 fallback for authoritative NS queries in query_dns(). When dig @ns returns no useful answer (e.g. IPv6 transport to the NS times out), retry with -4 to force IPv4. Also limits timeout and tries (+timeout=10 +tries=2) to avoid hanging.
all added code has unit-tests because it's hard to reliable reproduce in real-life. 

Fixes #851

### DIFF
--- a/cmdeploy/src/cmdeploy/remote/rdns.py
+++ b/cmdeploy/src/cmdeploy/remote/rdns.py
@@ -64,11 +64,13 @@ def get_dkim_entry(mail_domain, pre_command, dkim_selector):
     )
 
 
-def query_dns(typ, domain):
+def query_dns(typ, domain, shell_exec=None):
+    if shell_exec is None:
+        shell_exec = shell
     # Get autoritative nameserver from the SOA record.
     soa_answers = [
         x.split()
-        for x in shell(
+        for x in shell_exec(
             f"dig -r -q {domain} -t SOA +noall +authority +answer", print=log_progress
         ).split("\n")
     ]
@@ -78,8 +80,27 @@ def query_dns(typ, domain):
     ns = soa[0][4]
 
     # Query authoritative nameserver directly to bypass DNS cache.
-    res = shell(f"dig @{ns} -r -q {domain} -t {typ} +short", print=log_progress)
-    return next((line for line in res.split("\n") if not line.startswith(";")), "")
+    return _dig_authoritative(ns, domain, typ, shell_exec=shell_exec)
+
+
+def _parse_dig_result(output):
+    """Return first non-comment, non-empty line from dig output, or empty string."""
+    lines = [line for line in output.split("\n") if not line.startswith(";")]
+    return next((line for line in lines if line.strip()), "")
+
+
+def _dig_authoritative(ns, domain, typ, shell_exec=None):
+    """Query authoritative NS, falling back to IPv4-only if default fails."""
+    if shell_exec is None:
+        shell_exec = shell
+
+    # limit timeout and tries to not hang on a broken default NS
+    cmd = f"dig @{ns} -r -q {domain} -t {typ} +short +timeout=10 +tries=2"
+    result = _parse_dig_result(shell_exec(cmd, print=log_progress))
+    if result:
+        return result
+    # Fallback: force IPv4 transport (handles broken IPv6 to NS)
+    return _parse_dig_result(shell_exec(cmd + " -4", print=log_progress))
 
 
 def check_zonefile(zonefile, verbose=True):
--- a/cmdeploy/src/cmdeploy/tests/test_dns.py
+++ b/cmdeploy/src/cmdeploy/tests/test_dns.py
@@ -214,3 +214,59 @@ class TestZonefileChecks:
         assert not mockout.captured_red
         assert "correct" in mockout.captured_green[0]
         assert not mockout.captured_red
+
+
+class TestDigAuthoritative:
+    @pytest.fixture
+    def dig_auth(self):
+        """Helper that calls _dig_authoritative with a recording shell function."""
+        calls = []
+
+        def run(first_result, ipv4_result=None):
+            def shell(cmd, print=print):
+                calls.append(cmd)
+                if "-4" in cmd:
+                    return ipv4_result or ""
+                return first_result
+
+            result = remote.rdns._dig_authoritative(
+                "ns1.example.", "example.com", "A", shell_exec=shell
+            )
+            return result, calls
+
+        return run
+
+    def test_ipv4_fallback_on_error(self, dig_auth):
+        """Fallback to -4 when first query returns only error lines."""
+        result, calls = dig_auth(
+            first_result=(
+                ";; communications error to 2a01:4f8:10a:1044::80#53: timed out\n"
+                ";; communications error to 2a01:4f8:10a:1044::80#53: timed out\n"
+            ),
+            ipv4_result="1.2.3.4",
+        )
+        assert len(calls) == 2
+        assert "-4" not in calls[0]
+        assert "-4" in calls[1]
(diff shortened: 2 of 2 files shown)


hpk42:
nami is right i think https://github.com/chatmail/relay/issues/851#issuecomment-3944523597
so this PR is not adequate. closing it. Edit: then again, getting at the proper NS records is maybe not trivial either.

hpk42:
the PR approach here is probably worthwhile after all: we kind of always want to fallback if a DNS query times out and there is another (ipv4) candidate available.

missytake (approved):
I think to resolve #851 we need to query for SOA records only if NS records are not available though: https://github.com/chatmail/relay/pull/871#issuecomment-3946831698
That IPv4 worked in that case seems accidental.

Still looks good though.

hpk42:
> I think to resolve #851 we need to query for SOA records only if NS records are not available though: [#871 (comment)](https://github.com/chatmail/relay/pull/871#issuecomment-3946831698) That IPv4 worked in that case seems accidental.
> 
> Still looks good though.

let's see if it solves the reallife issue for @codiflow

codiflow:
@hpk42

I tried the fix from #870 but it seems to not resolve the issue with deSEC.

I changed `rdns.py` according to [this](https://github.com/chatmail/relay/pull/870/commits/7e8df45b9927936ed2c8c9a46016b69a4e32154a#diff-999746df9a943b360b7eadd59fe809a5cf0afbfb793bd4e459d0c775eaf31226) and executed `scripts/cmdeploy dns`

I still get `execnet.gateway_base.TimeoutError: no item after 60 seconds`.

Executing `dig @get.desec.io -r -q redacteddomain.de -t NS +short +timeout=10 +tries=2` manually returns the right values so from a DNS perspective it should work:

```
ns1.desec.io.
ns2.desec.org.
```

Then I tried the fix from #871 .

I changed `rdns.py` according to [this](https://github.com/chatmail/relay/pull/871/changes/fc48ac9042a98f78e8f53cb24a01a96fa2a77cb4#diff-999746df9a943b360b7eadd59fe809a5cf0afbfb793bd4e459d0c775eaf31226) and executed `scripts/cmdeploy dns`

After around 15s I get the following error message:

```
cmdeploy.sshexec.FuncError: Traceback (most recent call last):
  File "/root/from-cmdeploy/remote/_sshexec_bootstrap.py", line 18, in _handle_one_request
    res = func(**kwargs)
          ^^^^^^^^^^^^^^
  File "/root/from-cmdeploy/remote/rdns.py", line 25, in perform_initial_checks
    MTA_STS = query_dns("CNAME", f"mta-sts.{mail_domain}")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/from-cmdeploy/remote/rdns.py", line 77, in query_dns
    res = shell(f"dig @{ns} -r -q {domain} -t {typ} +short", print=log_progress)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/from-cmdeploy/remote/rshell.py", line 16, in shell
    return check_output(command, **args).decode().rstrip()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'dig @redacteddomain.de. -r -q mta-sts.redacteddomain.de -t CNAME +short' returned non-zero exit status 9.
```

Makes sense because it is using **my domain** as NS (see the @redacteddomain.de) instead of the SOA record.

Hope this helps 🙏

missytake:
@codiflow can you send me your domain name to missytake@systemli.org, so I can check for myself? I think without looking at the zone files directly it will be hard to figure out a path forward.

hpk42:
#954 is the proper way likely.

PR #915, opened 2026-04-09
fix(cmdeploy): Set permissions on dovecot pin prefs

mrdaemon:
Ensure the preferences.d snippet that pins dovecot packages to block Debian dist-upgrades is owned by root:root and has 644 permissions.

Files in this directory are generally expected to be world readable to ensure unprivileged operations such as apt-get in simulation mode. Having them not world readable breaks such usages.


missytake (approved):

missytake:
Thanks for finding this :)

PR #916, opened 2026-04-10
feat: Use filtermail for delivery to remote MTAs

j-g00da:


cmdeploy/src/cmdeploy/postfix/main.cf.j2
```diff
@@ -20,7 +20,11 @@ smtpd_tls_key_file={{ config.tls_key_path }}
 smtpd_tls_security_level=may
 
 smtp_tls_CApath=/etc/ssl/certs
-smtp_tls_security_level=verify
+# Messages are relayed to locally running filtermail-transport
+# which handles actual transport (see default_transport).
+# Opening TLS connection to local service is redundant.
+# The actual TLS connection with other MTAs is handled by filtermail-transport.
+smtp_tls_security_level=none
```
hpk42: better to also make lmtp tls security policy explicit and i suggest to shorten the comment. 
```suggestion
# Local default-transport uses LMTP to filtermail-transport which handles outbound TLS.
lmtp_tls_security_level = none
smtp_tls_security_level = none
```

hpk42 (approved):
lgtm except one nit on ltmp tls.

hpk42:
would be good to do an update of https://chatmail.at/doc/relay/overview.html#chatmail-relay-dependency-diagram as well, maybe as part of this PR here.

cmdeploy/src/cmdeploy/postfix/main.cf.j2
j-g00da: Ah this is not needed at all I think, lmtp doesn't use tls by default I'm pretty sure, this is a leftover from when we were to use smtp

hpk42: i think it's useful to explicitely talk about and set the tls_security_level, to make it explicit that the postfix smtp client is not used at all when this PR is merged.

hpk42 (approved):

j-g00da:
I will wait with this for the next release of filtermail, since it now caches smtp connections.

j-g00da:
Updated dependency diagram + now uses filtermail 0.6.3 with connection caching :)

Ready to deploy on relays.

cmdeploy/src/cmdeploy/postfix/main.cf.j2
```diff
@@ -20,7 +20,9 @@ smtpd_tls_key_file={{ config.tls_key_path }}
 smtpd_tls_security_level=may
 
 smtp_tls_CApath=/etc/ssl/certs
-smtp_tls_security_level=verify
+# Local default-transport uses LMTP to filtermail-transport which handles outbound TLS.
+lmtp_tls_security_level = none
+smtp_tls_security_level = none
```
link2xt: Is this change needed? Since locally LMTP is used, can `lmtp_tls_security_level` be set to `none`, but `smtp_tls_security_level` left at `verify` to avoid accidentally sending messages in plaintext if `smtp` service is used?

cmdeploy/src/cmdeploy/postfix/main.cf.j2
```diff
@@ -20,7 +20,9 @@ smtpd_tls_key_file={{ config.tls_key_path }}
 smtpd_tls_security_level=may
 
 smtp_tls_CApath=/etc/ssl/certs
-smtp_tls_security_level=verify
+# Local default-transport uses LMTP to filtermail-transport which handles outbound TLS.
+lmtp_tls_security_level = none
```
link2xt: This also does not need to be specified, default is either `may` or `none`, both are fine.

j-g00da: https://github.com/chatmail/relay/pull/916#discussion_r3101599479

I removed change to smtp_tls_security_level, but left the lmtp one to be explicit

cmdeploy/src/cmdeploy/postfix/main.cf.j2
```diff
@@ -20,6 +20,8 @@ smtpd_tls_key_file={{ config.tls_key_path }}
 smtpd_tls_security_level=may
 
 smtp_tls_CApath=/etc/ssl/certs
+# Local default-transport uses LMTP to filtermail-transport which handles outbound TLS.
+lmtp_tls_security_level=none
```
link2xt: It is not needed, i'd at least move it to `master.cf.j2`, below `lmtp-filtermail`.

link2xt (approved):

link2xt (approved):

hpk42 (changes requested):
we need to find some way to automatically test cross-relay delivery using https.  Right now, also filtermail itself does not contain much tests regarding the new https endpoint.  It may make sense to extend "cmlxc test-mini ..." multi-relay runs, to block certain ports and see if delivery succeeds.

j-g00da:
This PR adds only SMTP channel, http channel is not yet merged and released (https://github.com/chatmail/filtermail/pull/122).

This is still true, but out of scope of this PR.

j-g00da:
Marking as draft for now as we need to implement implicit MX: https://github.com/chatmail/filtermail/issues/126

hpk42 (approved):
ok, then my request-changes comment is misplaced here.  lost track a bit of the various inter-dependent PRs.

hpk42:
but why is this PR still marked as "draft"?  also rebase on master would be good.

j-g00da:
> but why is this PR still marked as "draft"?

https://github.com/chatmail/relay/pull/916#issuecomment-4326894426

j-g00da:
Note: We decided to go with mxdeliv/https channel already in this PR. Bumped filtermail to 0.6.4 that includes it.

j-g00da:
This needs ref swapped in the CI after https://github.com/chatmail/cmlxc/pull/11 is merged and new version of cmlxc is released, but other than that is ready now.

PR #917, opened 2026-04-12
re-use cmlxc workflow, replacing CI with hetzner staging servers with local lxc containers

hpk42:
cmlxc is meanwhile regularly published (via the https://github.com/chatmail/cmlxc repo) via an OIDC-automated github workflow (which means anyone with write rights can create a new release, see README): 

https://pypi.org/project/cmlxc/

The chatmail/cmlxc repo also offers a re-usable "lxc-test" workflow that is used here in this PR, to replace the staging-* hetzner host testing with testing against local containers.   The "cmlxc" commands executed in the github CI should be possible to run unmodified locally to recreate any problems.  

i think it would be sufficient to only daily deploy to "real" servers, maybe laso doing the docker tests there. It is not neccessary for each commit in a PR to cause the whole remote dance and dependency (if something fails in staging, it's not easy to replicate locally). 

A sister PR is also in the madmail repo, also re-using the same lxc-test workflow. 

sidenote: The "chatmail/cmlxc" is still incrementally developing, although i think i finished the largest refactorings i wanted to do.  I'll see to give a sign when i feel the code base is sufficiently stable, and worth a more thorough review. Meanwhile PRs or issue-filing ist welcome.


.github/workflows/ci.yaml
```diff
@@ -38,5 +38,14 @@ jobs:
       - name: run deploy-chatmail offline tests 
         run: pytest --pyargs cmdeploy 
 
-      # all other cmdeploy commands require a staging server 
-      # see https://github.com/deltachat/chatmail/issues/100
+  lxc-test:
+    name: LXC deploy and test
+    uses: chatmail/cmlxc/.github/workflows/lxc-test.yml@main
```
link2xt: This should probably be pinned and check out pinned version internally so it does not break just because cmxlc got a commit pushed. Currently CI is failing.

.github/workflows/ci.yaml
missytake, line 5: I think running this on `pull_request` is enough, no? Otherwise this job is always triggered twice.

.github/workflows/ci.yaml
```diff
@@ -38,5 +38,14 @@ jobs:
...
+    name: LXC deploy and test
+    uses: chatmail/cmlxc/.github/workflows/lxc-test.yml@v0.8.1
+    with:
+      cmlxc_commands: |
+        cmlxc init 
+        cmlxc deploy-cmdeploy --source ./repo -vv cm0 
+        cmlxc test-mini cm0 
+        cmlxc test-cmdeploy -vv cm0 
+        cmlxc deploy-cmdeploy --source ./repo -vv --ipv4-only cm1
+        cmlxc test-cmdeploy -vv cm0 cm1
```
missytake: This is a good start :) I wonder whether the deploys could maybe run concurrently to speed this up a bit - but we should first gather some experience with this new workflow.

.github/workflows/test-and-deploy-ipv4only.yaml
```diff
@@ -1,104 +0,0 @@
...
-          rsync -avz root@staging-ipv4.testrun.org:/var/lib/acme acme-ipv4 || true
-          rsync -avz root@staging-ipv4.testrun.org:/etc/dkimkeys dkimkeys-ipv4 || true
-          # store previous acme & dkim state on ns.testrun.org, if it contains useful certs
-          if [ -f dkimkeys-ipv4/dkimkeys/opendkim.private ]; then rsync -avz -e "ssh -o StrictHostKeyChecking=accept-new" dkimkeys-ipv4 root@ns.testrun.org:/tmp/ || true; fi
-          if [ "$(ls -A acme-ipv4/acme/certs)" ]; then rsync -avz -e "ssh -o StrictHostKeyChecking=accept-new" acme-ipv4 root@ns.testrun.org:/tmp/ || true; fi
-          # make sure CAA record isn't set
-          scp -o StrictHostKeyChecking=accept-new .github/workflows/staging-ipv4.testrun.org-default.zone root@ns.testrun.org:/etc/nsd/staging-ipv4.testrun.org.zone
-          ssh root@ns.testrun.org sed -i '/CAA/d' /etc/nsd/staging-ipv4.testrun.org.zone
-          ssh root@ns.testrun.org nsd-checkzone staging-ipv4.testrun.org /etc/nsd/staging-ipv4.testrun.org.zone
-          ssh root@ns.testrun.org systemctl reload nsd
```
missytake, line 42: ngl, I'm going to miss this caching logic a little bit.

missytake (approved):

.github/workflows/ci.yaml
hpk42: tests are generally very fast, it's the deploys that are slow -- they could run concurrently against several hosts but then the option-handling gets more complex. for now it's simpler, and more helpful for logging and debugging any issues, to run things sequentially.

.github/workflows/ci.yaml
hpk42: i revamped and documented the choices.  I always get confused by these configs.

.github/workflows/ci.yaml
```diff
@@ -1,15 +1,26 @@
-name: CI
+name: Run unit-tests and container-based deploy+test verification
```
j4n, line 1:
```suggestion
name: CI: unit tests + LXC deploy
```
So its always visible in UI

.github/workflows/ci.yaml
```diff
@@ -1,15 +1,26 @@
...
 
 on:
-  pull_request:
+  # Triggers when a PR is merged into main or a direct push occurs
   push:
```
j4n, line 5: maybe we should bring back excluding the non-code files:
```
paths-ignore:
- 'scripts/**'
- '**/README.md'
- 'CHANGELOG.md'
- 'LICENSE'
```

j4n (approved):
Looks good, could not test much; I suppose this will mean ci-testing locally and then having a nightly run of the old-fashioned non-local testing?

.github/workflows/ci.yaml
link2xt: I see workflow is pinned now, but I think checkout inside the workflow is not.

.github/workflows/ci.yaml
missytake: don't forget `doc/**` ;)

hpk42: it's kind of rare that only these files change, so i am not sure it's worth it :)

PR #918, opened 2026-04-13
feat: remove dovecot.index.cache files if larger than 500k 

### LINKED ISSUE #644: Restore in-memory index

missytake:
fix #644 

The dovecot config setting doesn't seem to work; this PR instead expires the dovecot.index.cache files by deleting them regularly.


missytake:
Seems this change is actually not needed: https://github.com/chatmail/relay/issues/644#issuecomment-4240081614

PR #920, opened 2026-04-15
Docker Preparations: remove @docker, consolidate is_in_container()

j4n:
Separating these into a separate PR:
###  cmdeploy: consolidate container detection into is_in_container() helper
    
Both deployers.py (CHATMAIL_NOPORTCHECK env var) and dovecot/deployer.py
(_can_set_inotify_limits function) used separate container detection logic.
Replace both with a single is_in_container() helper in basedeploy.py using
the simpler systemd-detect-virt -c pattern.

###  cmdeploy/sshexec: remove dead @docker SSH host
    
The `@docker` SSH host was added in Docker development to route
cmdeploy commands into a running container from outside. This is no
longer needed because chatmail-init.sh deploys with `@local` inside
the container, and all cmdeploy commands (test, dns, status) work

### DIFF
--- a/cmdeploy/src/cmdeploy/basedeploy.py
+++ b/cmdeploy/src/cmdeploy/basedeploy.py
@@ -3,12 +3,25 @@ import io
 import os
 from contextlib import contextmanager
 
+from pyinfra import host
+from pyinfra.facts.server import Command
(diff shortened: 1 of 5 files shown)


hpk42 (approved):
looks good!

PR #921, opened 2026-04-15
refactor: Rename filtermail_http_port to filtermail_http_port_incoming

j-g00da:
Since http port will be used for MTA-to-MTA,
it should be suffixed with "incoming" for consistency.

This will also make it clearer if we decide to
introduce client-relay http channel in the future.


hpk42 (approved):
fine.  It's also good that this is not part of chatmail.ini

PR #922, opened 2026-04-16
fix(cmdeploy/dovecot): detect stale dovecot binary and force restart in activate()

j4n:
When a previous deploy installed dovecot packages but the restart was blocked (policy-rc.d) or the deploy aborted before activate(), the next deploy sees the correct package version already installed and skips restart. Extend activate() to check /proc/MainPID/exe for "(deleted)" before the restart decision.


missytake (approved):
Looks good, thanks for fixing that test that failed for me locally.

PR #925, opened 2026-04-17
feat(dovecot): disable fsync for LMTP and IMAP services

link2xt:
This is aimed at reducing SSD wear level.
SSDs wear out because of writes
according to <https://superuser.com/a/440219/1777696>,
so anything reducing the writes should be helpful.

For online users Maildir format that we use
results in first storing the message in new/
and then moving to cur/ and then maybe even deleting
it immediately for users with a single device
or bots. Syncing all these changes to disk
is unnecessary and wears SSDs.


hpk42 (approved):
lgtm.

link2xt:
This did not make any visible difference in SSD usage.

Probably most writes are due to postfix writing to the queue, and postfix does not have an option to disable fsync. More specifically, `cleanup` service is doing a lot of writes according to `iotop`. `dovecot/lmtp` ever more, but fsync is already disabled for it with this change.

https://serverfault.com/a/673103/1008846 suggests remounting /var/spool/postfix with `-o nobarrier`. Well, the answer suggests to remount existing partition that postfix queue, but remounting / will likely break something actually important, like not syncing the filesystem after package upgrades.

PR #926, opened 2026-04-18
feat: Automatic per-user quota-keeping  aka "make quota warnings a thing of the past" 

hpk42:
Replace daily timer-based message expire script with Dovecot quota-warning-triggered cleanup. 

This mechanism prevents a user from reaching their quota, staying below the warning level. 

When a user reaches 90% of their mailbox quota Dovecot calls the new script which removes the largest and oldest messages until usage drops below 80%.  Probably those thresholds need tweaking so chatmail clients don't show unneccessary "quota" warnings to users.   

There is no "overall" automatic storage housekeeping but it could potentially be added by having "disk" quota at 90% lower the dovecot per-user quota, thus removing more messages from everyeone.  Or account registration could be closed or so.  This global aspect is out of scope of this PR which only is about making per-user quota management automatic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog for chatmail deployment 
 
+## Unreleased
+
+### Features
+
+- Automated per-user quota-keeping.
+  Replace daily timer-based message expire script
+  with Dovecot quota-warning-triggered cleanup (`chatmail-quota-expire`).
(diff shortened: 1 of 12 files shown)


hpk42:
this quota PR needs to be redone, and become additional I think.  We should not directly  substitute global unconditional expiry with it, but do it in addition for now, and maybe put it to grafana or so. Closing for now.

PR #928, opened 2026-04-20
fix(cmdeploy): replace resolvconf with static resolv.conf

### LINKED ISSUE #541: Unbound / resolv.conf management
### Background

I checked with a few Chatmail operators and there is a mixed experience of having `systemd-resolved` installed. This is an *optional* service that manages `/etc/resolv.conf`. As long as it is installed it can override the values in `/etc/resolv.conf`.

From what I can tell cmdeploy does not directly manage `/etc/resolv.conf` so I do not know how 127.0.0.1 is getting in there.

https://github.com/chatmail/relay/blob/5ba99dc78214808d0e95b86e7e4c37b974b7eef3/cmdeploy/src/cmdeploy/__init__.py#L588-L607

Simply installing and enabling the unbound service does not trigger `systemd-resolved` to do anything. If you really want to use `systemd-resolved` the correct thing to do would be to create a file like `/etc/systemd/resolved.conf.d/unbound.conf` with the following contents:

```
[Resolve]
DNS=127.0.0.1
DNSSEC=yes
```

And then trigger a restart of the `systemd-resolved` service. This will guarantee that the `nameserver 127.0.0.1` line is added to the file as the first/primary nameserver while retaining the default ones that were provided by DHCP. There is another step that also need to be taken: 

`/etc/resolvconf/update.d/unbound` should have its executable bit removed so this hook does not execute, otherwise it will override the Unbound config and force it to forward queries to the DHCP resolvers instead of doing full recursion from the root nameservers. You risk losing DNSSEC as a result which is still important for us as we validate DKIM.

### Recommendation

There isn't a clear need for the additional capabilities that `systemd-resolved` provides. I suggest we remove the package and directly manage the contents of `/etc/resolv.conf` so we are confident what the behavior is. Either way we should be more explicit about how the DNS resolvers are configured as not all servers appear to have `systemd-resolved` installed.

j4n:
1e376f7 explicitly installed resolvconf to fix DNS breakage after ff541b8 disabled APT recommends. But resolvconf adds dynamic resolver management and is unnecessary on a server.

Write a static resolv.conf with unbound as primary and 9.9.9.9 as fallback. Purge resolvconf on existing relays. The pre-deploy guard that added 9.9.9.9 when unbound was not on port 53 is now redundant: the static file always contains the fallback, and the port check at deploy_chatmail() would fail the deploy anyway if something else held port 53.

Also closes https://github.com/chatmail/relay/issues/541.

### DIFF
--- a/cmdeploy/src/cmdeploy/deployers.py
+++ b/cmdeploy/src/cmdeploy/deployers.py
@@ -150,18 +150,40 @@ class UnboundDeployer(Deployer):
         self.need_restart = False
 
     def install(self):
-        # Run local DNS resolver `unbound`. `resolvconf` takes care of
-        # setting up /etc/resolv.conf to use 127.0.0.1 as the resolver.
-
         # On an IPv4-only system, if unbound is started but not configured,
         # it causes subsequent steps to fail to resolve hosts.
         with blocked_service_startup():
             apt.packages(
                 name="Install unbound",
-                packages=["unbound", "unbound-anchor", "dnsutils", "resolvconf"],
+                packages=["unbound", "unbound-anchor", "dnsutils"],
             )
 
     def configure(self):
+        # Remove dynamic resolver managers that compete for /etc/resolv.conf.
+        apt.packages(
+            name="Purge resolvconf",
+            packages=["resolvconf"],
+            present=False,
+            extra_uninstall_args="--purge",
+        )
+        # systemd-resolved can't be purged due to dependencies; stop and mask.
+        server.shell(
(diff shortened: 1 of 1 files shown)


j-g00da (approved):

missytake:
Related: https://github.com/chatmail/relay/issues/541#issuecomment-2798811012

I think it's a good approach to finally settle these problems. What about systemd-resolved, can this also be in the way?

missytake (approved):
Looks good :)

PR #929, opened 2026-04-20
feat: automatic oldest-first message removal from mailboxes to always stay under max_mailbox_size

hpk42:
Revamped #927 , trying to address all suggestions from there. 

Both dovecot-quota-threshold triggers and the daily expiry routine will expunge oldest messages from mailboxes automatically when the mailbox reaches 75% of max_mailbox_size.

Delta Chat users should not see any warnings (at 80/95 percent) or bounce messages, and existing over-quota mailboxes should start receiving mails again.

### DIFF
--- a/chatmaild/src/chatmaild/expire.py
+++ b/chatmaild/src/chatmaild/expire.py
@@ -4,17 +4,26 @@ Expire old messages and addresses.
 """
 
 import os
+import re
 import shutil
 import sys
 import time
 from argparse import ArgumentParser
 from collections import namedtuple
 from datetime import datetime
+from pathlib import Path
 from stat import S_ISREG
 
 from chatmaild.config import read_config
 
 FileEntry = namedtuple("FileEntry", ("path", "mtime", "size"))
+QuotaFileEntry = namedtuple("QuotaFileEntry", ("mtime", "quota_size", "path"))
+
+# Quota cleanup factor of max_mailbox_size. The mailbox is reset to this size.
+QUOTA_CLEANUP_FACTOR = 0.7
+
+# e.g. "cur/1775324677.M448978P3029757.exam,S=3235,W=3305:2,S"
+_dovecot_fn_rex = re.compile(r".+/(\d+)\..+,S=(\d+)")
 
 
 def iter_mailboxes(basedir, maxnum):
@@ -72,6 +81,42 @@ class MailboxStat:
                     if name == "password":
                         self.last_login = entry.mtime
         self.extrafiles.sort(key=lambda x: -x.size)
+
+
+def parse_dovecot_filename(relpath):
+    m = _dovecot_fn_rex.match(relpath)
+    if not m:
+        return None
+    return QuotaFileEntry(int(m.group(1)), int(m.group(2)), relpath)
+
+
+def scan_mailbox_messages(mbox):
+    messages = []
+    for sub in ("cur", "new"):
+        for name in os_listdir_if_exists(mbox / sub):
+            if entry := parse_dovecot_filename(f"{sub}/{name}"):
+                messages.append(entry)
+    return messages
+
+
+def expire_to_target(mbox, target_bytes):
+    messages = scan_mailbox_messages(mbox)
+    total_size = sum(m.quota_size for m in messages)
+    # Keep recent 24 hours of messages protected from expiry because
+    # likely something is wrong with interactions on that address
+    # and quota-full signal can help the address owner's device to notice it
+    undeletable_messages_cutoff = time.time() - (3600 * 24)
+    removed = 0
+    for entry in sorted(messages):
+        if total_size <= target_bytes:
+            break
+        if entry.mtime > undeletable_messages_cutoff:
+            break
+        (mbox / entry.path).unlink(missing_ok=True)
+        total_size -= entry.quota_size
+        removed += 1
+
+    return removed
 
 
 def print_info(msg):
@@ -143,6 +188,19 @@ class Expiry:
             else:
                 continue
             changed = True
+
+        target_bytes = (
+            self.config.max_mailbox_size_mb * 1024 * 1024 * QUOTA_CLEANUP_FACTOR
+        )
+        removed = expire_to_target(Path(mbox.basedir), target_bytes)
+        if removed:
+            changed = True
+            self.del_files += removed
+            if self.verbose:
+                print_info(
+                    f"quota-expire: removed {removed} message(s) from {mboxname}"
+                )
+
         if changed:
             self.remove_file(f"{mbox.basedir}/maildirsize")
 
@@ -154,9 +212,9 @@ class Expiry:
         )
 
 
-def main(args=None):
+def daily_expire_main(args=None):
     """Expire mailboxes and messages according to chatmail config"""
-    parser = ArgumentParser(description=main.__doc__)
+    parser = ArgumentParser(description=daily_expire_main.__doc__)
     ini = "/usr/local/lib/chatmaild/chatmail.ini"
     parser.add_argument(
         "chatmail_ini",
@@ -202,5 +260,33 @@ def main(args=None):
     print(exp.get_summary())
 
 
-if __name__ == "__main__":
-    main(sys.argv[1:])
+def quota_expire_main(args=None):
+    """Remove mailbox messages to stay within a megabyte target.
+
+    This entry point is called by dovecot when a quota threshold is passed.
+    """
+
+    parser = ArgumentParser(description=quota_expire_main.__doc__)
+    parser.add_argument(
+        "target_mb",
+        type=int,
+        help="target mailbox size in megabytes",
+    )
+    parser.add_argument(
+        "mailbox_path",
+        type=Path,
+        help="path to a user mailbox",
+    )
+    args = parser.parse_args(args)
+
+    target_bytes = args.target_mb * 1024 * 1024
+
+    removed_count = expire_to_target(args.mailbox_path, target_bytes)
+    if removed_count:
+        (args.mailbox_path / "maildirsize").unlink(missing_ok=True)
+        print(
+            f"quota-expire: removed {removed_count} message(s)"
+            f" from {args.mailbox_path.name}",
+            file=sys.stderr,
+        )
+    return 0
--- a/chatmaild/src/chatmaild/tests/test_expire.py
+++ b/chatmaild/src/chatmaild/tests/test_expire.py
@@ -1,5 +1,7 @@
+import itertools
 import os
 import random
+import time
 from datetime import datetime
 from fnmatch import fnmatch
 from pathlib import Path
@@ -9,12 +11,18 @@ import pytest
 from chatmaild.expire import (
     FileEntry,
     MailboxStat,
+    expire_to_target,
     get_file_entry,
     iter_mailboxes,
     os_listdir_if_exists,
+    parse_dovecot_filename,
+    quota_expire_main,
+    scan_mailbox_messages,
 )
-from chatmaild.expire import main as expiry_main
+from chatmaild.expire import daily_expire_main as expiry_main
 from chatmaild.fsreport import main as report_main
+
+MB = 1024 * 1024
 
 
 def fill_mbox(folderdir):
@@ -196,3 +204,51 @@ def test_os_listdir_if_exists(tmp_path):
     tmp_path.joinpath("x").write_text("hello")
     assert len(os_listdir_if_exists(str(tmp_path))) == 1
     assert len(os_listdir_if_exists(str(tmp_path.joinpath("123123")))) == 0
+
+
+# --- quota expire tests ---
+
+_msg_counter = itertools.count(1)
+
+
+def _create_message(basedir, sub, size, days_old=0, disk_size=None):
+    seq = next(_msg_counter)
+    mtime = int(time.time() - days_old * 86400)
+    name = f"{mtime}.M1P1Q{seq}.hostname,S={size},W={size}:2,S"
+    path = basedir / sub / name
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_bytes(b"x" * (disk_size if disk_size is not None else size))
+    os.utime(path, (mtime, mtime))
+    return path
+
+
+def test_parse_dovecot_filename():
+    e = parse_dovecot_filename("cur/1775324677.M448978P3029757.exam,S=3235,W=3305:2,S")
+    assert e.path == "cur/1775324677.M448978P3029757.exam,S=3235,W=3305:2,S"
+    assert e.mtime == 1775324677
+    assert e.quota_size == 3235
+    assert parse_dovecot_filename("cur/msg_without_structure") is None
+
+
+def test_expire_to_target(tmp_path):
+    _create_message(tmp_path, "cur", MB, days_old=10, disk_size=100)
+    _create_message(tmp_path, "new", MB, days_old=5)
+    _create_message(tmp_path, "cur", MB, days_old=0)  # undeletable (<1 hour)
+    assert len(scan_mailbox_messages(tmp_path)) == 3
+    # removes oldest first, uses S= size not disk size
+    removed = expire_to_target(tmp_path, MB)
+    assert removed == 2
+    msgs = scan_mailbox_messages(tmp_path)
+    assert len(msgs) == 1
+    # the surviving message is the fresh undeletable one
+    assert msgs[0].mtime > time.time() - 3600
+
+
+def test_quota_expire_main(tmp_path, capsys):
+    mbox = tmp_path / "user@example.org"
+    _create_message(mbox, "cur", 2 * MB, days_old=5)
+    (mbox / "maildirsize").write_text("x")
+    quota_expire_main([str(1), str(mbox)])
+    _, err = capsys.readouterr()
+    assert "quota-expire: removed 1 message(s) from user@example.org" in err
+    assert not (mbox / "maildirsize").exists()
--- a/chatmaild/pyproject.toml
+++ b/chatmaild/pyproject.toml
@@ -21,7 +21,8 @@ where = ['src']
 [project.scripts]
 doveauth = "chatmaild.doveauth:main"
 chatmail-metadata = "chatmaild.metadata:main"
-chatmail-expire = "chatmaild.expire:main"
+chatmail-expire = "chatmaild.expire:daily_expire_main"
+chatmail-quota-expire = "chatmaild.expire:quota_expire_main"
 chatmail-fsreport = "chatmaild.fsreport:main"
 lastlogin = "chatmaild.lastlogin:main"
 turnserver = "chatmaild.turnserver:main"
--- a/chatmaild/src/chatmaild/config.py
+++ b/chatmaild/src/chatmaild/config.py
@@ -95,6 +95,11 @@ class Config:
         # old unused option (except for first migration from sqlite to maildir store)
         self.passdb_path = Path(params.get("passdb_path", "/home/vmail/passdb.sqlite"))
 
+    @property
+    def max_mailbox_size_mb(self):
+        """Return max_mailbox_size as an integer in megabytes."""
+        return parse_size_mb(self.max_mailbox_size)
+
     def _getbytefile(self):
         return open(self._inipath, "rb")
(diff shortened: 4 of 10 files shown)


chatmaild/src/chatmaild/expire.py
```diff
@@ -4,17 +4,26 @@
...
 from argparse import ArgumentParser
 from collections import namedtuple
 from datetime import datetime
+from pathlib import Path
 from stat import S_ISREG
 
 from chatmaild.config import read_config
 
 FileEntry = namedtuple("FileEntry", ("path", "mtime", "size"))
+QuotaFileEntry = namedtuple("QuotaFileEntry", ("mtime", "quota_size", "path"))
```
missytake, line 20: wouldn't it be more straightforward if both namedtuples had the values in order of `path`, `mtime`, `size`?

chatmaild/src/chatmaild/expire.py
```diff
@@ -143,6 +182,19 @@ def process_mailbox_stat(self, mbox):
             else:
                 continue
             changed = True
+
+        target_bytes = (
+            self.config.max_mailbox_size_mb * 1024 * 1024 * QUOTA_CLEANUP_FACTOR
+        )
+        removed = expire_to_target(Path(mbox.basedir), target_bytes)
```
j4n, line 189: this calls `scan_mailbox_messages` and rescans the mailbox, we could reuse the information from `mbox.messages` to save on IO?

j4n (changes requested):
looks good, tested overfilling a mailbox with 
```
for i in $(seq 1 400); do   dd if=/dev/urandom bs=1M count=1 2>/dev/null | base64 |     (echo "From: test@example.org"; echo "To: test@do.main"; echo "Subject: fill $i"; echo; cat)     | doveadm save -u test@do.main; done
```
Running `usr/local/lib/chatmaild/venv/bin/chatmail-quota-expire 70 **/home/vmail/mail/...` worked, but when looking at the logs, it did not fire automatically:
```
doveadm(test@do.main): Error: program unix:/run/dovecot/quota-warning: net_connect_unix(/run/dovecot/quota-warning) failed: Permission denied (euid=999(vmail) egid=996(vmail) missing +r perm: /run/dovecot/quota-warning, dir owned by 0:0 mode=0755)
doveadm(test@do.ain): Error: Saving failed: Quota exceeded (mailbox for user is full)
...
```

`cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2`, the `unix_listener quota-warning` socket needs explicit permissions for vmail to connect:

```diff
   unix_listener quota-warning {
+    user = vmail
+    mode = 0600
   }
```

Then it works: 
```
Apr 20 13:07:14 host dovecot[3834579]: quota-warning: Error: quota-expire: removed 6 message(s) from test@do.main
```

chatmaild/src/chatmaild/expire.py
```diff
@@ -74,6 +83,36 @@ def scandir(self, folderdir):
...
+            if entry := parse_dovecot_filename(f"{sub}/{name}"):
+                messages.append(entry)
+    return messages
+
+
+def expire_to_target(mbox, target_bytes):
+    messages = scan_mailbox_messages(mbox)
+    total_size = sum(m.quota_size for m in messages)
+    removed = 0
+    for entry in sorted(messages):
```
missytake, line 106: Ah, because they are sorted by their first value here and it has to be mtime?

missytake (approved):
Sounds very promising!

Maybe it would make sense to first expire larger messages by date, and only then the small messages; missing an attachment is much more harmless than missing membership changes.

chatmaild/src/chatmaild/expire.py
hpk42: yes, mtime is the sorting key so i put it first in the tuple, to avoid having to do "key = lambda x: x.mtime" in the sort function.  named tuples are otherwise accessed by named attribute access, so it doesn't otherwise matter much.

hpk42:
In a prior version of the quota-expire effort, i had file-size play into the sorting, so larger messages would be deleted first if they are not much younger than small messages.  @link2xt was skeptical of that and eventually we arrived at saying: "let's just do oldest first for now" also because it's easier to understand it and reason about it. 

However, i am still wondering about it. Maybe we can keep the last 7 days of small messages unconditionally, so a few big messages can not push all the young small messages out of the queue?  

sidenote: with this PR merged, addresses conceptually become handles for message *queues* rather than static mail storage.

chatmaild/src/chatmaild/expire.py
hpk42: it's cached by the kernel, and expire_to_target only lists directory entries, no stat() calls.  So not worth to mix it with or refactor the other (working) code.   It could make sense to make daily expire also use the dovecot filename parsing, to avoid stat() calls touching the disk.

link2xt:
> Maybe it would make sense to first expire larger messages by date, and only then the small messages; missing an attachment is much more harmless than missing membership changes.

In https://github.com/chatmail/relay/pull/926 messages were scored by age multiplied by size. Could work, but then if you have really not read messages for some time, you will get some history but no "member added" messages and randomly dropped messages at the time when gossip happens instead of the history starting at roughly some timestamp. Can drop messages with `Chat-Is-Post-Message` header first or just very large messages first (likely videos) followed by all messages, but just dropping by date is likely good enough already.

> Maybe we can keep the last 7 days of small messages unconditionally, so a few big messages can not push all the young small messages out of the queue?

Something simple with two thresholds is probably fine, first delete huge messages (videos), then if it is not sufficient the same as the current logic. But then still not clear if anyone will actually notice the improvement, but you lose the "queue" property and it might be difficult to reimplement this logic if we get to replacing Maildir with something more efficient (chatmail basically needs a log storage, just write new messages into mbox and rotate "pages").

j4n (approved):
lgtm

chatmaild/src/chatmaild/tests/test_expire.py
```diff
@@ -196,3 +203,46 @@ def test_os_listdir_if_exists(tmp_path):
...
     assert len(os_listdir_if_exists(str(tmp_path))) == 1
     assert len(os_listdir_if_exists(str(tmp_path.joinpath("123123")))) == 0
+
+
+# --- quota expire tests ---
+
+
+def _create_message(basedir, sub, size, days_old=0, disk_size=None):
+    mtime = int(time.time() - days_old * 86400)
+    name = f"{mtime}.M1P1.host,S={size},W={size}:2,S"
```
link2xt: `M1P1` is not unique and `write_bytes` may overwrite the file if you try to create two files with the same size and same `days_old`

I'd just create a global variable with a number that increases each time `_create_message` is called and put it instead of `M1P1` to make sure the filename is never the same.
For "correctness" of the maildir format from https://cr.yp.to/proto/maildir.html it should be preceded with `Q`

chatmaild/src/chatmaild/expire.py
```diff
@@ -4,17 +4,26 @@
...
 
 from chatmaild.config import read_config
 
 FileEntry = namedtuple("FileEntry", ("path", "mtime", "size"))
+QuotaFileEntry = namedtuple("QuotaFileEntry", ("mtime", "quota_size", "path"))
+
+# Quota cleanup factor of max_mailbox_size. The mailbox is reset to this size.
+QUOTA_CLEANUP_FACTOR = 0.7
+
+# e.g. "cur/1775324677.M448978P3029757.nine,S=3235,W=3305:2,S"
```
link2xt: I'd replace `nine` with whatever example hostname or `hostname`, `nine` is already way too special by being a default relay.

chatmaild/src/chatmaild/expire.py
link2xt: does not look resolved but marked as resolved, not sure if by accident or just means wontfix

chatmaild/src/chatmaild/expire.py
```diff
@@ -74,6 +83,36 @@ def scandir(self, folderdir):
...
+def parse_dovecot_filename(relpath):
+    m = _dovecot_fn_rex.match(relpath)
+    if not m:
+        return None
+    return QuotaFileEntry(int(m.group(1)), int(m.group(2)), relpath)
+
+
+def scan_mailbox_messages(mbox):
+    messages = []
+    for sub in ("cur", "new", "tmp"):
```
link2xt: `tmp` should not be scanned: it is likely empty and otherwise contains messages that are in the process of being written so has only the new files. By the time you try to delete the file, it will already be moved. If there are stale entries because dovecot crashed, i guess dovecot already cleans up stale "tmp" folders somehow by itself.

So scanning `tmp` just makes the process slower by accessing another dir.

chatmaild/src/chatmaild/expire.py
missytake: Resolved by https://github.com/chatmail/relay/pull/929#discussion_r3115566370

chatmaild/src/chatmaild/expire.py
```diff
@@ -74,6 +83,36 @@ def scandir(self, folderdir):
...
+                messages.append(entry)
+    return messages
+
+
+def expire_to_target(mbox, target_bytes):
+    messages = scan_mailbox_messages(mbox)
+    total_size = sum(m.quota_size for m in messages)
+    removed = 0
+    for entry in sorted(messages):
+        if total_size <= target_bytes:
```
link2xt, line 111: One suggestion (from Ellie) is to unconditionally not delete messages that are new to make complete flushing of the mailbox noticeable:
> there could also be a mixed mode, for example only delete old messages if they have been around at least X day

Maybe stop here not only if we reached quota, but also if we reached the message that is e.g. 2 days old. Then if you get a flood of messages, you start using the remaining 30% and user will get 80% quota warning. And if it happens persistently, then you eventually get over quota and this does not resolve, but this actually means something has gone wrong and should be investigated.

tl;dr the suggestion is to add `or entry.mtime > time.time() - 2 * 86400` here and break out of the loop if we reached very new messages. If we eat through 400 MB of storage in two days, something has likely went really wrong and warning the user sounds reasonable.

hpk42: currently, if an attacker burst-floods a mailbox with small messages, expire_to_target will  delete them. when the burst is over, the address can receive messages again.   

with a 2-day floor, a single burst of small messages fills the mailbox and none of them can be removed for 2 days. Dovecot rejects all incoming mail during that window. the attacker just repeats the burst every 2 day, minimal effort, permanent denial of service.

ell1e: But it would be a visible denial of service, rather than a silent one, where you flush the messages out without a limit. I think for small groups that's better. For larger groups, the errors could be hidden, for example.

I think this is simply a needs of the large room vs needs of the small room scenario. In small rooms, just about anything seems preferrable to me to silent message loss. And whatever path is taken, that some messages won't reach a user if an attacker does something nasty, is unavoidable, so you can only control how visible that is.

chatmaild/src/chatmaild/expire.py
hpk42: removed

chatmaild/src/chatmaild/tests/test_expire.py
hpk42: fixed

chatmaild/src/chatmaild/expire.py
hpk42: it's not about visible versus invisible denial of service.  When you get flooded with messages, you will notice that in either case.   Only in the remove-oldest algo  you start receiving messages when the DoS stops, whereas with 2-day floor you stay incommunicado for two days after the first burst.

link2xt: DoS attack is pretty much hypothetical anyway, you can do more annoying stuff with less effort if you really want. More likely is some accidental flooding e.g. bot flooding the bot mailbox where nobody looks or users flooding their mailbox with invisible webxdc updates or subscribing to a channel that sends them videos every day. If we do this change it will either be pretty much never triggered or might uncover some problem if mailbox gets over quota and user actually investigates and reports it, but most likely it will just never get triggered because eating through 400 mb in two days accidentally is unlikely.

One last suggestion from me is adding `or entry.mtime > time.time() - 3600` or so, to not rotate through the mailbox continuously in case something is really-really broken and mailbox gets a new 10MB message every minute, but if not then just mark it as resolved.

link2xt (approved):
There is a seemingly unrelated CI failure, from IP-based relays.

chatmaild/src/chatmaild/expire.py
ell1e: > it's not about visible versus invisible denial of service. When you get flooded with messages, you will notice that in either case

Perhaps, but in that case the sender doesn't notice. Sender visibility of a sending failure in my opinion is even more important than the receiver noticing that something is up. Even if I notice on the receiving end, I can't just tell all the 50 people that ever talked to me that if they messaged me they need to resend. It doesn't scale. With my proposal, the sender will notice and will know to resend if it was important. I consider that a fairly essential quality of messengers, and I don't think it's just me (but if it is, ignore me).

3600 seconds honestly seems too short for that purpose. I think it's safe to assume most people will login every 1-2 days, but not that they will be online every hour.

hpk42: k, i added the 1h-guard as suggested.  https://github.com/chatmail/relay/pull/929/commits/80ccf4de839622323598fd89eca21b1a7a1f439d

addendum: only now saw ell1e's 6-hour ago comment for some reason, this message here was originally next to link2xt's last comment in the thread.

hpk42: > Sender visibility of a sending failure in my opinion is even more important than the receiver noticing that something is up. 

When a sender transmits a message to 3 addresses (a multi-relay chat contact), and one address fails because of quota overrun, it is the receiving device which should notice quota-full event and do something about it, like clearing the storage or switching to a different relay.  For users nothing in particular needs to happen.  No message needs to be flagged, no warning needs showing.  Everything works, and auto-repairs to continue to work.   <------ this is at least the rough UX direction of current discussions about "automated relay management" designed to work for mass-users who might not know any details of the relay network, but just onboard on it.

link2xt: > addendum: only now saw ell1e's 6-hour ago comment for some reason

GitHub broke the caching logic for messages now, it was already broken for issue counters and pull request counters recently. I also get emails about your messages, then go to the page and have to refresh it manually to see the message. Also when I reload the page, i don't see any comments at first without any spinner, it is indistinguishable from the PR having no comments so i always think for a moment i opened the wrong PR or something, and the comments load only later, not necessary in the latest state, without the new comment just received by email sometimes.

ell1e: I don't think multi relay is relevant for this problem, since an attacker could just spam all relays. (Edit: or am I missing something? I'm happy to be wrong! :heart: )

The only way to know that my message wasn't eaten by some spam a few hours after, while the user wasn't online yet, seems to be if there is a time window of at least 48 hours or so where it's guaranteed to be stored. That way it would actually arrive for most people. Otherwise you're allowing silent failures even though everything looked fine.

I honestly don't know if it's just me, but I abandoned multiple messengers *just* because of silent delivery failures. (That being XMPP OMEMO, as well as Wire messenger.) A single message not arriving without anybody being aware in a tense situation can be socially nuclear.

ell1e: Oh another scenario I just realized is relevant, I've had multiple people be offline for a week or two at some point. If you don't have a minimum retention time of two weeks, you would be able to write them really detailed messages in private while they might never arrive and nobody involved might ever know. While with the old configuration, you absolutely would know once their inbox ran full that both the new messages didn't arrive, and that the old messages would typically have been retained until read.

I realize this is an extreme case and it might not make sense to tweak this for the extreme cases. However, I also think the assumption every user is offline at most for 60 minutes is the opposite extreme. People might e.g. be in a building without reception for half the day, that's pretty common. Or their battery simply ran out and they were too busy to deal with it for a day. It happens to the best of us...

link2xt: > I don't think multi relay is relevant for this problem, since an attacker could just spam all relays. (Edit: or am I missing something? I'm happy to be wrong! ❤️ )

There is a parallel discussion about handling NDNs (non-delivery notifications), the way they are displayed on messages and what users can do about them. With multi-relay getting an NDN that says "this mailbox is over quota" does not necessarily mean that the message was not received when there are other addresses which did not send an NDN back.

The only way to really know the message was not eaten by some middlebox is to get a positive end-to-end confirmation from the receiver. E.g. Signal and WhatsApp has one checkmark for delivery receipt, another checkmark for read receipt, and not just one checkmark. This however has privacy concerns, see https://support.delta.chat/t/careless-whisper-on-deltachat/4396 and https://anarc.at/blog/2018-07-27-signal-metadata/

link2xt: Messages also don't arrive for other reasons, more likely than the message being delivered and get flushed with this PR logic. There are DNS problems with the message getting stuck in a "deferred" queue for a week, there are spam filters which move messages to spam folder or silently eat them on purpose, people may overlook the message because of ordering problems (https://github.com/chatmail/core/issues/8113) or just so even when the message arrives.

ell1e: I'm honestly unsure what you're trying to argue, "silently dropped messages are already possible so perhaps it's not too bad if we make it worse"? I'm not sure I agree.

To me, the difference of the points you mention vs the scenario I'm describing is 1. mine is definitely not a hypothetical (yours might also not be I'm not sure, but I'm just saying I haven't seen all of those), and 2. has happened multiple times in rooms I've been in even without anybody actively pushing it, both mailboxes running full and somebody not having been online for a few days, and 3. is relatively obvious to trigger by somebody who isn't even a man in the middle and who wants to cause problems.

ell1e: Perhaps this table is helpful to illustrate why I don't understand why you wouldn't put a 48 hour minimum retention:

|  | Scenario of somebody deliberately spams a target to try to induce message loss *(if multiple relays are used then the attacker spams all of them)*|Scenario of message box runs full accidentally in a busy room, with person being offline for a day or two *(which worst case can happen to all relays used at once)*|Scenario of other issues like DNS or spam filters causing lost messages|
|-|-|-|-|
|With minimum 48 hour message retention:| :warning: New messages might be noisily blocked, but if user is online every 48 hours then no messages are **silently** lost no matter how much attacker tries to spam| :heavy_check_mark: Solved, since pretty much no room is going to fill up the inbox as fast as 48 hours. Therefore, rotation should work without noisy blocks of new messages| :no_entry_sign: Broken and essentially unaffected |
|Without minimum 48 hour message retention:| :no_entry_sign: New messages might be noisily blocked if the volume exceeds the 60 minutes time window, but you'll also have silent message loss on top of it | :heavy_check_mark: Solved | :no_entry_sign: Broken and essentially unaffected |

link2xt: > I'm honestly unsure what you're trying to argue, "silently dropped messages are already possible so perhaps it's not too bad if we make it worse"? I'm not sure I agree.

I was basically arguing that the leftmost column is much more rare than third column and other reasons combined. At some point it is better that we move to thinking about fixing other problems that cause missing messages, like reordering and blocked internet connections and what we do about them.

I'm fine with 1 hour or 48 hours or whatever, just don't put 10 days or so there so the PR is actually working.
I also found a good constant here - 36 hours: https://doc.dovecot.org/2.3/settings/core/#core_setting-mail_temp_scan_interval

ell1e: For what it's worth, I understand now and agree something like 10 days or bigger will likely defy the point of what this new mechanism is trying to solve. 48h sounds best to me personally.

However, I was made aware that I might be missing some context here, so I don't want to distract anymore and cause more unnecessary noise! Thank you for your thoughtful responses! :heart: I don't want to take up more of anyone's time.

hpk42:
ok, i set the unconditionally retain time to 24h for now. merging.

r10s:
i created a PR to remove the quota warning at 80% at https://github.com/chatmail/core/pull/8168

missytake:
We now have some weeks of data about this improvement :)

image

PR #932, opened 2026-04-22
feat: do not bind SMTP client sockets to public addresses

link2xt:
This change reverts 06560dd0710167d1e819f10b3e8fe1a6285e7a6c

Main reason for using the same address for sending as the one used in DNS is to pass FCrDNS
(forward-confirmed reverse DNS) checks:
IP address used by SMTP client should resolve
to the domain which in turn resolves to the same IP. chatmail relays don't do check reverse DNS
for incoming connections,
but other email servers may do and reject email
if the check does not pass.

Most chatmail relays only have one IP address per address family, so this configuration does not change anything.

For chatmail relays that have multiple addresses
and only publishing one IP to DNS,
source address used for outgoing SMTP connections
should be the public IP.

This can be ensured by configuring the source
address in the routing table,
e.g. with the `src` argument
to `ip route add/change/replace` command.

Solving this by binding SMTP client address
on the application level prevents chatmail relays
from configuring alternative routes.

Besides, some chatmail relays are NATed
and NAT is responsible for translating the address to the public one, in which case using `smtp_bind_address_enforce`
will result in unnecessarily deferring all mails.

### DIFF
--- a/chatmaild/src/chatmaild/config.py
+++ b/chatmaild/src/chatmaild/config.py
@@ -1,4 +1,3 @@
-import os
 from pathlib import Path
 
 import iniconfig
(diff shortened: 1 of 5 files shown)


link2xt:
Draft because I still want to add some docs.

j-g00da (approved):

ccclxxiii (approved):

missytake (approved):
Thanks for the extensive documentation!

(If anyone wants to check: https://staging.chatmail.at/doc/relay/932/reverse_dns.html )

link2xt:
Updated the documentation a bit to mention IP reputation which is apparently the main reason for floating IP.

link2xt:
Merged it already because it seems unlikely that we decide the approach of using routing table for multi-homed setups does not work. Worst case we can revert it.

PR #935, opened 2026-04-23
for last 7 days of messages remove large messages first, then by age

hpk42:
This prevents a few fresh big messages from clogging the inbox and pushing all small messages out.  If an attacker wants to clog/ddos a recipient address they now have to constantly send many small messages to prevent small messages from arriving to the recipient.  Just sending a few large ones does not suffice.


hpk42:
maybe it's not worth it for now, though -- i'll keep the branch around for future reference.

PR #939, opened 2026-04-29
Add allow_account_autocreation parameter

localzet:


chatmaild/src/chatmaild/doveauth.py
```diff
@@ -27,6 +29,10 @@ def encrypt_password(password: str):
 
 def is_allowed_to_create(config: Config, user, cleartext_password) -> bool:
     """Return True if user and password are admissable."""
+    if not config.allow_account_autocreation:
+        logging.warning("blocked account creation because allow_account_autocreation is false")
+        return False
+
```
missytake, line 35: Instead of checking the config, I suggest to instead let `cmdeploy run` create or remove the `NOCREATE_FILE` depending on whether it is `true`, `false`, or don't change it if the setting is not set.

Otherwise, setting `allow_account_autocreation` to `false` will not change anything if the `NOCREATE_FILE` is present; and removing the `NOCREATE_FILE` will not enable creating accounts if `allow_account_autocreation` is `false.`

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -43,6 +49,15 @@ def is_allowed_to_create(config: Config, user, cleartext_password) -> bool:
         logging.warning(f"user {user!r} is not a proper e-mail address")
         return False
     localpart, domain = parts
+    normalized_user = user.lower()
+    normalized_domain = domain.lower()
+
+    if is_blocked_address(normalized_user):
+        logging.warning("blocked account creation for banned address %s", normalized_user)
+        return False
```
missytake, line 57: What do you need this feature for?

If you want to ban an address from being created, I suggest to do `mkdir /home/vmail/mail/{domain}/address@{domain} && echo "asdf" > /home/vmail/mail/{domain}/address@{domain}/password`.

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -43,6 +49,15 @@ def is_allowed_to_create(config: Config, user, cleartext_password) -> bool:
...
+    if is_blocked_domain(normalized_domain):
+        logging.warning("blocked account creation for banned domain %s", normalized_domain)
+        return False
```
missytake, line 60: You can anyway only create addresses with the `mail_domain_deliverable` so far, blocking a domain will not change anything; if you want to block people from creating addresses with your `mail_domain_deliverable`, just run `touch /etc/chatmail-nocreate`

missytake (commented):
Right now, I don't see much benefit from this PR - we like to keep config options to a minimum, and a centralized file which tracks forbidden addresses is rather expensive performance-wise, compared to a failing password check.

Making `touch /etc/chatmail-nocreate` also available as a config option might be more friendly to some users:

hpk42:
I concur with @missytake 's comments, and also test are missing.  This PR as is does not do something that can't be done already with other means.  Giving more control to operators about how addresses are created is a worthwhile topic, but it shouldn't be done in an ad-hoc way.

PR #944, opened 2026-04-30
fix: make www tests work with editable instead of just plain installs

hpk42:


link2xt (approved):

cmdeploy/src/cmdeploy/tests/test_helpers.py
```diff
@@ -1,11 +1,10 @@
-import importlib.resources
+from pathlib import Path
 
 from cmdeploy.www import build_webpages
 
 
 def test_build_webpages(tmp_path, make_config):
-    pkgroot = importlib.resources.files("cmdeploy")
-    src_dir = pkgroot.joinpath("../../../www/src").resolve()
+    src_dir = (Path(__file__).resolve().parent / "../../../../www/src").resolve()
```
link2xt: probably can add another "../" instead of `.parent` and not mix them

hpk42: changed -- even though i have used this pattern for some decades (getting dirpath then joining) :)

PR #953, opened 2026-05-06
fix(cmdeploy): stop and disable unbound-resolvconf

punkero-org:
Commit 825831e purges resolvconf, however the unbound service activates a 'wants' unit for async resolvconf updates. This results in errors in systemd startup as the unit will now always fail.

Stop and disable the unbound-resolvconf unit activation


punkero-org:
Side note: this is a trivial issue, it was only discovered as I was rebooting the relay to the newly released Debian kernel upgrade and reviewing my logs. The failure itself does not result in any changes to the system, /etc/resolv.conf stays as intended

missytake (approved):
Thanks!

PR #955, opened 2026-05-07
WIP: Try ns 851 change

hpk42:


cmdeploy/src/cmdeploy/remote/rdns.py
```diff
@@ -64,21 +64,26 @@ def get_dkim_entry(mail_domain, pre_command, dkim_selector):
...
+    filtered_replies = [a for a in ns_replies if len(a) >= 5 and a[3] in ("SOA", "NS")]
+    if not filtered_replies:
         return
-    ns = soa[0][4]
+    return filtered_replies[0][4]
+
+
+def query_dns(typ, domain):
+    apex = ".".join(domain.split(".")[-2:])
+    ns = get_authoritative_ns(apex)
```
missytake: tricky - e.g. for staging2.testrun.org the authoritative NS is ns.testrun.org, while it's hetzner (and hidden primary ns.merlinux.eu) for testrun.org...

hpk42: fair point -- i was still trying to get it to work at all.  but i think i found the culprit (somewhere deep in cmlxc powerdns config), and cmlxc 0.13.5 should fix it. We'll see when the CI finishes.

PR #919, opened 2026-04-14
feat: support setup with only an IPv4 address, but no domain

missytake:
replaces #894
fix https://github.com/chatmail/relay/issues/936

This is a slightly alternative approach to ipv4-only relays. It splits up `mail_domain` in 3 different possible config values:

- `mail_domain`: can be the domain of a relay, or `13.12.23.42`: this is used for routing
- `mail_domain_deliverable`: can be the domain of a relay, or `[13.12.23.42]`: this is used in the email addresses
- `mail_domain_hostname`: can be the domain of a relay, or `42.23.12.13.in-addr.arpa`: this is used as `myhostname` in postfix

todo:

- [x] fix tests
- [x] look up every usage of `mail_domain` and see whether it's correctly used
- [x] let cmdeploy dns fail gracefully
- [x] make cmdeploy status work
- [x] document experimental IPv4-only setups
  - [x] explain that strict TLS & DKIM checking is disabled for them
- [x] make CI test IPv4-only setups with https://github.com/chatmail/cmlxc/pull/4/

PR #956, opened 2026-05-07
better naming for ip4-only PR #919

hpk42:
retain "config.mail_domain" as the domain part of @ email addresses which means for ipv4 relays  "[1.2.3.4]"

and introduce config.ipv4_relay, config.postfix_myhostname and config.mail_domain_bare helpers. 
i found the multiple "mail_domain*" names confusing and think the new names make several places read better.

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/master.cf.j2
+++ b/cmdeploy/src/cmdeploy/postfix/master.cf.j2
@@ -81,7 +81,7 @@ filter    unix -        n       n       -       -       lmtp
   -o syslog_name=postfix/reinject
   -o milter_macro_daemon_name=ORIGINATING
   -o cleanup_service_name=authclean
-{% if config.mail_domain == config.mail_domain_deliverable %}  -o smtpd_milters=unix:opendkim/opendkim.sock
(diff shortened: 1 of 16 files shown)


j4n (approved):
Indeed much more readable. Tested deploying v4-only and with domain.

cmdeploy/src/cmdeploy/postfix/master.cf.j2
```diff
@@ -81,7 +81,7 @@ filter    unix -        n       n       -       -       lmtp
   -o syslog_name=postfix/reinject
   -o milter_macro_daemon_name=ORIGINATING
   -o cleanup_service_name=authclean
-{% if config.mail_domain == config.mail_domain_deliverable %}  -o smtpd_milters=unix:opendkim/opendkim.sock
+{% if not config.ipv4_relay %}  -o smtpd_milters=unix:opendkim/opendkim.sock
```
missytake, line 84: yeah, that's much prettier.

missytake (approved):
Yeah, I guess we can also do it like this.

PR #959, opened 2026-05-08
fix: avoid lost notifications by properly migrating legacy data structure for tokens

hpk42:
metadata storage used list type, but if no new setmetadata happened, the user would not be notified even if device tokens were present. 

FWIW i checked with nine  -- 434 users  have token lists instead of dicts, leading to them not getting notified on incoming messages.  The issue was introduced June 2025 in #583  -- while it by now only affects a fraction of people, we should still fix it.

### DIFF
--- a/chatmaild/src/chatmaild/tests/test_metadata.py
+++ b/chatmaild/src/chatmaild/tests/test_metadata.py
@@ -372,3 +372,14 @@ def test_iroh_relay(dictproxy):
     dictproxy.iroh_relay = "https://example.org/"
     dictproxy.loop_forever(rfile, wfile)
     assert wfile.getvalue() == b"Ohttps://example.org/\n"
+
+
+def test_legacy_token_migration(metadata, testaddr):
+    with metadata.get_metadata_dict(testaddr).modify() as data:
+        data[metadata.DEVICETOKEN_KEY] = ["oldtoken1", "oldtoken2"]
+
+    assert metadata.get_tokens_for_addr(testaddr) == ["oldtoken1", "oldtoken2"]
+    mdict = metadata.get_metadata_dict(testaddr).read()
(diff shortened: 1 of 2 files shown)


chatmaild/src/chatmaild/tests/test_metadata.py
```diff
@@ -372,3 +372,14 @@ def test_iroh_relay(dictproxy):
...
+
+def test_legacy_token_migration(metadata, testaddr):
+    with metadata.get_metadata_dict(testaddr).modify() as data:
+        data[metadata.DEVICETOKEN_KEY] = ["oldtoken1", "oldtoken2"]
+
+    assert metadata.get_tokens_for_addr(testaddr) == ["oldtoken1", "oldtoken2"]
+    mdict = metadata.get_metadata_dict(testaddr).read()
+    tokens = mdict[metadata.DEVICETOKEN_KEY]
+    assert isinstance(tokens, dict)
+    assert "oldtoken1" in tokens
```
missytake: don't we need `oldtoken2` as well? shouldn't the test check for that either?

missytake (approved):

chatmaild/src/chatmaild/tests/test_metadata.py
hpk42: k, added it for completeness.

link2xt:
It is broken since https://github.com/chatmail/relay/pull/583, but back then core was storing unencrypted tokens on each reconnect, so this was supposed to trigger modification. Legacy format that is still not upgraded means the user never stored the token afterwards, so they did not use the old version that stored the token each time and also did not upgrade to the version that encrypts the token. Maybe they are just desktop users or bots that had token once (because profile was shared with admin), but then dropped the device with a token. Probably the tokens stored in this format are expired already.

link2xt (approved):

PR #960, opened 2026-05-09
feat: expose metadata "maxsmtprecipients" value

hpk42:
also add metadata tests and make metadata lookup method more readable by using structural match/case syntax


j4n (approved):
looks good to me. new key worked on test deployment.

PR #961, opened 2026-05-10
fix(nginx): align port 80 redirect logging

punkero-org:
The port 80 redirect server block has no access_log defined, so it is falling back to the default of a physical file. This is not aligned with the rest of the configured logging to systemd.

Implement the expected logging path for the port 80 redirect server

### DIFF
--- a/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
+++ b/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
@@ -159,6 +159,7 @@ http {
 		{% endif %}
 
 		return 301 https://$host$request_uri;
+		access_log syslog:server=unix:/dev/log,facility=local7;
 	}
 
 	{% if config.tls_cert_mode == "acme" %}


punkero-org:
**Test**

Before change: (random visitor)
```
# tail -1 /var/log/nginx/access.log

IPV4REDACTED - - [10/May/2026:11:32:33 +0000] "GET / HTTP/1.1" 301 162 "-" "Mozilla/5.0 zgrab/0.x"
```

After change: (me with curl)
```
# journalctl -u nginx.service | grep curl | tail -1

May 10 11:34:34 REDACTED nginx[19107]: REDACTED nginx: IPV6REDACTED - - [10/May/2026:11:34:34 +0000] "GET / HTTP/1.1" 301 162 "-" "curl/8.20.0"
```

hpk42:
thanks for the fix.  took the liberty to take your fix and generalize it so server sections will always use http context access_log setting.  #962

PR #962, opened 2026-05-10
fix(logging): log all http requests to syslog

hpk42:
replaces #961 and adds a test that nobody adds unnecessary access_log lines.

### DIFF
--- a/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
+++ b/cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
@@ -42,6 +42,9 @@ stream {
 }
 
 http {
+	# access_log setting is inherited by all server sections
+	access_log syslog:server=unix:/dev/log,facility=local7;
+
 {% if config.tls_cert_mode == "self" %}
 	limit_req_zone $binary_remote_addr zone=newaccount:10m rate=2r/s;
 {% endif %}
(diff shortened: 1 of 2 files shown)


cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
```diff
@@ -42,6 +42,9 @@ stream {
 }
 
 http {
+    # access_log setting is inherited by all server sections
+    access_log syslog:server=unix:/dev/log,facility=local7;
```
link2xt: This uses spaces, but other lines below that i checked use tabs. Will likely be not aligned with 8-spaces tabs. Debian uses tabs in default configs.

link2xt (approved):

cmdeploy/src/cmdeploy/nginx/nginx.conf.j2
hpk42: fixed

punkero-org: In case this is helpful, I use this in my .vimrc:

```
" highlight extra/unwanted whitespace
autocmd ColorScheme * highlight ExtraWhitespace ctermbg=red guibg=red
highlight ExtraWhitespace ctermbg=red guibg=red guifg=red
match ExtraWhitespace /\s\+$\| \+\ze\t/
```

The template is full of extra spaces, most of it "blank" lines (a line of nothing more than space indents).  A quick check of a stock nginx.conf shows they're not in there, so added by accident along the way here. $0.02 hth

PR #963, opened 2026-05-10
feat: support setup without domain, with only an IPv4 address

hpk42:
this is #919 but rebased on current main and with adapated CI

### DIFF
--- a/.github/workflows/ci-no-dns.yaml
+++ b/.github/workflows/ci-no-dns.yaml
@@ -9,6 +9,8 @@ on:
   pull_request:
     branches: [ "main" ]
 
+permissions: {}
+
 # Newest push wins: Prevents multiple runs from clashing and wasting runner efforts
 concurrency:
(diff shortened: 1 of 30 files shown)


j4n (approved):
looks good to me, I guess regular <-> ipv4 interoperability test will be deferred to the nightly?

missytake (approved):
Okay, I guess we all agree on a naming scheme for the possible domain variables now :)

.github/workflows/ci-no-dns.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+on:
+  # Triggers when a PR is merged into main or a direct push occurs
+  push:
+    branches: [ "main" ]
+    
+  # Triggers for any PR (and its subsequent commits) targeting the main branch
+  pull_request:
+    branches: [ "main" ]
+
```
missytake, line 11:
```suggestion

permissions: {}

```

This would avoid a zizmor warning :)

hpk42: done

PR #874, opened 2026-02-26
feat: Sign bounce messages

j-g00da:
Closes #873

PR #964, opened 2026-05-11
feat: DKIM-sign bounce messages (mainly "user does not exist")

hpk42:
closes #873 This is based on, and replaces Jagoda's https://github.com/chatmail/relay/pull/874 but comes with a simpler, more robust test. Also takes care to not do dkim-signing for IP-relays.

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/main.cf.j2
+++ b/cmdeploy/src/cmdeploy/postfix/main.cf.j2
@@ -103,3 +103,11 @@ smtpd_peername_lookup = no
 default_transport = lmtp-filtermail:inet:[127.0.0.1]:{{ config.filtermail_lmtp_port_transport }}
 lmtp-filtermail_initial_destination_concurrency=10000
 lmtp-filtermail_destination_concurrency_limit=10000
+
+{% if not config.ipv4_relay %}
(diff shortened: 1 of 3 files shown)


cmdeploy/src/cmdeploy/postfix/master.cf.j2
```diff
@@ -51,12 +51,16 @@ smtps     inet  n       -       y       -       5000    smtpd
   -o smtpd_proxy_filter=127.0.0.1:{{ config.filtermail_smtp_port }}
 #628       inet  n       -       y       -       -       qmqpd
 pickup    unix  n       -       y       60      1       pickup
-cleanup   unix  n       -       y       -       0       cleanup
+{% if not config.ipv4_relay %}  -o cleanup_service_name=signlocal
+{% endif %}cleanup   unix  n       -       y       -       0       cleanup
```
j-g00da: Pretty sure this is redundant, IIRC I first tried to hook the cleanup to pickup as every bounce goes through it, but pickup doesn't honor that option and it has to be set in bounce instead.

j-g00da (approved):
Looks good, let's first merge filtermail so the CI passes.

j-g00da:
filtermail 0.6.5 -> https://github.com/chatmail/relay/pull/966

PR #969, opened 2026-05-13
don't merge: try out lmtp no subject

hpk42:
problem is that we still have some miniscule number of cleartext receivers on some relays.


hpk42:
k, so core does not depend on Subject or Date headers but for unencrypted received messages it might still play a historic role.

PR #970, opened 2026-05-14
feat!: remove passthrough_sender and passthrough_recipients  and warn about unused ini settings

hpk42:
this eliminates one more source of unencrypted messages that is hardly used practically. 

When running "cmdeploy run" operators will see a warning if their chatmail.ini contains these unused options.

```
...
Deploy completed, call `cmdeploy dns` next.
WARNING: chatmail.ini contains settings that have no effect: passthrough_senders, passthrough_recipients
Please remove them from chatmail.ini.
```

### DIFF
--- a/chatmaild/src/chatmaild/config.py
+++ b/chatmaild/src/chatmaild/config.py
@@ -41,8 +41,8 @@ class Config:
         self.username_min_length = int(params.get("username_min_length", 9))
         self.username_max_length = int(params.get("username_max_length", 9))
         self.password_min_length = int(params.get("password_min_length", 9))
-        self.passthrough_senders = params.get("passthrough_senders", "").split()
-        self.passthrough_recipients = params.get("passthrough_recipients", "").split()
+        _unused = ("passthrough_senders", "passthrough_recipients")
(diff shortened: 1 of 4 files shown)


chatmaild/src/chatmaild/config.py
```diff
@@ -41,8 +41,8 @@ def __init__(self, inipath, params):
         self.username_min_length = int(params.get("username_min_length", 9))
         self.username_max_length = int(params.get("username_max_length", 9))
         self.password_min_length = int(params.get("password_min_length", 9))
-        self.passthrough_senders = params.get("passthrough_senders", "").split()
-        self.passthrough_recipients = params.get("passthrough_recipients", "").split()
+        _unused = ("passthrough_senders", "passthrough_recipients")
```
link2xt: Maybe use `params.pop()` instead of `params.get()` above and then warn about all keys that were not used instead of keeping a list of deprecated keys here?

hpk42: k, that's a bit more of a refactor and general mechanism then. i added a commit to go for your suggestion.

link2xt (approved):

PR #971, opened 2026-05-14
fix: fix #972 by increasing file descriptors for filtermail, and reduce destination concurrency

### LINKED ISSUE #972: filtermail opens too many files
A relay operator had a broken relay because the filtermail-transport service kept crashing. They had tens of thousands of mails in their mailq and systemd limits open files by default to 1024. I did an override for their service file to increase the open files limit to 500,000. This fixed the crash and mails started delivering again.

attached is what I caught with strace

[filtermail.txt](https://github.com/user-attachments/files/27772216/filtermail.txt)

hpk42:


link2xt:
This references wrong issue, the issue is in another repo: https://github.com/chatmail/relay/issues/972

cmdeploy/src/cmdeploy/postfix/main.cf.j2
```diff
@@ -101,8 +101,8 @@ smtpd_peername_lookup = no
 # so instead this is handled in filtermail.
 # We use LMTP instead SMTP so we can communicate per-recipient errors back to postfix.
 default_transport = lmtp-filtermail:inet:[127.0.0.1]:{{ config.filtermail_lmtp_port_transport }}
-lmtp-filtermail_initial_destination_concurrency=10000
-lmtp-filtermail_destination_concurrency_limit=10000
+lmtp-filtermail_initial_destination_concurrency=500
```
link2xt: I would not change this, this effectively reverts https://github.com/chatmail/relay/commit/b05e26819fbdc25102a6804b91c082e9163e877d and will result in head of line blocking that the change was fixing.

cmdeploy/src/cmdeploy/filtermail/filtermail-transport.service.j2
```diff
@@ -6,6 +6,7 @@ ExecStart={{ bin_path }} {{ config_path }} transport
 Restart=always
 RestartSec=30
 User=vmail
+LimitNOFILE=524288
```
ccclxxiii, line 9: @hpk42 consider parameterizing this value if we continue this path via `chatmail.ini`, otherwise this should work.

ccclxxiii (approved):

cmdeploy/src/cmdeploy/postfix/main.cf.j2
hpk42: ok, i reverted. i guess the file descriptor increase needs to suffice then.

cmdeploy/src/cmdeploy/filtermail/filtermail-transport.service.j2
hpk42: it's all going to change eventually when https://github.com/chatmail/filtermail/issues/141 gets implemented, so wouldn't expose anything as operator config right now.

link2xt (approved):
Commit message is now wrong

link2xt:
Also related fix for filtermail issue, but this will need a new filtermail release: https://github.com/chatmail/filtermail/pull/159

PR #975, opened 2026-05-16
docs: add scripts/initenv.sh to upgrade instructions

missytake:
Someone was missing a dependency which was introduced by `git pull`, this makes sure that all dependencies are there as well.

### DIFF
--- a/doc/source/faq.rst
+++ b/doc/source/faq.rst
@@ -60,6 +60,7 @@ and run the following commands:
    ::
 
        git pull origin main --rebase --autostash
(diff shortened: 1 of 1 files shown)


punkero-org:
I just tried to upgrade and ran into this as well, thanks. In my case I prefer to use a detached head tag method, like so:

```
# I'm on detached head 1.10.0
git checkout main
git pull
git checkout tags/1.11.0
```

I confirm this PR fixes my method as well, it added dnspython and domain_validator modules.

PR #977, opened 2026-05-16
fix: set relay restrictions per smtpd service with default reject

link2xt:
We never want to defer email with a tepporary error when it has destination that we cannot deliver locally and don't want to relay. To avoid doing this accidentally, set default action to "reject" and then override it with the minimal restrictions per smtpd.

Submission ports already had smtpd_relay_restrictions=permit_sasl_authenticated,reject override.

Each smtpd port must have at least one of
reject, reject_unauth_destination, defer, defer_if_permit, defer_unauth_destination according to <https://www.postfix.org/postconf.5.html#smtpd_relay_restrictions>.

I have set smtpd_relay_restrictions=reject_unauth_destination for port 25 and incoming reinject port, and smtpd_relay_restrictions=permit_mynetworks,reject for outgoing reinject port.


j-g00da (approved):

j4n (approved):
looks good to me, maybe it would be helpful to already phrase out a changelog entry that informs relay operators of what this implies and why we're changing it. tested deploy and rejections.

PR #978, opened 2026-05-19
ci: auto-trigger docker build on release tag push

j4n:
docker-dispatch.yaml previously only fired on push to main and manual workflow_dispatch, so tagging 1.11.0 did not build the release image. This change adds matching of X.Y.Z tag.


missytake (approved):
Nice, this rests on https://github.com/chatmail/docker/commit/613d3218242c97d6fe689c89999538ca58756b7c using the tag as image label, right?

PR #979, opened 2026-05-19
fix: limit the number of LMTP clients for filtermail-transport to 1000

link2xt:
Postfix does not have jitter for deferred mails
and scans the queue periodically every
queue_run_delay (<https://www.postfix.org/postconf.5.html#queue_run_delay>). As a result it is likely
to try delivering many deferred messages
at the same time.

Normally the number of outgoing connections
should be low even with unreachable destinations,
but after the server downtime
or if admin flushes the queue manually
it is possible that a lot of messages
to the same unreachable destination
expire at once and are moved
from "deferred" into the "active" queue.

Trying to deliver them all at once
may make the server run out of memory
by starting many LMTP clients.
Limiting the number of LMTP processes
turns OOM problem into head of line blocking problem. Messages sent to reachable destinations
will be delayed as well,
but at least deferred messages will
get distributed over time.

In this case "active" queue may grow
(up to qmgr_message_active_limit defaulting to 20000), but then admin may notice the problem
and solve it e.g. by making the destinations reachable or setting up a transport map to route
messages for known dead servers into discard transport.

Eventually the problem should be solved
by filtermail-transport quickly returning temporary errors for destinations which already have many messages queued, then we can reduce "maxproc" further.

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/main.cf.j2
+++ b/cmdeploy/src/cmdeploy/postfix/main.cf.j2
@@ -102,6 +102,18 @@ smtpd_peername_lookup = no
 # so instead this is handled in filtermail.
 # We use LMTP instead SMTP so we can communicate per-recipient errors back to postfix.
 default_transport = lmtp-filtermail:inet:[127.0.0.1]:{{ config.filtermail_lmtp_port_transport }}
+
+# All deliveries over lmtp-filtermail are treated
+# as having the same destination [127.0.0.1],
+# so it is not possible to limit per-destination concurrency here,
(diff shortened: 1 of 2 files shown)


link2xt:
Related, not merged proposal to reduce destination concurrency: https://github.com/chatmail/relay/pull/971#discussion_r3243711156

Reducing "maxproc" to 500 or even 100 would be nice, but once filtermail rejects mails to broken destinations (https://github.com/chatmail/filtermail/issues/141) otherwise we will be delaying messages on relays that have memory.

missytake (approved):
Should help for now :)

PR #980, opened 2026-05-19
fix: reduce maxproc for filtermail-transport LMTP client to 500

link2xt:
This further reduces it from 1000.
For small servers this may be needed if they have low memory. For large servers may be increased manually for now.


link2xt:
Follow up to #979

hpk42 (approved):
Lgtm

PR #981, opened 2026-05-20
feat: reduce maximal_queue_lifetime from 5d to 2d

link2xt:
If the message is not delivered within 2 days,
it is unlikely to be delivered in 5 days either.


hpk42 (approved):
Makes sense. the SMTP defaults from decades ago do not make sense for instant messaging.  With multi relays becoming more common id even consider shorter time frames thab 2d.

PR #984, opened 2026-05-28
test: avoid race in sender ip cleanup imap check

ccclxxiii:
### pr overview:

remove the `wait_for_incoming_msg()` call from `test_hide_senders_ip_address` and poll imap directly before delta chat consumes or moves the message.

this may fix the ci failure where:

```text
assert msgs, "expected at least one message"
```

fails after the message has already been processed by the client.

#### changes

* open imap connection before sending the message
* remove `user2.wait_for_incoming_msg()`
* poll `mailbox.fetch(mark_seen=False)` with a short timeout
* preserve the existing sender ip cleanup assertion

### notes

this is intended as a ci stabilization change for integration testing and should not affect runtime behavior.


ccclxxiii:
closing while we wait on https://github.com/chatmail/cmlxc/pull/19

PR #986, opened 2026-05-28
fix: pin chatmail core to 2.49.0

link2xt:
2.50.0 has delete_server_after config removed,
but it is still used in the tests.


link2xt:
Why is this PR closed?

PR #990, opened 2026-05-28
fix: core 2.50.0 does not have delete_server_after config anymore.

hpk42:


link2xt:
No-DNS also fails because it has pinned cmlxc with unpinned core and `delete_server_after`, but otherwise looks good: 
https://github.com/chatmail/relay/blob/4b04aae83b5545e551aa80c62d1e4399706271dd/.github/workflows/ci-no-dns.yaml#L23

link2xt (approved):

missytake (approved):
CI is green now, great :)

PR #991, opened 2026-05-28
test: set socket security for IMAP and SMTP to "TLS" in "dclogin"

link2xt:
With "default" (like it was for SMTP) or not set (like it was for IMAP), both TLS and STARTTLS are tried.
Trying STARTTLS against TLS port is going to timeout because in STARTTLS server talks first,
but when connected to TLS port the server
waits for TLS client hello and does not send anything.

Should not actually matter in tests which connect successfully on the first try because implicit TLS is tried first.


link2xt:
Related documentation PR: https://github.com/deltachat/interface/pull/92

missytake (approved):

PR #993, opened 2026-06-02
fix: Always deploy unbound.conf.d/chatmail.conf

j-g00da:
This fixes issue with negative cache
being only disabled in ipv4-only mode.

Follow up to #992


missytake (approved):
Great, makes the code easier to understand as well.

j4n (commented):
looks good to me! havent tested.

PR #994, opened 2026-06-03
fix: expire empty directories

### LINKED ISSUE #759: Expire should delete account folders with no `password`
Currently accounts with no last_login date (because no `password` was found) are treated as recently logged in and never get deleted:
https://github.com/chatmail/relay/blob/b000213c68bb91e80a217ea96c33033a373638d4/chatmaild/src/chatmaild/expire.py#L129

missytake:
fix #759

### DIFF
--- a/chatmaild/src/chatmaild/expire.py
+++ b/chatmaild/src/chatmaild/expire.py
@@ -168,6 +168,13 @@ class Expiry:
         if mbox.last_login and mbox.last_login < cutoff_without_login:
             self.remove_mailbox(mbox.basedir)
             return
+        elif mbox.last_login is None:
+            try:
+                os.rmdir(mbox.basedir)
+            except OSError:
+                print_info(
+                    f"Skipped deleting {mbox.basedir}, doesn't have last_login but isn't empty"
+                )
 
         mboxname = os.path.basename(mbox.basedir)
         if self.verbose:
--- a/chatmaild/src/chatmaild/expire.py
+++ b/chatmaild/src/chatmaild/expire.py
@@ -170,7 +170,8 @@ class Expiry:
             return
         elif mbox.last_login is None:
             try:
-                os.rmdir(mbox.basedir)
+                if not self.dry:
+                    os.rmdir(mbox.basedir)
             except OSError:
                 print_info(
                     f"Skipped deleting {mbox.basedir}, doesn't have last_login but isn't empty"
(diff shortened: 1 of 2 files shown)


chatmaild/src/chatmaild/expire.py
```diff
@@ -168,6 +168,13 @@ def process_mailbox_stat(self, mbox):
         if mbox.last_login and mbox.last_login < cutoff_without_login:
             self.remove_mailbox(mbox.basedir)
             return
+        elif mbox.last_login is None:
+            try:
+                os.rmdir(mbox.basedir)
+            except OSError:
+                print_info(
+                    f"Skipped deleting {mbox.basedir}, doesn't have last_login but isn't empty"
+                )
```
j4n, line 179:
```suggestion
        elif mbox.last_login is None:
            if not self.dry:
                try:
                    os.rmdir(mbox.basedir)
                    self.del_mboxes += 1
                    return
                except OSError:
                    print_info(
                        f"Skipped deleting {mbox.basedir}, doesn't have last_login but isn't empty"
                    )
            else:
                return
```

chatmaild/src/chatmaild/tests/test_expire.py
```diff
@@ -104,6 +106,33 @@ def test_stats_mailbox(mbox1):
...
+
+    for entry in os.scandir(mbox_rescan.basedir):
+        if os.path.isdir(entry):
+            shutil.rmtree(entry)
+        else:
+            os.remove(entry)
+
+    exp.process_mailbox_stat(mbox_rescan)
+    out, err = capsys.readouterr()
+    assert not "doesn't have last_login but isn't empty" in err
```
j4n:
```suggestion
    assert "doesn't have last_login but isn't empty" not in err
```

chatmaild/src/chatmaild/tests/test_expire.py
```diff
@@ -104,6 +106,33 @@ def test_stats_mailbox(mbox1):
     assert mbox3.last_login is None
 
 
+def test_mbox_without_password(mbox1, example_config, capsys):
+    password = Path(mbox1.basedir).joinpath("password")
+    os.remove(password)
+    mbox_rescan = MailboxStat(mbox1.basedir)
+    mbox_rescan.scandir(mbox_rescan.basedir)
```
j4n: any reason to rescan? init calls scandir()

j4n (approved):
looks good to me, not tested yet.

chatmaild/src/chatmaild/tests/test_expire.py
missytake: Ah, right. adding lines until it works is almost as good as removing them afterwards.

PR #997, opened 2026-06-03
fix: still overwrite /etc/resolv.conf if it is a symbolic link

### LINKED ISSUE #515: cmdeploy run issues [Contabo]

missytake:
fix #515

### DIFF
--- a/cmdeploy/src/cmdeploy/basedeploy.py
+++ b/cmdeploy/src/cmdeploy/basedeploy.py
@@ -166,7 +166,7 @@ class Deployer:
             return self.put_template(src, dest, **kwargs)
         return self.put_file(src, dest)
 
-    def put_file(self, src, dest, mode="644"):
+    def put_file(self, src, dest, mode="644", **kwargs):
         if isinstance(src, str):
             src = get_resource(src)
         res = files.put(
@@ -176,6 +176,7 @@ class Deployer:
             user="root",
             group="root",
             mode=mode,
+            **kwargs,
         )
 
         return self._update_restart_signals(dest, res)
--- a/cmdeploy/src/cmdeploy/deployers.py
+++ b/cmdeploy/src/cmdeploy/deployers.py
@@ -164,6 +164,7 @@ class UnboundDeployer(Deployer):
         self.put_file(
             src=BytesIO(b"nameserver 127.0.0.1\nnameserver 9.9.9.9\n"),
             dest="/etc/resolv.conf",
+            force=True,
         )
         server.shell(
             name="Generate root keys for validating DNSSEC",


cmdeploy/src/cmdeploy/basedeploy.py
```diff
@@ -166,7 +166,7 @@ def ensure_systemd_unit(self, src, **kwargs):
             return self.put_template(src, dest, **kwargs)
         return self.put_file(src, dest)
```
j4n, line 168:
```suggestion

        # Ensure /etc/resolv.conf is a real file, not a symlink
        server.shell(
            name="Remove resolv.conf symlink if present",
            commands=["[ -L /etc/resolv.conf ] && rm -f /etc/resolv.conf || true"],
        )
```

j4n (changes requested):
I think we need to remove the link if it is a link; otherwise we're writing to the linked file which might be overwritten again by resolvconf and the likes:
```
# for i in /etc/resolv.*; do echo "== $i =="; cat $i; done
== /etc/resolv.conf ==
nameserver 127.0.0.1
nameserver 9.9.9.9
== /etc/resolv.conf.bad ==
nameserver 1.1.1.1
nameserver 8.8.8.8
== /etc/resolv.conf.good ==
nameserver 127.0.0.1
nameserver 9.9.9.9
# cmdeploy ...
for i in /etc/resolv.*; do echo "== $i =="; cat $i; done
== /etc/resolv.conf ==
nameserver 127.0.0.1
nameserver 9.9.9.9
== /etc/resolv.conf.bad ==
nameserver 127.0.0.1
nameserver 9.9.9.9
== /etc/resolv.conf.good ==
nameserver 127.0.0.1
nameserver 9.9.9.9

# ls -l /etc/resolv.conf
lrwxrwxrwx 1 root root 20 Jun  9 15:57 /etc/resolv.conf -> /etc/resolv.conf.bad

```

we might also just get rid of the force as its not really helpful probably.

j4n (approved):
nvm,  this is working as intended.

PR #1001, opened 2026-06-05
WIP: chore(update): use dovecot 2.3 built for Debian 13

missytake:
using dovecot build from https://github.com/chatmail/dovecot/pull/8

still needs cmlxc to use debian 13, then we can sort out the errors.


missytake:
Superseded by #1002.

PR #1003, opened 2026-06-05
fix: crypt-r dependency was declared for wrong Python version

feld:
The original crypt library was last supported in Python 3.12, so it's not needed until Python 3.13 is default (Trixie)

> The crypt_r module is a renamed copy of the crypt module as it was present in Python 3.12 before it was removed.


hpk42 (approved):
LGTM. please rebase and merge.

PR #1005, opened 2026-06-05
Restore Aggressive LMTP header cleanup

feld:
Originally #816 

Reverted with #1004

Filtermail seems to fail DKIM validation for some messages with this deployed and needs further investigation

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
+++ b/cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
@@ -1,3 +1,23 @@
-/^DKIM-Signature:/            IGNORE
-/^Authentication-Results:/            IGNORE
-/^Received:/            IGNORE
+# List of headers for incoming messages 
+# that must be retained for functionality and compatibility reasons 
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-/            DUNNO
+/^Content-Type:/    DUNNO
+
+# For receiving clear-text messages (still supported in May 2026)
+/^Subject:/         DUNNO
+/^Date:/            DUNNO
+/^To:/              DUNNO
+/^CC:/              DUNNO
+/^References:/      DUNNO
+/^In-Reply-To:/     DUNNO
+
+# Senders might support Autocrypt 1 but not RFC9788 (Header Protection) 
+/^Autocrypt:/       DUNNO
+
+# SecureJoin V2 protocol headers (for backward compatibility) 
+/^Secure-Join/      DUNNO
+
+# Ignore all other headers
+/.*/                IGNORE


feld:
Very interested in seeing if #1016 is the root cause of the problem causing DKIM issues. Perhaps it had nothing to do with this change?

cmdeploy/src/cmdeploy/postfix/lmtp_header_cleanup
```diff
@@ -1,3 +1,23 @@
-/^DKIM-Signature:/            IGNORE
-/^Authentication-Results:/            IGNORE
-/^Received:/            IGNORE
+# List of headers for incoming messages 
+# that must be retained for functionality and compatibility reasons 
+/^From:/            DUNNO
+/^Message-Id:/      DUNNO
+/^Chat-/            DUNNO
+/^Content-Type:/    DUNNO
```
link2xt, line 6: For safety i'd list all ["Structural headers"](https://www.rfc-editor.org/rfc/rfc9787.html#name-structural-header-fields):
- MIME-Version
- Content-Type
- Content-Transfer-Encoding
- Content-Disposition

Removing `Content-Transfer-Encoding` from multipart MIME message sounds dangerous regardless of the signatures on the message.

link2xt:
> Very interested in seeing if #1016 is the root cause of the problem causing DKIM issues. Perhaps it had nothing to do with this change?

As far as I understand without #1016 the change in this PR also applied filters to the body of the message and could have broken DKIM e.g. by removing headers like `Content-Description: PGP/MIME version identification` from the inner parts of OpenPGP messages.

Don't know what was broken last time when https://github.com/chatmail/relay/pull/1004 was reverted, but worth rebasing and trying again on some test server and check that it still federates properly.

For outgoing messages filtermail-transport is now the process signing the messages and sending them, so there should be no way this breaks the signatures on outgoing messages as postfix never sees outgoing signed messages.

I still don't know why `lmtp_header_checks` apply to incoming messages at all, maybe it was when OpenDKIM was still there checking the signatures. It seems that the filters should only apply to this service which is only used for the final local delivery to Dovecot, after all DKIM checks done by filtermail-incoming:
https://github.com/chatmail/relay/blob/deca1ef2a3eec6aebca3bb5b34e5ebee209567ec/cmdeploy/src/cmdeploy/postfix/master.cf.j2#L75
We can also move the configuration from main.cf to master.cf to make it clear.

The other LMTP service which does not overwrite `lmtp_header_checks` set in main.cf seems to be unused: https://github.com/chatmail/relay/pull/1022

link2xt:
Somewhat related, I also recently discovered that postfix by default rewrites addresses in the `To` field that look like local e.g. "foobar" with "foobar@example.org": https://www.postfix.org/ADDRESS_REWRITING_README.html

Except for maybe bounce messages (don't know if they are generated with a local address or with the domain from the start) it looks like we want to disable this address rewriting logic as much as possible.

Maybe removing From and To fields completely can cause some problems with some internal logic of Postfix, they are not treated as just opaque values.

Issue #1012, opened 2026-06-25
please make a command to remove all parts you deployed


grince:
Hi,
part of a lifecycle of an application is building and running it - but also changing, decommissioning and removing. this is easy with docker containers (just throw them away) but not so easy with vservers or bare metal systems. since you have a quite comfortable script to deploy the application on a server, please add a command to remove all content and packages (and services) you installed by your script. maybe with a question "do you really want to delete all content and services?", but make it possible to purge the chatmail/relay system from a server.

thanks in advance,
Rince

ccclxxiii:
@grince thanks for opening issue. working on a branch for this now.

hpk42:
why not wipe the disk and put a fresh OS on it?

ccclxxiii:
> why not wipe the disk and put a fresh OS on it?

why not use the `remove` command? some people have single-server for many purposes.  this is a frequent pattern we see.

grince:
if you expect everyone to have a server single-used, then Container are the way to go (and then you can use docker, lxc or other). but this Is not the purpose of cmdeploy, it wants to use an existing system and deploy everything _on top_. but at no point I do see what the changes really are (other than reading the sourcecode). you can have eg. a system acting as a Webserver on another IP or having another service running on that system. since I am toying a little bit with this chat relay (including changing. the relay-domain) I wanted to see wether I can remove everything and start from scratch - but that is not (yet) possible.

j4n:
Currently, we only officially support running a dedicated system; you may use the [docker images](https://github.com/chatmail/docker) if you want to play around with the relay. 

For sake of overview, at present (may change any day), a manual remove would need to:

## Uninstall apt-managed packages

`apt purge unbound* dnsutils python3-virtualenv gcc python3-dev postfix dovecot* opendkim* nginx* certbot, python3-certbot-nginx`

## Manually remove:

### chatmaild and other binaries
   - `/usr/local/lib/chatmaild/`
   - `/usr/local/bin/chatmail-turn`
   - `/usr/local/bin/iroh-relay`
   - `/usr/lib/cgi-bin/newemail.py`

### Config Files
   - `/etc/postfix/`
   - `/etc/dovecot/`
   - `/etc/unbound/`
   - `/etc/iroh-relay.toml`
   - `/etc/nginx/`
   - `/etc/chatmail-version`
   - `/etc/apt/apt.conf.d/00InstallRecommends`
   - `/etc/apt/preferences.d/pin-dovecot`
[...]

PR #1013, opened 2026-06-25
add cmdeploy remove command

ccclxxiii:
pr overview:

- related to #1012 
- adds `cmdeploy` remove so operators can tear down relay from server before redeploying or decommissioning it.
- covers services, systemd units, relay config/state, mailboxes, web files, local binaries, users/groups, packages, and legacy leftovers. includes `--dry-run`, `--yes`, and `--keep-packages`.
- also documents the removal flow and the CAA `accounturi` gotcha when redeploying after ACME state is removed.
- smoke-test against `darkrun.dev` live

### DIFF
--- a/cmdeploy/src/cmdeploy/cmdeploy.py
+++ b/cmdeploy/src/cmdeploy/cmdeploy.py
@@ -104,7 +104,9 @@ def run_cmd(args, out):
         args.dns_check_disabled = True
     if not args.dns_check_disabled:
         remote_data = dns.get_initial_remote_data(sshexec, args.config.mail_domain)
-        if not dns.check_initial_remote_data(remote_data, strict_tls=strict_tls, print=out.red):
+        if not dns.check_initial_remote_data(
+            remote_data, strict_tls=strict_tls, print=out.red
+        ):
             return 1
 
     env = os.environ.copy()
@@ -127,7 +129,11 @@ def run_cmd(args, out):
         out.check_call(cmd, env=env)
         if args.website_only:
             out.green("Website deployment completed.")
-        elif not args.dns_check_disabled and strict_tls and not remote_data["acme_account_url"]:
+        elif (
+            not args.dns_check_disabled
+            and strict_tls
+            and not remote_data["acme_account_url"]
+        ):
             out.red("Deploy completed but letsencrypt not configured")
             out.red("Run 'cmdeploy run' again")
         elif args.config.ipv4_relay:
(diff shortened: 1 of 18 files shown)


cmdeploy/src/cmdeploy/cmdeploy.py
```diff
@@ -141,6 +147,88 @@ def run_cmd(args, out):
...
+        "WARNING: this removes chatmail services, configuration, package state, "
+        f"and mailbox data for {domain!r}."
+    )
+    if args.config.tls_cert_mode == "acme":
+        out.red(
+            "DNS records are not removed. If your DNS has a CAA record with "
+            "a Let's Encrypt accounturi, remove or relax it before deploying "
+            "this relay again."
+        )
+    out.red("Type the chatmail domain to continue.")
```
j-g00da, line 187: Was about to suggest this but it's already here ;p neat!

cmdeploy/src/cmdeploy/removers.py
```diff
@@ -0,0 +1,290 @@
...
+
+EMPTY_SYSTEMD_DIRS = [
+    "/etc/systemd/system/dovecot.service.d",
+    "/etc/systemd/system/opendkim.service.d",
+    "/etc/systemd/system/postfix@.service.d",
+]
+
+RELAY_USERS = ["vmail", "iroh", "opendkim"]
+RELAY_GROUPS = ["vmail", "iroh", "opendkim"]
+
```
j-g00da, line 147: Maybe we should keep paths/file names/package names in dicts in something like `constants.py`, then reference them everywhere instead of hardcoding strings all over the place. This way we can be sure that what is deployed matches what is removed (this can very easily diverge and be missed in a review).

ccclxxiii: @j-g00da that's a good call, thank you.  let's see if anyone else has input.  but a const would make sense so you change it in one place.  appreciate the feedback.

ccclxxiii: @j-g00da made another pass in [d7a16d2be1d99fa4569785a2780f4a4e8ae074c1](https://github.com/chatmail/relay/pull/1013/commits/d7a16d2be1d99fa4569785a2780f4a4e8ae074c1)

ccclxxiii:
closing [re:](https://github.com/chatmail/relay/issues/1012#issuecomment-4799993333)

PR #1015, opened 2026-06-29
ignore: testing CI

hpk42:


hpk42:
ok, new cmlxc fixes seem to fix the issue of incus-failures.

PR #1016, opened 2026-06-30
feat(postfix): disable processing of MIME headers

link2xt:
Default behavior is processing headers
that are actually part of the message body.
This is unlikely to be needed
as headers such as Received are not added into MIME parts, but may result in broken DKIM signatures
if the body is modified.


link2xt:
This setting looks like what we want for simplicity in any case, as parsing nested and MIME headers is unexpected (didn't know Postfix does this).

May help with https://github.com/chatmail/relay/pull/1005 and https://support.delta.chat/t/how-can-i-access-unencrypted-emails-on-the-nine-testrun-org-server/5474 as there is one less possibility of postfix modifying the message body and breaking DKIM signatures.

j-g00da (approved):
Oh wow, I wonder if this causes any DKIM issues in the wild

PR #1019, opened 2026-07-17
 protect small servers by rejecting new addresses under resource constraints / capping smtp/imap connection limits

hpk42:
Mainly, this PR rejects new address creation when system is under load, has not much RAM or disk storage left. 

It also makes postfix/dovecot max connection limit available in chatmail.ini and lowers the defaults which were geared for nine.testrun.org operations.

### DIFF
--- a/chatmaild/src/chatmaild/doveauth.py
+++ b/chatmaild/src/chatmaild/doveauth.py
@@ -14,6 +14,7 @@ except ImportError:
 from .config import Config, read_config
 from .dictproxy import DictProxy
 from .migrate_db import migrate_from_db_to_maildir
+from .syslimits import has_sufficient_resources
 
 NOCREATE_FILE = "/etc/chatmail-nocreate"
 VALID_LOCALPART_RE = re.compile(r"^[a-z0-9._-]+$")
@@ -146,6 +147,8 @@ class AuthDictProxy(DictProxy):
         if userdata:
             return userdata
         if not is_allowed_to_create(self.config, addr, cleartext_password):
+            return
+        if not has_sufficient_resources(self.config):
             return
 
         lock = filelock.FileLock(str(user.password_path) + ".lock", timeout=5)
--- a/chatmaild/src/chatmaild/ini/chatmail.ini.f
+++ b/chatmaild/src/chatmaild/ini/chatmail.ini.f
@@ -42,6 +42,33 @@ mail_domain = {mail_domain}
 # minimum length a password must have
 #password_min_length = 9
 
+#
+# System resource limits
+#
+
+# The following three limits refuse creation of new addresses
+# while existing addresses keep working.
+# Rejections are logged by the doveauth service.
+
+# Maximum 1-minute load average, as reported by "uptime";
+# it counts processes waiting for disk I/O as well as for CPU.
+#max_load_1m = 5
+
+# Minimum memory available without swapping.
+#min_available_memory = 200M
+
+# Minimum free disk space on the file system holding the mailboxes.
+#min_free_disk_space = 1G
+
+# Maximum number of concurrent IMAP connections
+# (the Dovecot imap process limit).
+#max_imap_connections = 10000
+
+# Maximum number of concurrent SMTP connections
+# on each of the submission and smtps ports (the Postfix process limit).
+# A single client IP may use up to a fifth of this.
+#max_smtp_connections = 1000
+
 # Use externally managed TLS certificates instead of built-in acmetool.
 # Paths refer to files on the deployment server (not the build machine).
 # Both files must already exist before running cmdeploy.
--- a/chatmaild/src/chatmaild/tests/test_doveauth.py
+++ b/chatmaild/src/chatmaild/tests/test_doveauth.py
@@ -202,3 +202,17 @@ def test_50_concurrent_lookups_different_accounts(gencreds, dictproxy):
         res = results.get()
         if res is not None:
             pytest.fail(f"concurrent lookup failed\n{res}")
+
+
+def test_insufficient_resources_block_creation_not_existing_logins(
+    dictproxy, gencreds, monkeypatch
+):
+    addr, password = gencreds()
+    assert dictproxy.lookup_passdb(addr, password)
+
+    monkeypatch.setattr(
+        chatmaild.doveauth, "has_sufficient_resources", lambda config: False
+    )
+    newaddr, newpassword = gencreds()
+    assert not dictproxy.lookup_passdb(newaddr, newpassword)
+    assert dictproxy.lookup_passdb(addr, password)
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 
 # Chatmail relays for end-to-end encrypted email
 
-Chatmail relay servers are interoperable Mail Transport Agents (MTAs) designed for: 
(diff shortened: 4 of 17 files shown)


missytake (approved):
Thanks for pushing this! Let's ask operators which limits they deem sufficient.

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -25,6 +26,46 @@ def encrypt_password(password: str):
     return "{SHA512-CRYPT}" + passhash
 
 
+def get_available_memory_mb():
+    with open("/proc/meminfo") as f:
+        for line in f:
+            if line.startswith("MemAvailable:"):
```
link2xt: Seems to be fine, but need to be careful that this is not some value that always gets to 0 eventually.

https://stackoverflow.com/questions/30869297/difference-between-memfree-and-memavailable and https://docs.kernel.org/filesystems/proc.html suggests that it's the correct value, using MemFree would be wrong

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -42,6 +42,23 @@
 # minimum length a password must have
 #password_min_length = 9
 
+# Reject new address creation when system load is too high.
+# Maximum 1-minute load average per CPU core.
+#max_load_per_cpu_1m = 3
```
link2xt: 3 is to make sure it never triggers? Anything above 1 is essentially disabling the check?

hpk42: load average on linux also captures IO pressure.  You might have 1% cpu usage but load average 5 because of slow network/hardware.

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -25,6 +26,46 @@ def encrypt_password(password: str):
...
+                return int(line.split()[1]) // 1024
+    return 0
+
+
+def has_sufficient_resources(config):
+    load_1m, _load_5m, _load_15m = os.getloadavg()
+    max_load = config.max_load_per_cpu_1m * (os.cpu_count() or 1)
```
link2xt: I checked "man uptime" now, it says that load average is simply a number of processes using CPU (same as loadavg in https://docs.kernel.org/filesystems/proc.html), waiting for CPU, or waiting for I/O. Then i think it should not be scaled per CPU, just compare the load average to the constant in the config. It is likely that I/O is the bottleneck in a real system. Then admin can also compare "uptime" or "top" or some other monitoring to the config value directly to figure out the correct value, without dividing it by the CPU number.

If someone has a 32 CPU system with a slow disk, it will not be able to handle 32 times more users than a 1 CPU system with the same disk.

chatmaild/src/chatmaild/ini/chatmail.ini.f
link2xt: I commented below at https://github.com/chatmail/relay/pull/1019/changes#r3622341264, i think it's better to put unscaled value here then, since I/O is likely to be a bottleneck and having more logic scaling the number back and forth is more complex but not necessarily helping. Unscaled 3 seems ok for the default as new servers are likely to have this value near 0.

chatmaild/src/chatmaild/ini/chatmail.ini.f
```diff
@@ -42,6 +42,23 @@
...
+# Minimum free disk space of mailboxes dir for new registrations.
+#min_free_disk_space = 1G
+
+# Maximum number of concurrent IMAP connections overall.
+#max_imap_connections = 10000
+
+# Maximum number of concurrent SMTP connections overall
+# on submission/smtps ports.
+#max_smtp_connections = 1000
+
```
j4n, line 71: These new lower(?) default settings should be well communicated to the existing relay operators so to not cause surprises.

chatmaild/src/chatmaild/tests/test_doveauth.py
```diff
@@ -202,3 +203,18 @@ def lookup():
         res = results.get()
         if res is not None:
             pytest.fail(f"concurrent lookup failed\n{res}")
+
+
+def test_has_sufficient_resources_accepts_normal(make_config):
+    config = make_config("chat.example.org")
+    assert has_sufficient_resources(config)
```
j4n: this might fail on a busy ci worker possibly, maybe set really low values similar to test below?

chatmaild/src/chatmaild/doveauth.py
```diff
@@ -25,6 +26,46 @@ def encrypt_password(password: str):
     return "{SHA512-CRYPT}" + passhash
 
 
+def get_available_memory_mb():
+    with open("/proc/meminfo") as f:
```
j4n: This will fail hard on non-Linux platforms such as FreeBSD, maybe wrap in try/catch or do a check on `sys.platform` or `os.path.exists("/proc/meminfo")` so we can fail gracefully, as in the CPU check

j4n (commented):
Looks good to me, can't speak to the validity of the default limits; two robustness changes suggested.

chatmaild/src/chatmaild/newemail.py
```diff
@@ -42,6 +43,14 @@ def create_dclogin_url(config, email, password):
 
 def print_new_account():
     config = read_config(CONFIG_PATH)
+
+    if not has_sufficient_resources(config):
```
missytake: This causes a 502 Bad Gateway error, because cgi-bin doesn't have the permissions to stat `/home/vmail`.

It can be removed completely, the error message is shown to users via IMAP's authentication failed, not as an HTTP error.

link2xt: There is also a pending change https://github.com/chatmail/relay/pull/858 to not point the users to the https URL.

And for automatic multi-relay onboarding where these limits matter the most we should also skip this HTTPS request.

chatmaild/src/chatmaild/ini/chatmail.ini.f
hpk42: they are not lower i think, at least for SMTP but IMAP connections needs higher limit for big servers. We can tune these limits after the PR is merged.   
image

We can tweak defaults after the PR is merged.  It remains to be seen in reality.  At least there is now an "automatic" mechanism for rejecting addresses.

j4n: `process_limit` in `cmdeploy/src/cmdeploy/dovecot/dovecot.conf.j2` was 50k, now 10k; for smtpd it was 5k now 1k if I read the diff correctly.

chatmaild/src/chatmaild/tests/test_doveauth.py
hpk42: fixed it.

chatmaild/src/chatmaild/doveauth.py
hpk42: fixed through using pustil.

chatmaild/src/chatmaild/ini/chatmail.ini.f
hpk42: yes, because the limits were adjusted for nine.  Now we are trying to adjust for minimal-requirements.

j4n (approved):
looks great to me!

PR #1022, opened 2026-07-22
refactor(postfix): remove unused "filter" lmtp service

link2xt:
It looks like the reason it exists in the default configuration is an example for <https://www.postfix.org/FILTER_README.html> Using this filter requires configuration such as
"-o content_filter=filter:dummy" and we don't have it.

### DIFF
--- a/cmdeploy/src/cmdeploy/postfix/master.cf.j2
+++ b/cmdeploy/src/cmdeploy/postfix/master.cf.j2
@@ -76,7 +76,6 @@ lmtp      unix  -       -       y       -       -       lmtp
 anvil     unix  -       -       y       -       1       anvil
 scache    unix  -       -       y       -       1       scache
 postlog   unix-dgram n  -       n       -       1       postlogd
-filter    unix -        n       n       -       -       lmtp
 # Local SMTP server for reinjecting outgoing filtered mail.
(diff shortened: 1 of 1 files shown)


link2xt:
Looks like if CI passes then it should be ok, but maybe someone wants to double-check that it is really not used.

The reason i looked at this: i was checking where do LMTP filters apply for https://github.com/chatmail/relay/pull/1005

missytake (approved):
The process doesn't seem to be running on nine, probably because nothing calls it:

```
nine ~ # ps aux | grep filter | grep -v smtpd_proxy_filter | grep -v lmtp-filtermail
vmail     368216  0.5  0.5 1028460 373564 ?      Ssl  Jul02 167:10 /usr/local/bin/filtermail /usr/local/lib/chatmaild/chatmail.ini outgoing
vmail     368321  0.1  0.5 950328 352160 ?       Ssl  Jul02  54:41 /usr/local/bin/filtermail /usr/local/lib/chatmaild/chatmail.ini transport
root     1197503  0.0  0.0   6340  2208 pts/0    S+   22:26   0:00 grep --color=auto filter
vmail    4028937  0.0  0.1 993264 98748 ?        Ssl  Jul06  20:59 /usr/local/bin/filtermail /usr/local/lib/chatmaild/chatmail.ini incoming
```

PR #1023, opened 2026-07-26
chore: merge chatmail/filtermail into chatmail/relay to have a monorepo

missytake:
This is a proposal to merge chatmail/filtermail and chatmail/relay into one monorepo, so we can change filtermail with only one PR in the future.

The paths in the filtermail commits have been rewritten so that they were always in the filtermail/ subdirectory, so the commit history of each file is preserved. What I did to merge the repositories, in case it needs to be repeated slightly differently:

```
cd $USER/code/relay/..
git clone https://github.com/chatmail/filtermail
cd filtermail
git filter-repo --to-subdirectory filtermail --tag-rename v:filtermail-
cd ../relay
git checkout main
git pull --rebase
git checkout -b merge-filtermail
git remote add filtermail-merge ../filtermail
git fetch filtermail-merge
git merge --allow-unrelated-histories filtermail-merge/main
```

Then I did 3cf3937 on top of it to integrate filtermail's .github/ and .gitignore into chatmail/relay.

open questions:
- how should we distinguish tags? does filtermail still need its own software versions? Other projects import filtermail, afaik. I renamed the filtermail tags from v1.2.3 to filtermail-1.2.3 for now, but didn't push them so far.
- should we rename the relay tags as well, from 1.2.3 to relay-1.2.3?


j-g00da:
> does filtermail still need its own software versions?

In theory it does not, but then we probably want to tag relay more often, as it won't be possible to deploy relay from main if filtermail changed without compiling it locally.

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,70 @@
+on:
+  release:
+    types: [published]
+
+
+jobs:
+  build:
+    permissions: {}
```
j4n: Again unfiltered, this fires on relay releases too and will build filtermail and attach it to the release.

Also: adding top-level `permissions: {}` would match the other workflows.

```suggestion
on:
  release:
    types: [published]

permissions: {}

jobs:
  build:
    if: startsWith(github.ref_name, 'filtermail-')
```

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,70 @@
...
+    name: Build binaries.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+          path: filtermail
```
j4n: Original repo had no `path:` because repo root was crate root. Keeping `path: filtermail` puts the repo at `./filtermail` and the crate at `./filtermail/filtermail`, so the `cd filtermail` below lands at repo root, where there's no `Cargo.toml`, so the build can't find the crate.


```suggestion
      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
        with:
          persist-credentials: false
```

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,70 @@
...
+      - uses: taiki-e/install-action@2ca9b94c269419b7b0c711c09d0b21c4e1d51145 # v2.83.1
+        with:
+          tool: cargo-zigbuild
+      - name: Build
+        run: |
+          RUSTFLAGS="-Ctarget-feature=+crt-static -Clink-self-contained=yes" \
+          cd filtermail && cargo zigbuild \
+            --release \
+            --target x86_64-unknown-linux-musl \
+            --target aarch64-unknown-linux-musl
```
j4n, line 33: Here, `RUSTFLAGS` is attached to cd instead of cargo and that means its lost silently, match pattern from `filtermail-ci.yml`.

```suggestion
      - name: Build
        working-directory: filtermail
        run: |
          RUSTFLAGS="-Ctarget-feature=+crt-static -Clink-self-contained=yes" \
          cargo zigbuild \
            --release \
            --target x86_64-unknown-linux-musl \
            --target aarch64-unknown-linux-musl
```

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          name: filtermail-x86_64
+          path: target/x86_64-unknown-linux-musl/release/filtermail
+          if-no-files-found: error
+      - name: Upload aarch64 binary
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: filtermail-aarch64
+          path: target/aarch64-unknown-linux-musl/release/filtermail
+          if-no-files-found: error
```
j4n, line 45: The artifact paths need to be adapted too:

```suggestion
      - name: Upload x86_64 binary
        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
        with:
          name: filtermail-x86_64
          path: filtermail/target/x86_64-unknown-linux-musl/release/filtermail
          if-no-files-found: error
      - name: Upload aarch64 binary
        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
        with:
          name: filtermail-aarch64
          path: filtermail/target/aarch64-unknown-linux-musl/release/filtermail
          if-no-files-found: error
```

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,70 @@
...
+
+  publish:
+    name: Upload binaries to the release
+    needs: ["build"]
+    runs-on: ubuntu-latest
+    steps:
```
j4n, line 53: `gh release upload` needs `publish` has no `contents: write` with the top-level `permissions: {}`.

```suggestion
  publish:
    name: Upload binaries to the release
    needs: ["build"]
    runs-on: ubuntu-latest
    permissions:
      contents: write
    steps:
```

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Download aarch64 binary
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: filtermail-x86_64
+          path: filtermail-x86_64
+      - name: Download x86_64 binary
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: filtermail-aarch64
+          path: filtermail-aarch64
```
j4n: Nit: the two step names are swapped. 

```suggestion
      - name: Download x86_64 binary
        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
        with:
          name: filtermail-x86_64
          path: filtermail-x86_64
      - name: Download aarch64 binary
        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
        with:
          name: filtermail-aarch64
          path: filtermail-aarch64
```

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Upload binaries to the GitHub release
+        if: github.event_name == 'release'
```
j4n, line 65: Nit: `on: release` is the only trigger, so `github.event_name` is always `release`.

```suggestion
      - name: Upload binaries to the GitHub release
```

.github/workflows/filtermail-ci.yml
```diff
@@ -0,0 +1,107 @@
+name: CI
```
j4n: `ci.yaml` is also `name: CI` and thus  these two workflows will cancel each other on main.

.github/workflows/filtermail-ci.yml
```diff
@@ -0,0 +1,107 @@
...
+permissions: {}
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
j4n, line 15: This is unfiltered, and every relay PR runs the full Rust matrix plus a second LXC suite. Ideally, we fold `relay-test` and `ci.yaml`'s `lxc-test` into one job.

```suggestion
on:
  workflow_dispatch:
  push:
    branches:
      - main
    paths:
      - 'filtermail/**'
      - '.github/workflows/filtermail-ci.yml'
  pull_request:
    branches:
      - main
    paths:
      - 'filtermail/**'
      - '.github/workflows/filtermail-ci.yml'
```

filtermail/Cargo.toml
```diff
@@ -0,0 +1,61 @@
+[package]
+name = "filtermail"
+version = "0.7.4"
+edition = "2024"
+license = "MIT"
+readme = "README.md"
+authors = ["Chatmail Developers"]
+description = "A mail filtering service used by chatmail relay"
+repository = "https://github.com/chatmail/filtermail"
```
j4n: Stale after the merge.

```suggestion
repository = "https://github.com/chatmail/relay"
```

j4n (commented):
Thanks for tackling this, two non-inline comments:
- [ ] `ci.yaml` still polls the old filtermail release, adapt that to the build pattern from `filtermail-ci.yaml`.
- [ ] Changelog: There are now two and two `cliff.toml`; needs some `--include-path` work so this won't clobber.

Re versioning:  prefixed `filtermail-X.Y.Z` tags, with relay staying unprefixed, seems the conventional shape; I would not retroactively rename relay's tags to `relay-X.Y.Z` as it breaks existing release-download URLs.

.github/workflows/filtermail-release.yml
missytake: Not anymore, now it's `tags: filtermail-*`

cmdeploy/src/cmdeploy/mtail/filtermail.mtail
j4n, line 1: This file should go to cmdeploy/src/cmdeploy/mtail so its in the canonical place and #1036 works.

missytake:
CI requires https://github.com/chatmail/cmlxc/pull/31 I think - **edit:** apparently not, it's green :)

.github/workflows/ci.yaml
```diff
@@ -61,6 +45,10 @@ jobs:
     with:
       cmlxc_version: main
       cmlxc_commands: |
+        sudo apt-get update && sudo apt-get install -y musl-tools
+        rustup target add x86_64-unknown-linux-musl
+        export CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER=musl-gcc
+        cd repo/filtermail && RUSTFLAGS="-Ctarget-feature=+crt-static -Clink-self-contained=yes" cargo build --release --target x86_64-unknown-linux-musl
         cmlxc init 
         # single cmdeploy relay test
         cmlxc -v deploy-cmdeploy --source ./repo cm0
```
j4n:
```suggestion
        cmlxc -v deploy-cmdeploy --source ./repo --filtermail repo/filtermail/target/x86_64-unknown-linux-musl/release/filtermail cm0
```
if we do build filtermail at all times, we should probably also use it in the test, otherwise we dont need to build it; though we will have a path clash here that will need fixing in cmlxc.

.github/workflows/ci.yaml
j4n:
```suggestion
        cmlxc -v deploy-cmdeploy --source ./repo --ipv4-only --filtermail repo/filtermail/target/x86_64-unknown-linux-musl/release/filtermail cm1
```
same as above

.github/workflows/ci.yaml
```diff
@@ -61,6 +45,10 @@ jobs:
     with:
       cmlxc_version: main
       cmlxc_commands: |
+        sudo apt-get update && sudo apt-get install -y musl-tools
+        rustup target add x86_64-unknown-linux-musl
+        export CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER=musl-gcc
```
j4n:
```suggestion
```
Looking at the job log, does not reach the next step as each line runs in its own environment and thus is a noop.

cliff.toml
```diff
@@ -92,3 +92,4 @@ topo_order_commits = true
 sort_commits = "oldest"
 # Process submodules commits
 recurse_submodules = false
+exclude_paths = ["filtermail/"]
```
j4n, line 95:
```suggestion
exclude_paths = ["filtermail/"]
tag_pattern = '^[0-9]+\.[0-9]+\.[0-9]+$'
```
Exclude filtermail tags and match only relay tags.

filtermail/cliff.toml
```diff
@@ -0,0 +1,94 @@
...
+use_branch_tags = false
+# Order releases topologically instead of chronologically.
+topo_order = false
+# Order commits topologically instead of chronologically.
+topo_order_commits = true
+# Order of commits in each group/release within the changelog.
+# Allowed values: newest, oldest
+sort_commits = "oldest"
+# Process submodules commits
+recurse_submodules = false
```
j4n, line 94:
```suggestion
recurse_submodules = false
include_paths = ["filtermail/"]
tag_pattern = "^filtermail-"
```
match only filtermail

.github/workflows/chatmaild-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        # Checkout pull request HEAD commit instead of merge commit
+        # Otherwise `test_deployed_state` will be unhappy.
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+      - name: download filtermail
+        run: curl -L https://github.com/chatmail/filtermail/releases/download/v0.7.4/filtermail-x86_64 -o /usr/local/bin/filtermail && chmod +x /usr/local/bin/filtermail
```
j4n:
```suggestion
        run: curl -f -L https://github.com/chatmail/filtermail/releases/download/v0.7.4/filtermail-x86_64 -o /usr/local/bin/filtermail && chmod +x /usr/local/bin/filtermail
```
`-f` means failing on 404 instead of downloading the error page to filtermail binary location. Maybe we should  consider always building and using https://github.com/Swatinem/rust-cache as in filtermail-ci.yaml.

.github/workflows/filtermail-ci.yml
```diff
@@ -0,0 +1,79 @@
...
+    paths:
+      - "filtermail/**"
```
j4n, line 17:
```suggestion
    paths:
      - "filtermail/**"
      - ".github/workflows/filtermail-ci.yml"
  pull_request:
    branches:
      - main
    paths:
      - "filtermail/**"
      - ".github/workflows/filtermail-ci.yml"
```
`push` otherwise runs the full rust matrix for nothing on pushes to main; and add the workflow file.

filtermail/README.md
```diff
@@ -0,0 +1,203 @@
...
+  (Trivial changes may still be considered,
+  please open an issue to discuss any such changes before working on them).
+
+
+## Releases
+
+Filtermail is distributed as a statically linked linux binary,
+available for `x86_64` and `aarch64` architectures.
+
+Binaries are available on the [releases page](https://github.com/chatmail/filtermail/releases).
```
j4n, line 196: This needs to be updated.

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,75 @@
+on:
+  push:
+    tags:
+      - 'filtermail-*'
+
+permissions: {}
+
+jobs:
+  build:
+    name: Build binaries.
```
j4n, line 9:
```suggestion
    name: Build binaries.
    if: startsWith(github.event.release.tag_name, 'filtermail-')
```
Guard here..

.github/workflows/filtermail-release.yml
```diff
@@ -0,0 +1,75 @@
+on:
+  push:
+    tags:
+      - 'filtermail-*'
```
j4n:
```suggestion
  release:
    types: [published]
```
see filter on line 10

j4n (approved):
Changes looks good. Some more comments inline.

One thing for a followup-PR: `RELEASE.md` is currently relay-only and should have a filtermail section, something like:

```md
# Releasing chatmail relay
...

# Releasing filtermail

filtermail is versioned independently and released from the same repo,
using `filtermail-` prefixed tags. To release filtermail 0.7.5:

1. Update the changelog:
   `git cliff --config filtermail/cliff.toml -u -t filtermail-0.7.5 -p filtermail/CHANGELOG.md`
2. Bump `version` in `filtermail/Cargo.toml` and commit `filtermail/Cargo.lock`.
3. Commit with the message `feat(release): prepare filtermail for 0.7.5`;  or chore() if it should be excluded from relay changelog.
4. Open a PR with the new commit, merge it to main after review.
5. In the web interface, create a GitHub release, tell it to create a new tag `filtermail-0.7.5`. Do not push the tag by hand: the release must exist before the upload job runs.
6. Bump the pinned version and sha256sums in`cmdeploy/src/cmdeploy/filtermail/deployer.py`.
```

filtermail/cliff.toml
missytake: The tag_pattern makes sense, but the include path is set implicitly by running git cliff in filtermail/

filtermail/README.md
missytake: Hm, how do we want to distribute binaries? Maybe https://download.delta.chat/filtermail? I couldn't see a way to filter github.com/chatmail/relay/releases by tag prefix :/ searching for `?q=tag:1.12.0` matches one release, but `?q=tag:1.1` matched none

missytake:
> 6. Bump the pinned version and sha256sums in`cmdeploy/src/cmdeploy/filtermail/deployer.py`.

Hm, I added it - but this means we again need 2 PRs until filtermail changes arrive in cmdeploy? Wasn't this the purpose of this PR, that this is simpler now, e.g. by building filtermail in cmdeploy if there are changes?

.github/workflows/ci.yaml
missytake: I assume this is the path clash you mean? https://github.com/chatmail/relay/actions/runs/32125762857/job/95675867660

j-g00da:
> but this means we again need 2 PRs until filtermail changes arrive in cmdeploy? Wasn't this the purpose of this PR, that this is simpler now, e.g. by building filtermail in cmdeploy if there are changes?

Yes I brought up this topic on a group chat some time ago. If we don't come up with a way to do this in one PR, then I don't really see any benefit of merging the repos. In fact it IMO only makes things more confusing.

missytake:
> > but this means we again need 2 PRs until filtermail changes arrive in cmdeploy? Wasn't this the purpose of this PR, that this is simpler now, e.g. by building filtermail in cmdeploy if there are changes?
> 
> Yes I brought up this topic on a group chat some time ago. If we don't come up with a way to do this in one PR, then I don't really see any benefit of merging the repos. In fact it IMO only makes things more confusing.

Hm, the options I see:
- stop having filtermail releases, and using the main branch instead:
  - a) We can build filtermail from source during cmdeploy run. (this is basically what we do with chatmaild.)
  - b) We could have a github action which builds filtermail and uploads it to https://download.delta.chat/filtermail on pushes to main which change the filtermail/ directory.
    - Any PR would need to update the sha256sum in cmdeploy, a github action or git hook could take care we don't forget it.
- c) keep it as it is.

What do you think? I think I prefer option b).

j-g00da:
I think a) won't work as IIRC we don't want to force users to build it locally. I like option b), but then what do we want to do with version string, ideally we would auto-bump it on each change.

Alternative, (let's call it "d)") I had in mind was to use the latest available binary _or_ build from source if there were any changes in filtermail subdir. This way we can still build filtermail only when tagging relay, but if someone doesn't mind building from source - they can use main branch.

I'm fine with either b) or d) whichever seems simpler.

hpk42:
this filtermail/relay merging all turned into pretty much of a quagmire :sweat_smile:   Considering that we actually want to go for a new single-file installer i don't think it's worth to spend our collective time on this PR any further, i am afraid.  I don't have easy answers to some of the questions above, and also think that if "single-PR" changes of both relay and filtermail don't become reliably and somewhat easily possible,  we don't win much. 

So i suggest for now to close this PR, and stay with the two-repo solution until we get more clarity on the single-file installer layout/changes.   Sorry for triggering this two-repo merge early on without thinking it through, and thanks Nami and all for engaging.

missytake:
Sounds good to me :)

PR #1025, opened 2026-07-27
chore(ci): update doc staging upload path

j4n:
This updates the doc staging workflow to the relative path needed for rrsync on www / delta.chat.


hpk42 (approved):
can be merged if the CI actually passes :)

PR #1027, opened 2026-07-28
refactor: install dns-root-data instead of using unbound-anchor

### LINKED ISSUE #542: cmdeploy running unbound-anchor is redundant
https://github.com/chatmail/relay/blob/5ba99dc78214808d0e95b86e7e4c37b974b7eef3/cmdeploy/src/cmdeploy/__init__.py#L595-L601

This is redundant as the unbound service automatically calls this:

```
[Unit]
Description=Unbound DNS server
Documentation=man:unbound(8)
After=network.target
Before=nss-lookup.target
Wants=nss-lookup.target

[Service]
Type=notify
Restart=on-failure
EnvironmentFile=-/etc/default/unbound
ExecStartPre=-/usr/libexec/unbound-helper chroot_setup
ExecStartPre=-/usr/libexec/unbound-helper root_trust_anchor_update
ExecStart=/usr/sbin/unbound -d -p $DAEMON_OPTS
ExecStopPost=-/usr/libexec/unbound-helper chroot_teardown
ExecReload=+/bin/kill -HUP $MAINPID

[Install]
WantedBy=multi-user.target
```

It's handled by `ExecStartPre` in the systemd unit file

link2xt:
unbound-anchor retrieves root trust anchor directly via HTTPS, but Debian already distributes the same files in dns-root-data package.

Closes #542

### DIFF
--- a/cmdeploy/src/cmdeploy/deployers.py
+++ b/cmdeploy/src/cmdeploy/deployers.py
@@ -138,9 +138,22 @@ class UnboundDeployer(Deployer):
         # On an IPv4-only system, if unbound is started but not configured,
         # it causes subsequent steps to fail to resolve hosts.
         with blocked_service_startup():
+            # dns-root-data is an optional package
+            # that contains /usr/share/dns/root.key
+            #
+            # This file is copied into /var/lib/unbound/root.key
+            # at the start of "unbound" systemd unit
+            # by /usr/libexec/unbound-helper shell script
+            # from the "unbound" package as of version 1.17.1-2+deb12u4
+            #
+            # The same /var/lib/unbound/root.key can be retrieved directly
+            # following the procedure from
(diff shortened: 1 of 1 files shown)


link2xt:
We can also uninstall unbound-anchor. See also my comment https://github.com/chatmail/relay/issues/542#issuecomment-5110472502, this is not really fixing anything, just trading "upstream" approach for a "Debian-way" approach. We can also document the current approach instead, any option is fine, and the issue is not urgent, the only reason i looked into it now is the "wrong answer" submitted at https://github.com/chatmail/relay/pull/1026 (the combination of not installing dns-root-data and a mock test)

missytake (approved):
This is much nicer, also removes a server.shell call which is always nice.

hpk42 (approved):

PR #1028, opened 2026-07-29
restrict usable workflow to read permissions

hpk42:


j4n (approved):
looks good

.github/workflows/ci-no-dns.yaml
```diff
@@ -20,6 +20,8 @@ concurrency:
 jobs:
   no-dns:
     name: LXC deploy and test
+    permissions:
+      contents: read
```
link2xt, line 24: This adds more permissions. All permissions are already removed above on the workflow level.

hpk42:
On Wed, Jul 29, 2026 at 12:24 -0700, l wrote:
> @link2xt commented on this pull request.
> > @@ -20,6 +20,8 @@ concurrency:
>  jobs:
>    no-dns:
>      name: LXC deploy and test
> +    permissions:
> +      contents: read
> 
> This adds more permissions. All permissions are already removed above on the workflow level.

I blindly applied the same pattern as defined in
https://github.com/chatmail/workflows/blob/main/templates/ci.yml
where there is no "permissions: {}", so "contents: read" restricts
GITHUB_TOKEN to just "read" instead of the repo default.

Here it does add permissions, as you say. 
Removed in https://github.com/chatmail/relay/pull/1033

PR #1029, opened 2026-07-29
fix: Proxy the actual Iroh proxy endpoints used by 0.35 and 1.0

hpk42:
drop stale /relay/probe from pre-0.35 iroh versions


link2xt (approved):
For reference, here is 0.35: https://github.com/n0-computer/iroh/blob/v0.35.0/iroh-relay/src/http.rs

hpk42:
i also think the missing /ping prevents initializing iroh with multiple relays and let it choose itself. So it's better to get relays serve /ping soon.

PR #1030, opened 2026-07-29
feat: setup websockify

link2xt:
This change makes it possible
to connect the client over WebSocket.
Running `websocat wss://example.org/imap`
connects to the IMAP
and running `websocat wss://example.org/smtp`
connects to SMTP (submission service, not port 25).

Connecting from a web client will still need a CORS policy for clients not running on the same domain,
but for non-web clients this is already usable.

### DIFF
--- a/cmdeploy/src/cmdeploy/websockify/websockify-submission.service
+++ b/cmdeploy/src/cmdeploy/websockify/websockify-submission.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=WebSocket proxy for SMTP
+After=network.target
+
+[Service]
+Type=simple
+Restart=always
+ExecStart=/usr/bin/websockify 127.0.0.1:8587 localhost:587
+DynamicUser=true
+
+[Install]
+WantedBy=multi-user.target
--- a/cmdeploy/src/cmdeploy/deployers.py
+++ b/cmdeploy/src/cmdeploy/deployers.py
@@ -32,6 +32,7 @@ from .external.deployer import ExternalTlsDeployer
 from .filtermail.deployer import FiltermailDeployer
 from .mtail.deployer import MtailDeployer
 from .nginx.deployer import NginxDeployer
+from .websockify.deployer import WebsockifyDeployer
 from .opendkim.deployer import OpendkimDeployer
 from .postfix.deployer import PostfixDeployer
 from .selfsigned.deployer import SelfSignedTlsDeployer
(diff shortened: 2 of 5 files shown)


link2xt:
The next step is adding some support to the clients, with manual configuration for now. WebSocket can eventually phase out ALPN-based stream multiplexing, but i'm not removing it and it can coexist with websockets, there is no need to migrate existing users. Discovery of websockets, e.g. extending autoconfig XML or QR codes, is also left for the future.

DarkCat09:
What about listening on a unix socket?
Currently, with 443 multiplexing, we have nginx opening one more TCP connection to a local IMAPS/SMTPS port, that takes one available port.
If deploying websockify exactly as this PR currently suggests, we would take 2 extra local ports instead of 1: first when nginx does proxy_pass to websockify, and second when websockify proxies to the target.
Websockify supports listening on a UDS, we could use it instead -- which also removes the TCP overhead.

DarkCat09:
Also, did you measure the performance and resource usage of websockify?
I'm kind of skeptical about a web server written in Python.

link2xt:
> What about listening on a unix socket?

I'm using dynamicuser currently to avoid dealing with filesystem permissions. With unix socket we will need to think about how to make nginx access it, who creates the socket and where etc. Don't think using 2 ports is a problem.

> Also, did you measure the performance and resource usage of websockify?
> I'm kind of skeptical about a web server written in Python.

The reason i have set it up is because it is already avaliable in Debian. Don't know if it will cause problems, but it's good enough for testing. We need core implementation first that will allow to setup it manually, but even then it will not be used by a lot of people because it still needs manual configuration.

link2xt:
I opened corresponding issue https://github.com/chatmail/core/issues/8559 in the core, it is somewhat straightforward but i have not looked deeply into tokio-tungstenite and how difficult it will be to use custom connection establishment.

cmdeploy/src/cmdeploy/websockify/websockify-submission.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=WebSocket proxy for SMTP
+After=network.target
+
+[Service]
+Type=simple
+Restart=always
+ExecStart=/usr/bin/websockify 127.0.0.1:8587 localhost:587
```
hpk42, line 8: we probably need a separate SMTP submission from 587, that does not require tls.

PR #1033, opened 2026-07-30
cleanup: contents:read not needed for relay repo

hpk42:
public repos need no contents::read and there were permissions: {}


link2xt (approved):
As long as lxc-test.yml does not start requiring read permissions this should work. If really needed private repo can just copy the workflow instead of using reusable workflows.

hpk42:
> As long as lxc-test.yml does not start requiring read permissions this should work. If really needed private repo can just copy the workflow instead of using reusable workflows.

Well, my goal is to get away from copying workflow files around.  It's pretty messy, and very easy to diverge.

PR #1038, opened 2026-08-12
feat: serve an appversions.json index file to clients via IMAP metadata

### LINKED ISSUE #1037: add "client version information" to IMAP METADATA
we aim to inform about updates for installations outside of any appstore soon.

the information about app versions shall be provided by the relays.

for that, we need to have the following:

- add an `APPVERSIONS.json` to the root of the repo. this file is the "source of truth" and has the following format - the following is also the valid version information for _now_:
    ```
    {
       "clients": [
          {
            "clientId": "deltachat",
            "sources": [
              {
                "sourceId": "gplay",
                "versionInteger": 754,
                "versionString": "2.57.0",
                "downloadUrl": "https://github.com/deltachat/deltachat-android/releases/download/v2.57.0/deltachat-gplay-release-2.57.0.apk"
              }
            ]
          }
      ]
    }
    ```
    
    (in the future, the file will be edited on a regular base by PRs, by ppl without much knowledge about this repo (as me :), so it should not be deeply nested in some folder. we may aim for some automation for some clients, or even split the file to sort of `.ini` but that is another can of worms, and it is unclear if that is needed)

- the file shall be shipped together with relay installations, content of the file shall be passed as IMAP METADATA under the key `/shared/vendor/deltachat/appversions`
    
    that alone would allow updates as soon as a relay is updated. not sure how often that happens in practise, maybe this is already good enough (core is smart enough to ignore outdated versions, see https://github.com/chatmail/core/pull/8557 for some discussion)

- if relays are not updated every some week in practise, the relay installation shall update their `APPVERSIONS.json` from `https://github.com/chatmail/relay/APPVERSIONS.json` every 24 hours, as best effort. if the update cannot be done, then it's that, and relay should continue with existing `APPVERSIONS.json`

hpk42:
fixes #1037 
This is designed to help implement self-updating APKs (and later other clients), see core counterpart https://github.com/chatmail/core/pull/8557

### DIFF
--- a/chatmaild/src/chatmaild/defaults/appversions.json
+++ b/chatmaild/src/chatmaild/defaults/appversions.json
@@ -1,0 +1,15 @@
+{
+  "clients": [
+    {
+      "clientId": "deltachat",
+      "sources": [
+        {
+          "sourceId": "gplay",
+          "versionInteger": 754,
+          "versionString": "2.57.0",
+          "downloadUrl": "https://github.com/deltachat/deltachat-android/releases/download/v2.57.0/deltachat-gplay-release-2.57.0.apk"
+        }
+      ]
+    }
+  ]
+}
--- a/chatmaild/src/chatmaild/metadata.py
+++ b/chatmaild/src/chatmaild/metadata.py
@@ -1,8 +1,10 @@
+import json
 import logging
 import socket
 import sys
 import time
 from contextlib import contextmanager
+from importlib.resources import files
 
 from .config import read_config
 from .dictproxy import DictProxy
@@ -16,6 +18,18 @@ def turn_credentials(turn_socket_path):
         client_socket.connect(turn_socket_path)
         with client_socket.makefile("rb") as file:
             return file.readline().decode("utf-8").strip()
+
+
+def read_appversions(path):
+    try:
+        data = json.loads(path.read_bytes())
+    except FileNotFoundError:
+        return None
+    except (OSError, ValueError):
+        logging.exception(f"failed to read {path}")
+        return None
+    # the dict protocol is line-based, keep the value single-line
+    return json.dumps(data, separators=(",", ":"))
 
 
 def _is_valid_token_timestamp(timestamp, now):
@@ -101,6 +115,7 @@ class MetadataDictProxy(DictProxy):
         self.iroh_relay = iroh_relay
         self.turn_hostname = turn_hostname
         self.turn_socket_path = turn_socket_path
+        self.appversions_path = files(__package__).joinpath("defaults/appversions.json")
 
     def handle_lookup(self, parts):
         # Lpriv/43f5f508a7ea0366dff30200c15250e3/devicetoken\tlkj123poi@c2.testrun.org
@@ -125,6 +140,9 @@ class MetadataDictProxy(DictProxy):
                         case "maxsmtprecipients":
                             # postfix default  (see "postconf smtpd_recipient_limit")
                             return "O1000\n"
+                        case "appversions":
+                            value = read_appversions(self.appversions_path)
+                            return f"O{value}\n" if value else "N\n"
 
         logging.warning(f"lookup ignored: {parts!r}")
         return "N\n"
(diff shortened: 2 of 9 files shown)


adbenitez (approved):
lgtm

chatmaild/src/chatmaild/defaults/appversions.json
```diff
@@ -0,0 +1,15 @@
+{
+  "clients": [
+    {
+      "clientId": "deltachat",
```
missytake, line 4:
```suggestion
      "clientId": "deltachat-android",
```
wouldn't that be more clear, when deltachat-desktop gets added as well?

chatmaild/src/chatmaild/config.py
```diff
@@ -63,6 +63,9 @@ def __init__(self, inipath, params):
         self.turn_socket_path = params.pop(
             "turn_socket_path", "/run/chatmail-turn/turn.socket"
         )
+        self.appversions_path = Path(
+            params.pop("appversions_path", "/usr/local/lib/chatmaild/appversions.json")
+        )
```
missytake: Not sure whether this needs to be a configurable value tbh. I think hardcoding this path would be fine. Not so important though, as operators don't really need to find out about it.

cmdeploy/src/cmdeploy/deployers.py
```diff
@@ -126,6 +126,11 @@ def _configure_remote_venv_with_chatmaild(deployer, config) -> None:
         dest=remote_chatmail_inipath,
     )
 
+    deployer.put_file(
+        src=get_reporoot().joinpath("APPVERSIONS.json").open("rb"),
```
missytake: Hm, I think the file should rather be in cmdeploy/src/cmdeploy than reporoot - if we link to it from CONTRIBUTING.md, client devs will be able to find it I think. But that's just cleanliness preference I guess.

missytake (approved):
No big deal to merge this I think. three small nitpicks:

chatmaild/src/chatmaild/defaults/appversions.json
r10s: that would go to "sources" - the idea is to have other clients as parla or whatever on that level. in other words, a client can be multi-platform, and it is up to the client to make "source" unique

chatmaild/src/chatmaild/defaults/appversions.json
```diff
@@ -0,0 +1,15 @@
+{
```
j4n, line 1: Should this file maybe have a timestamp?

j4n (approved):
Looks good from what I can understand, however, the 24h refresh suggestion from #1037 seems worth mentioning as being not addressed yet; relay releases were indeed few and far between compared to client releases in the last months.

cmdeploy/src/cmdeploy/deployers.py
hpk42: agreed, discussed with @r10s and shifted it to chatmaild python package into a `chatmaild/defaults/appversions.json`  (no need to capitalize as is only common at repo root) where we can also move the chatmail.ini.

chatmaild/src/chatmaild/defaults/appversions.json
hpk42: it wouldn't help much currently with freshness of app versions to users. Freshness pipe from relay-repo -> relay-deployment -> core reading new metadata -> UI showing "update available" rather depends on core currently.   it's not clear what checks/automatizations we want to do, so let's stick with the minimum data for now.

chatmaild/src/chatmaild/config.py
hpk42: i removed it completely.

chatmaild/src/chatmaild/metadata.py
```diff
@@ -18,6 +20,18 @@ def turn_credentials(turn_socket_path):
...
+def read_appversions(path):
+    try:
+        data = json.loads(path.read_bytes())
+    except FileNotFoundError:
+        return None
+    except (OSError, ValueError):
+        logging.exception(f"failed to read {path}")
+        return None
+    # the dict protocol is line-based, keep the value single-line
+    return json.dumps(data, separators=(",", ":"))
```
hpk42, line 32: the separators are just to minimize the generated string. they would be ", " and ": " otherwise :)

PR #1041, opened 2026-08-24
ci: try to fix lack of RFC822 item support in madmail and make CI pass

hpk42:
some local debugging revealed madmail v2 does not support fetching RFC822 items (gives an empty body) so try BODY.PEEK[] instead.


j4n (approved):
looks good to me, thanks for the quick fix. Surprised it didnt come up on the previous run.

PR #1043, opened 2026-08-25
chore: follow the new mtail release source, upgrade 3.0.8 to 3.4.9

hpk42:
see https://github.com/google/mtail/issues/929 for the "official" move.  FWIW nix also seems to follow this new source https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/mt/mtail/package.nix


link2xt:
Debian also points to the new homepage now at https://packages.debian.org/sid/mtail
I also checked that this maintainer is listed in https://github.com/google/mtail/graphs/contributors for a long time, getting your account there requires at least hijacking the email address as far as i understand, even if commits are not signed.

link2xt (approved):

PR #1044, opened 2026-08-25
chore: un-hardcode executable paths in some systemd service files

hpk42:
makes it consistent with the other services.

### DIFF
--- a/cmdeploy/src/cmdeploy/basedeploy.py
+++ b/cmdeploy/src/cmdeploy/basedeploy.py
@@ -51,7 +51,7 @@ def get_resource(arg, pkg=__package__):
     return importlib.resources.files(pkg).joinpath(arg)
 
 
-def configure_remote_units(deployer, mail_domain, units) -> None:
+def configure_remote_units(deployer, mail_domain, units, **kwargs) -> None:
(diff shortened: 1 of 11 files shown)


cmdeploy/src/cmdeploy/basedeploy.py
```diff
@@ -51,7 +51,7 @@ def get_resource(arg, pkg=__package__):
     return importlib.resources.files(pkg).joinpath(arg)
 
 
-def configure_remote_units(deployer, mail_domain, units) -> None:
+def configure_remote_units(deployer, mail_domain, units, bin_path=None) -> None:
```
link2xt: Maybe add `**kwargs` instead of `bin_path` like in `ensure_systemd_unit`, and say this is passed to all systemd unit templates.

link2xt (approved):

cmdeploy/src/cmdeploy/basedeploy.py
hpk42: ok, i changed it.  FWIW I try to avoid passing `**kwargs` around usually, because looking up a function that passes it further down hides what the possible parameters are.

PR #1049, opened 2026-08-31
feat: move doveauth from dictproxy to lua/http as recommended by dovecot 2.4

hpk42:
[dovecot 2.4 is dropping userdb/passdb lookups via dictproxy, and recommends lua](https://doc.dovecot.org/2.4.1/installation/upgrade/2.3-to-2.4.html) 

This PR moves password verification to lua, and keeps creating addresses in python via a http request, so that `doveauth.py` keeps owning the creation checks, and nothing changes there. 

The new [auth.lua](https://github.com/chatmail/relay/pull/1049/changes#diff-faf13a4e84b0318c0ef5138c5d02414d7724af80c6dfcbce1be0a37ba9134ac9)  code is [extensively](https://github.com/chatmail/relay/pull/1049/changes#diff-7aa5d523128dcb9246a8be76965a382aa033ddaee574b3a4ed522d3ff046105a) [tested](https://github.com/chatmail/relay/pull/1049/changes#diff-7aa5d523128dcb9246a8be76965a382aa033ddaee574b3a4ed522d3ff046105a) and deploying it to an existing relay host works fine for me, and passes all tests.


cmdeploy/src/cmdeploy/dovecot/auth.lua.j2
```diff
@@ -0,0 +1,64 @@
...
+local function create(user, password)
+  local request = http_client:request({url = create_url, method = "POST"})
+  request:set_payload(user .. "\t" .. password)
+  return request:submit():status() == 200
+end
+
+-- Entry points called by dovecot
+
+function script_init()
+  http_client = dovecot.http.client({timeout = 5000, max_attempts = 1})
```
missytake: Why is the timeout so high?

cmdeploy/src/cmdeploy/dovecot/auth.lua.j2
```diff
@@ -0,0 +1,64 @@
...
+function auth_password_verify(req, password)
+  if not is_ours(req.user) then
+    return dovecot.auth.PASSDB_RESULT_USER_UNKNOWN, {}
+  end
+  local hash = password_hash(req.user)
+  if hash == nil then
+    if not create(req.user, password) then
+      return dovecot.auth.PASSDB_RESULT_USER_UNKNOWN, {}
+    end
+  elseif req:password_verify(hash, password) ~= 1 then
```
missytake, line 64: is `req:password_verify()` a built-in lua function in dovecot? Or where is it implemented?

missytake: Ah, if I understand it correctly, doveauth encrypts the password during create, dovecot verifies the password during lookup, and doveauth verifies the password only in rare cases when a user was accidentally created already while `create/` is called? Just asking again to verify.

missytake (approved):
Looks good! This already works with dovecot 2.3? looking forward to trying it out tomorrow.

lua isn't that bad it seems, I can read most of it. Hope the parts of the syntax which are not super intuitive don't contain surprises.

cmdeploy/src/cmdeploy/dovecot/auth.lua.j2
hpk42: glad you asked!  

a) it's milliseconds

b) the [dovecot 2.3 documentation](https://doc.dovecot.org/2.3/admin_manual/lua/#dovecot.http.client) and the [dovecot 2.3.21 source code](https://github.com/dovecot/core/blob/2.3.21/src/lib-lua/dlua-dovecot-http.c#L437-L456) somewhat disagree about the name, it is `request_timeout_msecs` and i fixed it here.  The docs have the function signature wrong.

cmdeploy/src/cmdeploy/dovecot/auth.lua.j2
hpk42: yes, that's exactly how it goes.  Dovecot has a builtin function to verify passwords, so logging in with existing accounts usually does not touch `doveauth.py`.  But due to SMTP/IMAP logins being potentially concurrent, we need to have the python side also verify if it detects concurrency.  Should be a rare happening, and doesn't matter as both sides agree on how to verify passwords.

cmdeploy/src/cmdeploy/dovecot/auth.lua.j2
missytake: ah, good to know :D so that's the kind of thing you need to watch out for with lua?^^

hpk42: yes. I could have noticed because the docs when they describe the args use `request_timeout_msecs`.  Usually dovecot docs are high quality.  Only the function signature name was wrong.

missytake:
Deployed on nine successfully.

PR #1051, opened 2026-09-10
chore(deps): Upgrade filtermail to v0.7.7

j-g00da:
### Performance

- Don't convert Bytes to Vec

### Miscellaneous Tasks

- Always use --locked flag in CI

### Performance

- Don't keep multiple copies of the same mail data in memory

### Performance

- Pass HttpsClient by reference instead of cloning it
- Remove connection pool

### Refactor

- Do not evaluate smtp_write! argument twice
- Do not clone the Config unnecessarily


missytake (approved):
Works on nine :)

PR #1052, opened 2026-09-14
feat: increase message size from 30 to 50 mb gross

r10s:
as discussed internally, this PR allows attachments and videos of about 38 mb net (50 mb gross).

once relays are sufficiently updated [^1], core will pick up the new default (core currently uses 18 mb only, an old default of non-chatmail-era)

this is quite some boost in quality or video length, roughly a doubling, while still being reasonably to overall storage
(with this PR we alllow a single message taking up to 1/10th of storage - when chatmail was started, this was even 1/3rd.
note, that large files may be deleted quickly to free up space, but there will add a pre-message still)

[^1]: i set myself a calendar reminder in 2 month to update core values :)


missytake (approved):
Sounds reasonable :)

missytake:
Note that many older instances will have the old value hardcoded though, as it used to be uncommented in the chatmail.ini. When this is merged, we should ping relay operators to increase `max_message_size`.

